PR #28, opened 2022-11-22
docs: explain why ci-docs.yaml is needed

davidhsingyuchen:
## License Acceptance

### DIFF
--- a/.github/workflows/ci-docs.yaml
+++ b/.github/workflows/ci-docs.yaml
@@ -1,3 +1,5 @@
+# This file is created to address the issue described in
+# https://stackoverflow.com/questions/66751567/return-passing-status-on-github-workflow-when-using-paths-ignore.
 name: CI
 on:
   push:
--- a/.github/workflows/ci-docs.yaml
+++ b/.github/workflows/ci-docs.yaml
@@ -1,5 +1,6 @@
 # This file is created to address the issue described in
 # https://stackoverflow.com/questions/66751567/return-passing-status-on-github-workflow-when-using-paths-ignore.
+# As a result, the set of jobs in this file should be equal to that in ci.yaml.
 name: CI
 on:
   push:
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yaml
```diff
@@ -1,4 +1,5 @@
-# When a third-party action is added (i.e., `uses`), please also add it to `download-licenses` in Makefile.
+# - When a third-party action is added (i.e., `uses`), please also add it to `download-licenses` in Makefile.
+# - The set of jobs in this file should be equal to that in ci-docs.yaml.
```
ningziwen, line 2: Do we need to also doc in ci-docs.yaml that "The set of jobs in this file should be equal to that in ci.yaml."?

ningziwen (approved):

.github/workflows/ci.yaml
davidhsingyuchen: Good point. Fixed in c379793. Also updated the comments in `ci.yaml` to be clearer when one needs to update `ci-docs.yaml`.

.github/workflows/ci-docs.yaml
```diff
@@ -1,3 +1,6 @@
+# This file is created to address the issue described in
```
ahsan-z-khan, line 1: Why do we need to link a stack overflow issue here? It's coming from GitHub's own documents. I think that would suffice.

ahsan-z-khan (changes requested):

weikequ (dismissed):
Great idea! Lgtm.

.github/workflows/ci-docs.yaml
weikequ: @davidhsingyuchen for context: https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/defining-the-mergeability-of-pull-requests/troubleshooting-required-status-checks#handling-skipped-but-required-checks

davidhsingyuchen: @weikequ Thanks a lot for the context! @ahsan-z-khan Fixed in [48e0498](https://github.com/runfinch/finch/pull/28/commits/48e0498635d9c3fb5c28de573c8ec82733987a77).

davidhsingyuchen: Resolving this thread due to https://github.com/runfinch/finch/pull/28#pullrequestreview-1190879386.

PR #39, opened 2022-11-23
fix: error readable by using the right placeholder

kokes:
*Description of changes:*

Logs aren't as readable as they could be, I saw this during `finch vm init`:

```
INFO[0009] sudoers file not found: %!w(*fs.PathError=&{open /etc/sudoers.d/finch-lima 2})
```

From the looks of it, we don't need the `%w` since the current logging implementation doesn't use `fmt.Errorf` under the hood anyway, and thus makes the output hard to read without the only potential benefit being utilised (unwrapping). And since the logging interface's methods themselves don't return anything, it's fairly likely unwrapping won't happen in any other implementation (there isn't even a reason for it anyway - the callee will have all the args available to it, so it can introspect the error passed directly).

(Since this `Infof` call is made right after an `errors.Is`, the whole `%v`/`%w` bit is redundant as it will just say "sudoers file not found: no such file or directory", but I didn't want to change error semantics in this PR.)

*Testing done:*

`make test-unit` and using `logrus` directly to verify none of its `*f` methods use `fmt.Errorf`.

#### License Acceptance


davidhsingyuchen (commented):
Thanks for the contribution! LGTM. Could you fix the `DCO` check by `git commit -s --amend --no-edit` and then `push -f`?

kokes:
> Could you fix the `DCO` check

Oh, right, done.

davidhsingyuchen (approved):

PR #64, opened 2022-11-29
docs: clarify installation instructions

Dentrax:
This commit clarifies some installation strategies for newcomers.

Signed-off-by: Furkan <furkan.turkal@trendyol.com>

Issue #, if available: `-`

*Description of changes:*
* Fix an issue where `make install` was just trying to symlink a file without building the actual binary itself. (Was a bit confusing for fresh users - and better UX to cover POLA law)
* Add `$` prefix to actual commands
* Add 2 new sections to installation section in the README
* Decreased build size  %15-%20 by stripping DWARF and debug symbols

*Testing done:*

* Issue the new commands that recently added to README and expect all works.

#### License Acceptance

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -119,7 +119,7 @@ The binary in _output can be directly used. E.g. initializing the vm and display
 ./_output/bin/finch version
 ```
 
-You can run `make install` to make finch binary globally accessible.
+You can also run `make install` to build and make finch binary globally accessible. However, note that this may conflict with the one installed via an official installer.
 
 
 ### Unit Testing
--- a/Makefile
+++ b/Makefile
@@ -26,7 +26,7 @@ CORE_OUTDIR := $(CURDIR)/$(CORE_FILENAME)/_output
 CORE_VDE_PREFIX ?= $(OUTDIR)/dependencies/vde/opt/finch
 LICENSEDIR := $(OUTDIR)/license-files
 VERSION := $(shell git describe --match 'v[0-9]*' --dirty='.modified' --always --tags)
-LDFLAGS := "-X $(PACKAGE)/pkg/version.Version=$(VERSION)"
+LDFLAGS := "-s -w -X $(PACKAGE)/pkg/version.Version=$(VERSION)"
 
 .DEFAULT_GOAL := all
 
@@ -114,7 +114,7 @@ copy:
 	(cd _output && tar c * | tar Cvx  $(DEST) )
 
 .PHONY: install
-install: copy
+install: finch copy
 	sudo ln -sf $(DEST)/bin/finch "$(BINDIR)/finch"
 
 uninstall.finch:
--- a/README.md
+++ b/README.md
@@ -24,14 +24,36 @@ To get started with Finch on macOS, the prerequisites are:
 * Intel or Apple Silicon M1 system for macOS
 * Recommended minimum configuration is 2 CPU, 4 GB memory
 
+##### From Go
+
+The easiest way to install the Finch is to use Go way by issuing the following command:
+
+```sh
+$ go install github.com/runfinch/finch/cmd@v0.1.0
+```
+
+##### From Releases
+
 Download a release package for your architecture from the [project's GitHub releases](https://github.com/runfinch/finch/releases) page, and once downloaded double click and follow the directions.
+
+##### From Source
+
+If you want to build and compile the source code, follow the instructions below:
+
+```sh
+$ git clone git@github.com:runfinch/finch.git
+$ cd finch
+$ make install
+```
+
+#### Running the VM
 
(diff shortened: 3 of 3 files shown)


Makefile
```diff
@@ -26,7 +26,7 @@ CORE_OUTDIR := $(CURDIR)/$(CORE_FILENAME)/_output
 CORE_VDE_PREFIX ?= $(OUTDIR)/dependencies/vde/opt/finch
 LICENSEDIR := $(OUTDIR)/license-files
 VERSION := $(shell git describe --match 'v[0-9]*' --dirty='.modified' --always --tags)
-LDFLAGS := "-X $(PACKAGE)/pkg/version.Version=$(VERSION)"
+LDFLAGS := "-s -w -X $(PACKAGE)/pkg/version.Version=$(VERSION)"
```
davidhsingyuchen, line 29: > Decreased build size %15-%20 by stripping DWARF and debug symbols

💯

CONTRIBUTING.md
```diff
@@ -119,7 +119,7 @@ The binary in _output can be directly used. E.g. initializing the vm and display
 ./_output/bin/finch version
 ```
 
-You can run `make install` to make finch binary globally accessible.
+You can also run `make install` to build and make finch binary globally accessible.
```
davidhsingyuchen, line 122: Not related to the changes, but since this part is being updated :)

```suggestion
You can also run `make install` to build and make finch binary globally accessible. However, note that this may conflict with the one installed via an official installer.
```

README.md
```diff
@@ -24,14 +24,36 @@ To get started with Finch on macOS, the prerequisites are:
...
+If you want to build and compile the source code, follow the instructions below:
+
+```sh
+$ git clone git@github.com:runfinch/finch.git
+$ cd finch
+$ make install
+```
+
+#### Running the VM
+
```
davidhsingyuchen, line 50: I'm a bit hesitant to add this because probably only `finch version` would work as the underlying VM is not there, and README is for users (say, compared to `CONTRIBUTING.md`, which is for devs), so instructions given here may need to yield a fully-functional thing. WDYT? /cc @ningziwen

davidhsingyuchen (commented):
Thank you for the contribution!

README.md
Dentrax: I'm not sure if I clearly follow your concern. All instructions here are fully-functional. For both installation methods, `finch version` works as expected.

davidhsingyuchen: My bad, I meant that `finch version` would probably be the **only** command that works. Important commands like VM and container ones won't work:

```sh
➜  finch git:(main) ./_output/bin/finch pull alpine
FATA[0000] fork/exec /local/home/davidhyc/davidhsingyuchen/finch/_output/lima/bin/limactl: no such file or directory
➜  finch git:(main) ./_output/bin/finch vm init
FATA[0000] fork/exec /local/home/davidhyc/davidhsingyuchen/finch/_output/lima/bin/limactl: no such file or directory
```

Dentrax: Ohhh, see now. Let me check what can I do to fix this. Maybe we should also install `limactl` if does not exist. Wdyt?

ningziwen: Lima and its deep dependencies are added by lima-and-qemu target in [makefile](https://github.com/runfinch/finch/blob/main/Makefile#L84), which is downloaded from the binaries generated by finch-core.

Basically finch go package is just the top layer of the whole Finch application. So it is expected that it can't work independently without the lower layers. Let me know if you have different thoughts.

Dentrax: I think I got lost on this and was thinking on the providing the best UX as possible.

For example, limactl, automatically [downloads](https://github.com/lima-vm/lima/blob/79a8791f965515e21f01f84e1230e05c9272fb98/pkg/downloader/downloader.go) all necessary artifacts (binaries etc.). And I have been thinking to implement the similar use case which would require the complete removal of the `lima-and-qemu` target in the makefile. So that we let finch do this job in the background. Eventually we will achieve,

* significantly reduce the binary size of release (don't need to maintain external artifacts)
* should work as a standalone binary
* compatible with `go install`, `brew`, and other package managers

stockholmux: It strikes me that the current install workflow makes total sense for those who just want to _use_ Finch, as it resolves all the dependencies in one step without giving much room for error.

Using Go or installing from source would probably confuse some users along the way as they could make a mistake or miss a step easily. Not to say that someone contributing or hacking Finch wouldn't use `go` or install from source but we probably want to keep these separate workflows and off the README. 

@Dentrax what do you think?

pendo324:
Closing since this has been inactive for 2 years. Please reply or re-open if you would like to continue this PR.

Dentrax:
Sorry for dropping the balls here @pendo324, almost ~2 years passed, so please feel free to take over if still needed or asked. 🤞

PR #67, opened 2022-11-29
docs: change branch protection from Strict to Loose

davidhsingyuchen:
## Summary

Changed the branch protection from `Strict` to `Loose` in the repo setting. PR adds documentation related to this change.

## License Acceptance


ningziwen:
Could you include the change of CHANGELOG.md rule should be in same PR? We need to make release please PR run the whole CI instead of ci.docs to make sure the tagged source code is healthy. https://github.com/runfinch/finch/blob/main/.github/workflows/ci.yaml#L9

davidhsingyuchen:
> Could you include the change of CHANGELOG.md rule should be in same PR? We need to make release please PR run the whole CI instead of ci.docs to make sure the tagged source code is healthy. https://github.com/runfinch/finch/blob/main/.github/workflows/ci.yaml#L9

I just realized that a non-user facing commit (e.g., chore) won't trigger a rebase of the release-please PR because there're no changes to CHANGELOG.md, but it can still introduce a bug, which means that even the checks are passing in the release-please PR, `main` could still contain a bug.

As a result, I feel that we don't need to trigger the real CI in release-please PRs; instead, we should document in our release process (i.e., the section added in this PR) that before merging a release-please PR, the latest commit on `main` must pass all the checks. WDYT?

ningziwen (approved):

Issue #23, opened 2022-11-22
Cannot see finch VM status


shaonm:
**What is the problem you're trying to solve?.**
I want to check the status of the VM that **finch** has started.

**Describe the feature you'd like**
`finch vm` should have a `status` command to see the vm status.

```
Usage:
  finch vm [command]

Available Commands:
  init        Initialize the virtual machine
  remove      Remove the virtual machine instance
  start       Start the virtual machine
  stop        Stop the virtual machine
```

Add `status` command,
`status        Status of the virtual machine`

AkihiroSuda:
Workaround:
```console
$ LIMA_HOME=/Applications/Finch/lima/data /Applications/Finch/lima/bin/limactl ls
NAME     STATUS     SSH                ARCH      CPUS    MEMORY    DISK      DIR
finch    Running    127.0.0.1:56835    x86_64    2       8GiB      100GiB    /Applications/Finch/lima/data/finch
```

niklasmtj:
Hey,
I would like to help out with this one and started implementing it over in a [fork](https://github.com/niklasmtj/finch/tree/feature/vm-status-command).  

One thing that I found which should be handled in a different way than with printing the output via `sva.logger.Infof("%s", logs)` is the output of lima's `ls` command. Because this way it is formatted in a wrong way as you can see below:
```
INFO[0000] NAME     STATUS     SSH                ARCH       CPUS    MEMORY    DISK      DIR
finch    Running    127.0.0.1:51977    aarch64    2       4GiB      100GiB    /Users/nm/Code/oss/finch/_output/lima/data/finch
```

Is there a custom logger available to print it in a right way? Or just use the regular `fmt.Printf`?

Since I'm pretty new to Golang I could also need a little help with testing the new command. So I am really open for help and advice.

ningziwen:
@niklasmtj Thanks! There is an existing internal [method](https://github.com/runfinch/finch/blob/a70b3d89fb672d3aca4393c73f26b6dc40153e9c/pkg/lima/lima.go#L29) to get the lima VM status. Maybe the new command can wrap this internal method and do the proper translation.

You can add new e2e tests [here](https://github.com/runfinch/finch/blob/a70b3d89fb672d3aca4393c73f26b6dc40153e9c/e2e/virtual_machine_test.go) and run "make test-e2e" to test it.

niklasmtj:
[...]

PR #83, opened 2022-12-02
feat: add `finch vm status` command

niklasmtj:
Issue #, if available: #23

*Description of changes:*
Adding the `finch vm status` command.

*Testing done:*
`make && make lint && make test-unit` since I did not yet updated the e2e tests where I need help to understand how to test the command

#### License Acceptance


ningziwen:
We also need e2e tests [here](https://github.com/runfinch/finch/blob/main/e2e/virtual_machine_test.go).

cmd/finch/virtual_machine_status.go
```diff
@@ -0,0 +1,61 @@
...
+		fmt.Println("Notexistent")
+		return nil
+	case lima.Stopped:
+		fmt.Println("Stopped")
+		return nil
+	case lima.Unknown:
+		fmt.Println("Unknown")
+		return nil
+	default:
+		fmt.Println("Undefined")
```
ningziwen, line 58: Even if "vm status" doens't assume vm is running, I think it assumes the vm status is valid. If it is "undefined", maybe it should return error?

cmd/finch/virtual_machine_status_test.go
```diff
@@ -0,0 +1,17 @@
...
+
+package main
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestNewStatusVMCommand(t *testing.T) {
```
ningziwen, line 12: I think there can be at least 5 unit test scenarioes, including 3 valid statuses and 2 invalid statuses. You can use "make coverage" to self-check the unit test coverage.

cmd/finch/virtual_machine_status.go
```diff
@@ -0,0 +1,61 @@
...
+		return nil
+	case lima.Nonexistent:
+		fmt.Println("Notexistent")
+		return nil
+	case lima.Stopped:
+		fmt.Println("Stopped")
+		return nil
+	case lima.Unknown:
+		fmt.Println("Unknown")
+		return nil
```
davidhsingyuchen, line 56: If the err returned by `GetVMStatus` is nil, but `GetVMStatus` also returns an `Unknown` status with it, then probably there's a bug. As a result, I think we can return an error for this case by incorporating the suggested changes in https://github.com/runfinch/finch/pull/83#discussion_r1038434102. WDYT?

```suggestion
```

cmd/finch/virtual_machine_status.go
```diff
@@ -0,0 +1,61 @@
...
+	switch status {
+	case lima.Running:
+		fmt.Println("Running")
+		return nil
+	case lima.Nonexistent:
+		fmt.Println("Notexistent")
+		return nil
+	case lima.Stopped:
+		fmt.Println("Stopped")
+		return nil
```
davidhsingyuchen, line 53: Not a blocker of this PR, can be done in a follow-up PR: We can use [`stringer`](https://pkg.go.dev/golang.org/x/tools/cmd/stringer) to simplify this part. Here is an [example](https://github.com/runfinch/finch/blob/6f9d33f11fd43cc8bc5765f34d2eb7e38c62e468/pkg/flog/log.go#L26) in our project.

cmd/finch/virtual_machine_status.go
niklasmtj: I removed the `fmt.Println` with a `return fmt.Errorf` to return the error since those two scenarios should not happen in a "regular" usage. Thanks againg for the feedback!

cmd/finch/virtual_machine_status.go
niklasmtj: That's a good point to know, that there is a bug when `Unknown` is returned. I added an error return for this scenario.

cmd/finch/virtual_machine_status_test.go
niklasmtj: Trying to figure out the mocking in the next days and try to implement the test scenarios.

ningziwen:
@niklasmtj Hey I see you pushed one commit labeled "apply pr feedback". We are still waiting for the changes of addressing other comments. Whenever you are ready, you can click "Re-request review" to let us know.

niklasmtj:
Hey, thanks. I will do that. Before that i have to get more into the mock testing. But will to this in the next couple of days.

niklasmtj:
Hey, I think I might need a helping hand here. I'm pretty new to Golang and have a hard time figuring out how the mocking is done. 
I did try to use the tests from the start or stop command to get a grasp on how to do it. I also got a couple of tests that are green but to be completely honest I don't know if this really tests the method in the right way. I will push my current tests and would appreciate any guidance on how to do it. Thank you in advance!

ningziwen:
> Hey, I think I might need a helping hand here. I'm pretty new to Golang and have a hard time figuring out how the mocking is done. I did try to use the tests from the start or stop command to get a grasp on how to do it. I also got a couple of tests that are green but to be completely honest I don't know if this really tests the method in the right way. I will push my current tests and would appreciate any guidance on how to do it. Thank you in advance!
> 
> If this PR takes too long you can also go forward and finish it. Reading the results would also help me a lot to understand how it have to be done :)

No rush for this change. I'm happy to use this PR to help you onboard Finch slowly. I will put comments in your existing change and give suggestions.

cmd/finch/virtual_machine_status_test.go
```diff
@@ -0,0 +1,126 @@
...
+	t.Parallel()
+
+	cmd := newStatusVMCommand(nil, nil)
+	assert.Equal(t, cmd.Name(), "status")
+}
+
```
ningziwen, line 22: Need one more func TestStatusVMAction_runAdapter() to improve coverage.   https://github.com/runfinch/finch/blob/d8174ff773f0f92ec94d6d97c753a872a98f74a0/cmd/finch/virtual_machine_init_test.go#L28

cmd/finch/virtual_machine_status_test.go
```diff
@@ -0,0 +1,126 @@
...
+			ctrl := gomock.NewController(t)
+			logger := mocks.NewLogger(ctrl)
+			lcc := mocks.NewLimaCmdCreator(ctrl)
+			lca := mocks.NewLimaConfigApplier(ctrl)
+
+			tc.mockSvc(lcc, logger, lca, ctrl)
+
+			err := newStatusVMAction(lcc, logger).run()
+			assert.Equal(t, err, tc.wantErr)
+		})
```
ningziwen, line 124: We also need to test the stdout here. It is not straightforward but this is the [example](https://github.com/runfinch/finch/blob/main/cmd/finch/version_test.go#L47).

Maybe we can extract the common method captureStdout to somewhere like a new package `fio`. @davidhsingyuchen may give good suggestions about where to extract to.

cmd/finch/virtual_machine_status_test.go
```diff
@@ -0,0 +1,126 @@
...
+	t.Parallel()
+
+	cmd := newStatusVMCommand(nil, nil)
+	assert.Equal(t, cmd.Name(), "status")
+}
+
+func TestStatusVMAction_run(t *testing.T) {
+	t.Parallel()
+
+	testCases := []struct {
```
ningziwen, line 26: Add a field for status output.

cmd/finch/virtual_machine_status_test.go
```diff
@@ -0,0 +1,126 @@
...
+)
+
+func TestNewStatusVMCommand(t *testing.T) {
+	t.Parallel()
+
+	cmd := newStatusVMCommand(nil, nil)
+	assert.Equal(t, cmd.Name(), "status")
+}
+
+func TestStatusVMAction_run(t *testing.T) {
```
ningziwen, line 23: Also need nonExistent and undefined use case

niklasmtj: I am not sure if we really need the [two cases here](https://github.com/runfinch/finch/blob/13eeaf5aea689eb7d261173459602c6c4e899a9c/cmd/finch/virtual_machine_status.go#L54:L57) since the `lima.GetVMStatus` returns an `error` when the instance state is unknown already. I added the `nonExistent` test in [453e2dc](https://github.com/runfinch/finch/pull/83/commits/453e2dc0b6d091b19b6fab2e6a19b0ba4e1fc267) where I also had to add the `Nonexistent` status in `toVMStatus`

cmd/finch/virtual_machine_status_test.go
niklasmtj: Added. Please check if this is okay :)

pkg/lima/lima.go
```diff
@@ -40,7 +40,7 @@ func GetVMStatus(creator command.LimaCmdCreator, logger flog.Logger, instanceNam
 func toVMStatus(status string, logger flog.Logger) (VMStatus, error) {
 	logger.Debugf("Status of virtual machine: %s", status)
 	switch status {
-	case "":
+	case "", "Nonexistent":
```
ningziwen, line 43: This method is to translate from upstream state to local package state. the upstream state doesn't have "Nonexistent". It just returns "" to show it is nonexistent.

cmd/finch/virtual_machine_status_test.go
```diff
@@ -0,0 +1,198 @@
...
+			statusOutput: "Nonexistent",
+			mockSvc: func(
+				lcc *mocks.LimaCmdCreator,
+				logger *mocks.Logger,
+				lca *mocks.LimaConfigApplier,
+				ctrl *gomock.Controller,
+			) {
+				getVMStatusC := mocks.NewCommand(ctrl)
+				lcc.EXPECT().CreateWithoutStdio("ls", "-f", "{{.Status}}", limaInstanceName).Return(getVMStatusC)
+				getVMStatusC.EXPECT().Output().Return([]byte("Nonexistent"), nil)
```
ningziwen, line 132: getVMStatusC.EXPECT().Output().Return([]byte(""), nil)

cmd/finch/virtual_machine_status.go
```diff
@@ -0,0 +1,59 @@
...
+	case lima.Running:
+		fmt.Println("Running")
+		return nil
+	case lima.Nonexistent:
+		fmt.Println("Nonexistent")
+		return nil
+	case lima.Stopped:
+		fmt.Println("Stopped")
+		return nil
+	case lima.Unknown:
```
ningziwen, line 54: I think unknown and undefined don't make a difference in this layer. Maybe no need to separate Unknown or Undefined here.

cmd/finch/virtual_machine_status_test.go
ningziwen: https://github.com/runfinch/finch/pull/83/files#r1056636231

pkg/lima/lima.go
niklasmtj: Got you, thanks. Removed it

cmd/finch/virtual_machine_status.go
niklasmtj: Alrighty, adjusted the case and removed the part and made it to the default case.

cmd/finch/virtual_machine_status_test.go
niklasmtj: Done that, but have to take a closer look into getting the stdout as you suggested above.

ningziwen: Yeah this change is not pushed yet.

ningziwen:
Not a big deal but fyi :) https://github.com/runfinch/finch/blob/main/CONTRIBUTING.md?plain=1#L88

niklasmtj:
Oh I'm sorry, this makes sense. Did not know this. Were used to use it as a reference for me of the things to do :)

cmd/finch/virtual_machine_status_test.go
niklasmtj: I tried implementing it like:
```go
statusActionFunc := func() {
	err := newStatusVMAction(lcc, logger).run()
	assert.Equal(t, err, tc.wantErr)
}

output := captureStdout(t, statusActionFunc)
assert.Equal(t, tc.wantStatusOutput, output)
```
For the sake of simplicity and testing I used the method already implemented. Let me know how we want to implement the `fio` package so I can use it in the right way. 

My problem is now that the tests are really flaky. Sometimes they pass sometimes they don't. 
One try it is:
```
expected: "Running\n"
actual  : ""
```
Next one it passes. 

Does this happen because of the parallel tests?

davidhsingyuchen: We can and probably should replace `captureStdout` with dependency injection. We can pass an `io.Writer` to `newStatusVMAction`, which will be `os.Stdout` in the real code and a `*bytes.Buffer` in the testing code. Demonstrating the relevant part with some real code:

```go
func test() {
	var b bytes.Buffer
	status(&b) // would be status(os.Stdout) in the real implementation (i.e., when not testing)
	if string(b.Bytes()) == "running\n" { // assert whatever we want here, please use the assert lib in the real testing code
		fmt.Println("pass")
	} else {
		fmt.Println("fail")
	}
}

func status(stdout io.Writer) {
	fmt.Fprintln(stdout, "running")
}
```

ningziwen: It is not common to have flaky unit tests. Could you try the dependency injection approach? If it is still flaky, you can push the commit to the PR so we can take a look. @niklasmtj

niklasmtj:
Started with the e2e tests but still trying to understand how to implement the tests that makes the most sense. I was thinking about adding one per each `When` block to see if the `status` command is returning the desired output. 

In terms of syntax can I use the `testVersion` as a reference point?

davidhsingyuchen:
> I was thinking about adding one per each `When` block to see if the `status` command is returning the desired output.

If you're referring to adding an `It` block to in each `When` block in https://github.com/runfinch/finch/blob/main/e2e/virtual_machine_test.go, then I think it makes sense.

> In terms of syntax can I use the `testVersion` as a reference point?

Sure. Every file in the `e2e` package should have a consistent style. If more references are needed, you can also take a look at https://github.com/runfinch/common-tests/tree/main/tests.

niklasmtj:
Hey, just added the e2e tests and ran `make && make lint && make test-unit && make test-e2e` which passed successfully. Will have a look at the conflicts if I'm able to resolve them by myself. But I guess I will need a bit of help here 😀

cmd/finch/virtual_machine_status.go
davidhsingyuchen: Resolving this comment as it's not a blocker of this PR.

cmd/finch/virtual_machine_status_test.go
```diff
@@ -0,0 +1,206 @@
...
+				logger *mocks.Logger,
+				lca *mocks.LimaConfigApplier,
+				ctrl *gomock.Controller,
+			) {
+				getVMStatusC := mocks.NewCommand(ctrl)
+				lcc.EXPECT().CreateWithoutStdio("ls", "-f", "{{.Status}}", limaInstanceName).Return(getVMStatusC)
+				getVMStatusC.EXPECT().Output().Return([]byte("Undefined"), nil)
+				logger.EXPECT().Debugf("Status of virtual machine: %s", "Undefined")
+			},
+		},
```
davidhsingyuchen, line 154: something minor, @ningziwen cmiiw: re. https://github.com/runfinch/finch/pull/83#discussion_r1038437152, the "2 invalid cases" seem to refer to:

1. `cmd.Output()` inside `GetVMStatus` returns an error.
2. `toVMStatus` inside `GetVMStatus` returns an error.

This test case seems to cover the second point above, but that should be already covered by the `"unknown VM status"` test case below, so we may not need this test case.

```suggestion
```

davidhsingyuchen (commented):
Thanks! LGTM overall, just a minor comment. You can ignore the e2e test failure as it's likely caused by flakiness and is tracked in #140.

davidhsingyuchen (approved):

ningziwen (approved):

PR #97, opened 2022-12-07
ci: add workflow for installer tests

ningziwen:
*Description of changes:*
Add workflow to run installer tests.
The arm64 workflow has defects that will fail but can be used for initial setup. The follow-up manual steps are required to complete and clean up it. Will open an issue to track the fix after the PR is merged.

*Testing done:*
Added push triggering to test it.
https://github.com/runfinch/finch/actions/runs/3635735114/jobs/6135091416

#### License Acceptance


.github/workflows/release-installer.yaml
```diff
@@ -0,0 +1,105 @@
...
+
+permissions:
+  id-token: write   # This is required for requesting the JWT
+  contents: read    # This is required for actions/checkout
+
+jobs:
+  # TODO: fix the arm64 test installer.
+  # Currently the first time of calling any Finch command in arm64 hosts triggered by Github action will fail by
+  # the error "Error: Process completed with exit code 137." So the arm64 job will fail.
+  # We temporarily use follow-up manual steps to complete and clean up it.
```
pendo324, line 18: Which manual steps specifically? Should we note that somewhere?

pendo324 (commented):
Just left a few questions, looks good overall

.github/workflows/release-installer.yaml
ningziwen: Just same steps after e2e tests.
```
INSTALLED=true make test-e2e
echo 'y' | sudo bash /Applications/Finch/uninstall.sh
rm -rf Finch-${GITHUB_REF_NAME}-aarch64.pkg
```
Do you think this should be added to the comment?

davidhsingyuchen: My understanding (maybe incorrect):

1. `INSTALLED=true make test-e2e` step below would fail.
2. We then SSH to the host to manually run `INSTALLED=true make test-e2e`, and it should pass.
3. We then manually run the cleanup commands (i.e., `echo 'y' | sudo bash /Applications/Finch/uninstall.sh` and `rm -rf Finch-${GITHUB_REF_NAME}-aarch64.pkg`)

My question:

Instead of manual retry, Have we considered retrying the `INSTALLED=true make test-e2e` step in the workflow?

.github/workflows/release-installer.yaml
```diff
@@ -0,0 +1,105 @@
+# When a third-party action is added (i.e., `uses`), please also add it to `download-licenses` in Makefile.
+# When the installers are ready in installer private bucket, run this workflow to test them.
+# TODO: Add job of uploading installers to Github release when installer tests are stable.
+# "release" runner tag is to make target hosts deterministic for easy clean up.
+# TODO: Remove the "release" runner tag when installer tests are stable.
```
davidhsingyuchen, line 5: `for easy clean up` -> do you mean the cleanup mentioned in https://github.com/runfinch/finch/pull/97/files#r1042749876? If yes, it seems that only arm hosts need that label?

ningziwen: This is one reason. The other reason is if the installer candidates of future versions have anything wrong, there is still risk to mess up the host. Even if the host has running job, installer test/release can just wait several minutes as it is normally not time-sensitive. It doesn't hurt to be a bit conservative now. We can remove this when installer release has better automation and self host runners have better cleanup strategy.

e2e/e2e_test.go
```diff
@@ -15,7 +16,12 @@ import (
 	"github.com/runfinch/common-tests/tests"
 )
 
-const virtualMachineRootCmd = "vm"
+const (
+	virtualMachineRootCmd = "vm"
+	installedTestSubject  = "finch"
+)
+
+var installed = flag.Bool("installed", false, "the flag to show whether the tests are ran against installed application")
```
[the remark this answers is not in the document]
ningziwen: It doesn't have to be the binary installed by an installer. It can also be installed by `make install` or homebrew.

.github/workflows/release-installer.yaml
```diff
@@ -0,0 +1,105 @@
...
+          role-to-assume: ${{ secrets.ROLE }}
+          role-session-name: download-installer-session
+          aws-region: ${{ secrets.REGION }}
+      - name: Download from S3
+        run: |
+          aws s3 cp s3://${{ secrets.INSTALLER_PRIVATE_BUCKET_NAME }}/Finch-${GITHUB_REF_NAME}-aarch64.pkg Finch-${GITHUB_REF_NAME}-aarch64.pkg
+      - name: Silently install
+        # Rosetta is only needed in arm64
+        run: |
+          echo 'A' | sudo softwareupdate --install-rosetta
```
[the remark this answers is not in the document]
ningziwen: Here is the log when I reproduce it in a fresh macOS host without Rosetta installed.
```
~ % sudo installer -pkg Finch-v0.1.1-aarch64.pkg -target /
installer: This package requires Rosetta 2 to be installed.
                Please install Rosetta 2 and then try again.
                    `sudo softwareupdate --install-rosetta`

installer: Error - Finch can’t be installed on this computer.
```

I followed this [doc](https://support.apple.com/en-us/HT211861) to run "Get Info" for the pkg in my M1 macOS.
The "Kind" is "Installer package" instead of either mentioned in the doc above. I could open an issue to investigate why installer command can't work without Rosetta.

.github/workflows/release-installer.yaml
davidhsingyuchen: I see. That's really weird. Since we still need to SSH to the host even if it's broken into 2 workflows, we may just do the following instead:

1. As you've already mentioned, create an issue to document and also track this.
2. At [the step that's expected to fail](https://github.com/runfinch/finch/pull/97/files#diff-4efbc8a659c1b30475339c903cf017cee4fc608ed879624d82518a24968755b3R57), link the issue and write comments to provide step-by-step instructions for folks to follow.

e2e/e2e_test.go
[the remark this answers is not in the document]
davidhsingyuchen: Ah I see. Then the current description is good enough. Thanks!

.github/workflows/release-installer.yaml
davidhsingyuchen: Thank you. Maybe the followin?

```sh
# "release" runner tag is to make target hosts deterministic for easy clean up.
# This not only refers to the cleanup described in <issue #> for arm hosts.
# Currently tests on amd64 hosts sometimes also need manual cleanup via SSHing to the host.
# TODO: Remove the "release" runner tag when installer tests are stable.
```

.github/workflows/release-installer.yaml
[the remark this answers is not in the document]
ningziwen: I could also add TODO and the issue link to the comment.

davidhsingyuchen (approved):

pendo324 (approved):
LGTM, thanks for answering the comment questions

Issue #12, opened 2022-11-22
Print `nerdctl version` in `finch version`


AkihiroSuda:
**What is the problem you're trying to solve?.**

The `nerdctl version` information is not printed in `finch version`
```console
$ finch version
Finch version: v0.1.0
```

**Describe the feature you'd like**
I'd expect `finch version` to work like this (same as `nerdctl version` except the first three lines)
```console
$ finch version
Finch:
 Version: v0.1.0
  
Client:
 Version:	v1.0.0
 OS/Arch:	linux/amd64
 Git commit:	c00780a1f5b905b09812722459c54936c9e070e6
 buildctl:
  Version:	v0.10.5
  GitCommit:	bc26045116045516ff2427201abd299043eaf8f7

Server:
 containerd:
  Version:	v1.6.8
  GitCommit:	9cd3357b7fd7218e4aec3eae239db1f68a5a6ec6
 runc:
  Version:	1.1.4
  GitCommit:	v1.1.4-0-g5fd4c4d1
```

**Additional context**

ningziwen:
How about this?
```
$ finch version
Client:
 Version:	v0.1.0
 OS/Arch:	linux/amd64
 Git commit:	c00780a1f5b905b09812722459c54936c9e070e6
 nerdctl: 
   Version: v1.0.0
   GitCommit: <commit>
 buildctl:
  Version:	v0.10.5
  GitCommit:	bc26045116045516ff2427201abd299043eaf8f7

Server:
 containerd:
  Version:	v1.6.8
  GitCommit:	9cd3357b7fd7218e4aec3eae239db1f68a5a6ec6
 runc:
  Version:	1.1.4
  GitCommit:	v1.1.4-0-g5fd4c4d1
```

I feel adding a separated Finch version block makes the output format of "finch version" too special. From Finch users' perspective, Finch itself is the client itself so Finch version is just the client version. Nerdctl can be added as part of Client dependency.

ahsan-z-khan:
should we also add `lima version` in the `Client` ?

sam-berning:
Yeah, I think the nerdctl/buildkit/containerd/runc version is useful to our users because there are user-facing features that might be added in a new version of each of those. I’d argue that Lima features aren’t really user-facing for users of Finch because the Finch VM is mostly transparent.

ChandruShetty:
[...]

PR #108, opened 2022-12-08
feat: Print version with Client and Server

ahsan-z-khan:
Issue #, if available:
#12 

*Description of changes:*

Update the implementation `finch version` according to https://github.com/runfinch/finch/issues/12#issuecomment-1347252973

*Testing done:*
Yet to be updated

#### License Acceptance

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -12,7 +12,7 @@ linters-settings:
   lll:
     # 145 is just a lax value that does not require too much work to add this check,
     # and we don't want this to be too strict anyway.
-    line-length: 160
+    line-length: 145
     tab-width: 4
   makezero:
     always: true
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -12,7 +12,7 @@ linters-settings:
   lll:
     # 145 is just a lax value that does not require too much work to add this check,
     # and we don't want this to be too strict anyway.
-    line-length: 145
+    line-length: 160
     tab-width: 4
   makezero:
     always: true
--- a/cmd/finch/version.go
+++ b/cmd/finch/version.go
@@ -5,6 +5,7 @@ package main
 
 import (
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
 
@@ -73,48 +74,52 @@ func (va *versionAction) runAdapter(cmd *cobra.Command, args []string) error {
 
 func (va *versionAction) run() error {
 	status, err := lima.GetVMStatus(va.creator, va.logger, limaInstanceName)
-	//nolint: gocritic // will change to switch statement later
 	if err != nil {
-		fmt.Fprintf(va.stdOut, "Finch version:\t%s", version.Version)
-		return err
-	} else if status != lima.Running {
-		fmt.Fprintf(va.stdOut, "Finch version:\t%s", version.Version)
-	} else {
-		limaArgs := []string{"shell", limaInstanceName, "nerdctl", "version", "--format", "json"}
-		out, err := va.creator.CreateWithoutStdio(limaArgs...).Output()
-		if err != nil {
-			fmt.Fprintf(va.stdOut, "Finch version:\t%s", version.Version)
-			return err
-		}
+		va.printVersion()
+		return fmt.Errorf("failed to get VM status: %w", err)
+	}
+	if status != lima.Running {
+		va.printVersion()
+		return errors.New("detailed version info is unavailable because VM is not running")
+	}
 
-		var nerdctlVersion NerdctlVersionOutput
-		err = json.Unmarshal(out, &nerdctlVersion)
+	limaArgs := []string{"shell", limaInstanceName, "nerdctl", "version", "--format", "json"}
+	out, err := va.creator.CreateWithoutStdio(limaArgs...).Output()
+	if err != nil {
+		return fmt.Errorf("failed to create the nerdctl version command: %w", err)
+	}
 
-		if err != nil {
-			fmt.Fprintf(va.stdOut, "Finch version:\t%s", version.Version)
-			return err
-		}
+	var nerdctlVersion NerdctlVersionOutput
+	err = json.Unmarshal(out, &nerdctlVersion)
 
-		fmt.Fprintf(va.stdOut, "Client:\n")
-		fmt.Fprintf(va.stdOut, " Version:\t%s\n", version.Version)
-		fmt.Fprintf(va.stdOut, " OS/Arch:\t%s/%s\n", nerdctlVersion.Client.Os, nerdctlVersion.Client.Arch)
-		fmt.Fprintf(va.stdOut, " GitCommit:\t%s\n", version.GitCommit)
-		fmt.Fprintf(va.stdOut, "nerdctl:\n")
-		fmt.Fprintf(va.stdOut, " Version:\t%s\n", nerdctlVersion.Client.Version)
-		fmt.Fprintf(va.stdOut, " GitCommit:\t%s\n", nerdctlVersion.Client.GitCommit)
-		for _, compo := range nerdctlVersion.Client.Components {
-			fmt.Fprintf(va.stdOut, "%s:\n", compo.Name)
-			fmt.Fprintf(va.stdOut, "  Version:\t%s\n", compo.Version)
-			fmt.Fprintf(va.stdOut, "  GitCommit:\t%s\n", compo.Details.GitCommit)
-		}
-		fmt.Fprintf(va.stdOut, "\n")
-		fmt.Fprintf(va.stdOut, "Server:\n")
-		for _, compo := range nerdctlVersion.Server.Components {
-			fmt.Fprintf(va.stdOut, " %s:\n", compo.Name)
-			fmt.Fprintf(va.stdOut, "  Version:\t%s\n", compo.Version)
-			fmt.Fprintf(va.stdOut, "  GitCommit:\t%s\n", compo.Details.GitCommit)
-		}
+	if err != nil {
+		va.printVersion()
+		return fmt.Errorf("failed to JSON-unmarshal the nerdctl version output: %w", err)
+	}
+
+	fmt.Fprintf(va.stdOut, "Client:\n")
+	fmt.Fprintf(va.stdOut, " Version:\t%s\n", version.Version)
+	fmt.Fprintf(va.stdOut, " OS/Arch:\t%s/%s\n", nerdctlVersion.Client.Os, nerdctlVersion.Client.Arch)
+	fmt.Fprintf(va.stdOut, " GitCommit:\t%s\n", version.GitCommit)
+	fmt.Fprintf(va.stdOut, " nerdctl:\n")
+	fmt.Fprintf(va.stdOut, "  Version:\t%s\n", nerdctlVersion.Client.Version)
+	fmt.Fprintf(va.stdOut, "  GitCommit:\t%s\n", nerdctlVersion.Client.GitCommit)
+	for _, compo := range nerdctlVersion.Client.Components {
+		fmt.Fprintf(va.stdOut, " %s:\n", compo.Name)
+		fmt.Fprintf(va.stdOut, "  Version:\t%s\n", compo.Version)
+		fmt.Fprintf(va.stdOut, "  GitCommit:\t%s\n", compo.Details.GitCommit)
+	}
+	fmt.Fprintf(va.stdOut, "\n")
+	fmt.Fprintf(va.stdOut, "Server:\n")
+	for _, compo := range nerdctlVersion.Server.Components {
+		fmt.Fprintf(va.stdOut, " %s:\n", compo.Name)
+		fmt.Fprintf(va.stdOut, "  Version:\t%s\n", compo.Version)
+		fmt.Fprintf(va.stdOut, "  GitCommit:\t%s\n", compo.Details.GitCommit)
 	}
 
 	return nil
+}
+
+func (va *versionAction) printVersion() {
+	(fmt.Fprintf(va.stdOut, "Finch version:\t%s\n", version.Version))
 }
--- a/cmd/finch/version.go
+++ b/cmd/finch/version.go
@@ -7,7 +7,6 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
-	"os"
 
 	"github.com/spf13/cobra"
 
@@ -17,12 +16,12 @@ import (
 	"github.com/runfinch/finch/pkg/version"
 )
 
-func newVersionCommand(limaCmdCreator command.LimaCmdCreator, logger flog.Logger) *cobra.Command {
+func newVersionCommand(limaCmdCreator command.LimaCmdCreator, logger flog.Logger, stdOut io.Writer) *cobra.Command {
 	versionCommand := &cobra.Command{
 		Use:   "version",
 		Args:  cobra.NoArgs,
 		Short: "Shows Finch version information",
-		RunE:  newVersionAction(limaCmdCreator, logger).runAdapter,
+		RunE:  newVersionAction(limaCmdCreator, logger, stdOut).runAdapter,
 	}
 
 	return versionCommand
@@ -31,6 +30,7 @@ func newVersionCommand(limaCmdCreator command.LimaCmdCreator, logger flog.Logger
 type versionAction struct {
 	creator command.LimaCmdCreator
 	logger  flog.Logger
+	stdOut  io.Writer
 }
 
 // NerdctlVersionOutput captures the nerdctl version.
@@ -63,48 +63,45 @@ type NerdctlComponentsOutput struct {
 	}
 }
 
-func newVersionAction(creator command.LimaCmdCreator, logger flog.Logger) *versionAction {
-	return &versionAction{creator: creator, logger: logger}
+func newVersionAction(creator command.LimaCmdCreator, logger flog.Logger, stdOut io.Writer) *versionAction {
+	return &versionAction{creator: creator, logger: logger, stdOut: stdOut}
 }
 
 func (va *versionAction) runAdapter(cmd *cobra.Command, args []string) error {
-	return va.printVersion()
+	return va.run()
 }
 
-func (va *versionAction) printVersion() error {
+func (va *versionAction) run() error {
 	status, err := lima.GetVMStatus(va.creator, va.logger, limaInstanceName)
 	if (status != lima.Running) || (err != nil) {
 		fmt.Println("Finch version:", version.Version)
-		fmt.Println("Please run finch vm init to start the virtual machine")
+		fmt.Println("Please run finch vm start to start the virtual machine")
 	} else {
-
 		limaArgs := []string{"shell", limaInstanceName, "nerdctl", "version", "--format", "json"}
 		out, _ := va.creator.CreateWithoutStdio(limaArgs...).Output()
 
 		var nerdctlVersion NerdctlVersionOutput
 		err = json.Unmarshal(out, &nerdctlVersion)
 
-		var w io.Writer = os.Stdout
-
 		if err == nil {
-			fmt.Fprintf(w, "Client:\n")
-			fmt.Fprintf(w, " Version:\t%s\n", version.Version)
-			fmt.Fprintf(w, " OS/Arch:\t%s/%s\n", nerdctlVersion.Client.Os, nerdctlVersion.Client.Arch)
-			fmt.Fprintf(w, " Git commit:\t%s\n", version.GitCommit)
-			fmt.Fprintf(w, "nerdctl:\n")
-			fmt.Fprintf(w, " Version:\t%s\n", nerdctlVersion.Client.Version)
-			fmt.Fprintf(w, " Git commit:\t%s\n", nerdctlVersion.Client.GitCommit)
+			fmt.Fprintf(va.stdOut, "Client:\n")
+			fmt.Fprintf(va.stdOut, " Version:\t%s\n", version.Version)
+			fmt.Fprintf(va.stdOut, " OS/Arch:\t%s/%s\n", nerdctlVersion.Client.Os, nerdctlVersion.Client.Arch)
+			fmt.Fprintf(va.stdOut, " Git commit:\t%s\n", version.GitCommit)
+			fmt.Fprintf(va.stdOut, "nerdctl:\n")
+			fmt.Fprintf(va.stdOut, " Version:\t%s\n", nerdctlVersion.Client.Version)
+			fmt.Fprintf(va.stdOut, " Git commit:\t%s\n", nerdctlVersion.Client.GitCommit)
 			for _, compo := range nerdctlVersion.Client.Components {
-				fmt.Fprintf(w, "%s:\n", compo.Name)
-				fmt.Fprintf(w, "  Version:\t%s\n", compo.Version)
-				fmt.Fprintf(w, "  Git commit:\t%s\n", compo.Details.GitCommit)
+				fmt.Fprintf(va.stdOut, "%s:\n", compo.Name)
+				fmt.Fprintf(va.stdOut, "  Version:\t%s\n", compo.Version)
+				fmt.Fprintf(va.stdOut, "  Git commit:\t%s\n", compo.Details.GitCommit)
 			}
-			fmt.Fprintf(w, "\n")
-			fmt.Fprintf(w, "Server:\n")
+			fmt.Fprintf(va.stdOut, "\n")
+			fmt.Fprintf(va.stdOut, "Server:\n")
 			for _, compo := range nerdctlVersion.Server.Components {
-				fmt.Fprintf(w, " %s:\n", compo.Name)
-				fmt.Fprintf(w, "  Version:\t%s\n", compo.Version)
-				fmt.Fprintf(w, "  Git commit:\t%s\n", compo.Details.GitCommit)
+				fmt.Fprintf(va.stdOut, " %s:\n", compo.Name)
+				fmt.Fprintf(va.stdOut, "  Version:\t%s\n", compo.Version)
+				fmt.Fprintf(va.stdOut, "  Git commit:\t%s\n", compo.Details.GitCommit)
 			}
 		}
 	}
--- a/cmd/finch/version.go
+++ b/cmd/finch/version.go
@@ -5,26 +5,49 @@ package main
 
 import (
 	"fmt"
+	"os/exec"
 
 	"github.com/spf13/cobra"
 
+	"github.com/runfinch/finch/pkg/command"
+	"github.com/runfinch/finch/pkg/flog"
+	"github.com/runfinch/finch/pkg/lima"
 	"github.com/runfinch/finch/pkg/version"
 )
 
-var finchVersion = version.Version
-
-func newVersionCommand() *cobra.Command {
+func newVersionCommand(limaCmdCreator command.LimaCmdCreator, logger flog.Logger) *cobra.Command {
 	versionCommand := &cobra.Command{
 		Use:   "version",
 		Args:  cobra.NoArgs,
 		Short: "Shows Finch version information",
-		RunE:  versionAction,
+		RunE:  newVersionAction(limaCmdCreator, logger).runAdapter,
 	}
 
 	return versionCommand
 }
 
-func versionAction(cmd *cobra.Command, args []string) error {
-	fmt.Println("Finch version:", finchVersion)
+type versionAction struct {
+	creator command.LimaCmdCreator
+	logger  flog.Logger
+}
+
+func newVersionAction(creator command.LimaCmdCreator, logger flog.Logger) *versionAction {
+	return &versionAction{creator: creator, logger: logger}
+}
+
+func (va *versionAction) runAdapter(cmd *cobra.Command, args []string) error {
+	return va.printVersion()
+}
+
+func (va *versionAction) printVersion() error {
+	_, err := lima.GetVMStatus(va.creator, va.logger, limaInstanceName)
+	if err != nil {
+		fmt.Println("Finch version:", version.Version)
+	}
+
+	v, _ := exec.Command("lima", "nerdctl", "version").Output()
+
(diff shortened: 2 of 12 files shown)


davidhsingyuchen (commented):
Leaving a comment now although it's still a draft PR because we want to make the discussion  transparent:

> my question is, what other way we can call nerdctl from version commant to generate finch and nerdctl version together

If `--format` is not there, we can probably just print finch version followed by printing the thing returned from `nerdctl version` so that we don't need to duplicate [the printing logic](https://github.com/containerd/nerdctl/blob/854985261da22f0e435ad963ecb13fbbaa37758a/cmd/nerdctl/version.go#L73-L94). That way, if a new thing is added to the output of `nerdctl version`, we don't have to update our code.

If `--format` is there, we may need to mimic [`VersionInfo`](https://github.com/containerd/nerdctl/blob/854985261da22f0e435ad963ecb13fbbaa37758a/pkg/inspecttypes/dockercompat/info.go#L72) to define a struct like:

```go
type VersionInfo struct {
	Finch FinchVersion
	Client ClientVersion
	Server *ServerVersion
}
```

Add then we can

1. Call `nerdctl version --format json` to populate the nerdctl part of the struct.
2. Populate the finch part of the struct.
3. Handle the template ([example](https://github.com/containerd/nerdctl/blob/854985261da22f0e435ad963ecb13fbbaa37758a/cmd/nerdctl/version.go#L55-L71)).

Please let me know if this makes sense to you, thanks!

cmd/finch/version.go
```diff
@@ -5,26 +5,53 @@ package main
...
+	return va.printVersion()
+}
+
+func (va *versionAction) printVersion() error {
+	_, err := lima.GetVMStatus(va.creator, va.logger, limaInstanceName)
+	if err != nil {
+		fmt.Println("Finch version:", version.Version)
+	}
+
+	v, _ := exec.Command("lima", "nerdctl", "version", "--format", "json").Output()
```
pendo324, line 48: I think you can just use the `va.creator` again, instead of shelling out with exec:

```suggestion
        limaArgs := []string{"shell", limaInstanceName, "nerdctl", "version", "--format", "json"}
	out, err := va.creator.CreateWithoutStdio(limaArgs...).Output()
```

ningziwen:
I added one comment to the issue. I think we should close the final output format first before reviewing the PR.

cmd/finch/version.go
```diff
@@ -5,26 +5,53 @@ package main
...
+
+func newVersionAction(creator command.LimaCmdCreator, logger flog.Logger) *versionAction {
+	return &versionAction{creator: creator, logger: logger}
+}
+
+func (va *versionAction) runAdapter(cmd *cobra.Command, args []string) error {
+	return va.printVersion()
+}
+
+func (va *versionAction) printVersion() error {
```
ChandruShetty, line 42: Suggest adding lima version too.

ChandruShetty (dismissed):

cmd/finch/version.go
```diff
@@ -4,27 +4,108 @@
...
+	_, err := lima.GetVMStatus(va.creator, va.logger, limaInstanceName)
+	if err != nil {
+		fmt.Println("Finch version:", version.Version)
```
azhouwd, line 77: If the VM is in a broken status, we print `Finch version: ...`. Should we also print the `Finch version: ...` if the VM is in good status? Also should we print out some warning message to inform users that the output is printed due to VM is broken?

cmd/finch/version.go
```diff
@@ -4,27 +4,108 @@
...
+	}
+
+	limaArgs := []string{"shell", limaInstanceName, "nerdctl", "version", "--format", "json"}
+	out, _ := va.creator.CreateWithoutStdio(limaArgs...).Output()
```
azhouwd, line 81: It's better to do the error handling here:
```suggestion
	out, err := va.creator.CreateWithoutStdio(limaArgs...).Output()
	if err != nil {...}
```

cmd/finch/version.go
```diff
@@ -4,27 +4,108 @@
...
+
+	var nerdctlVersion NerdctlVersionOutput
+	err = json.Unmarshal(out, &nerdctlVersion)
+
+	var w io.Writer = os.Stdout
+
+	if err == nil {
+		fmt.Fprintf(w, "Client:\n")
+		fmt.Fprintf(w, " Version:\t%s\n", version.Version)
```
azhouwd, line 90: I assume we can directly call `fmt.Printf("Version:\t%s\n", version.Version)` here, curious about what was the intention to use `Fprintf`?

cmd/finch/version_test.go
```diff
@@ -32,17 +35,22 @@ func captureStdout(t *testing.T, f func()) string {
...
-func TestVersionAction(t *testing.T) {
-	mockCmd := newVersionCommand()
-	versionActionFunc := func() {
-		require.NoError(t, versionAction(mockCmd, []string{}))
-	}
-
-	output := captureStdout(t, versionActionFunc)
-	assert.Contains(t, output, finchVersion)
-}
+//TODO: Ang's review
```
azhouwd, line 45: For the version info related to nerdctl, we can mock the result from `CreateWithoutStdio`, an example to follow:
https://github.com/runfinch/finch/blob/main/cmd/finch/virtual_machine_init_test.go#L239
from line 239 to line 241.

e2e/version_test.go
```diff
@@ -14,6 +14,7 @@ import (
 	"github.com/runfinch/finch/pkg/version"
 )
 
+//TODO: Ang's review
```
azhouwd, line 17: The only way I can come up with now is to shell into the VM using binary `/Applications/finch/lima/bin/limactl` and run the version command, then compare with `finch version`. This would work in CI

cmd/finch/version_test.go
```diff
@@ -32,17 +37,68 @@ func captureStdout(t *testing.T, f func()) string {
...
+
+			ctrl := gomock.NewController(t)
+			logger := mocks.NewLogger(ctrl)
+			lcc := mocks.NewLimaCmdCreator(ctrl)
+
+			versionActionFunc := func() {
+				require.NoError(t, newVersionAction(lcc, logger).printVersion())
+			}
+		
+			output := captureStdout(t, versionActionFunc)
```
davidhsingyuchen, line 84: https://github.com/runfinch/finch/pull/83#discussion_r1065119672 is also applicable to this PR.

cmd/finch/version.go
ahsan-z-khan: Added  `"Please run finch vm start to start the virtual machine"` when VM is not in a good status

cmd/finch/version.go
ahsan-z-khan: `Fprintf` writes the output to w

cmd/finch/version.go
```diff
@@ -4,27 +4,110 @@
...
+
+func newVersionAction(creator command.LimaCmdCreator, logger flog.Logger) *versionAction {
+	return &versionAction{creator: creator, logger: logger}
+}
+
+func (va *versionAction) runAdapter(cmd *cobra.Command, args []string) error {
+	return va.printVersion()
+}
+
+func (va *versionAction) printVersion() error {
```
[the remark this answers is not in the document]
ahsan-z-khan: updated

cmd/finch/version_test.go
ahsan-z-khan: Updated on the latest push. Can you check and suggest how to assert through `&stdOut'

cmd/finch/version_test.go
```diff
@@ -76,29 +82,75 @@ func TestPrintVersion(t *testing.T) {
...
+
+				command := mocks.NewCommand(ctrl)
+				lcc.EXPECT().CreateWithoutStdio("shell", limaInstanceName, "nerdctl", "version", "--format", "json").Return(command)
+				command.EXPECT().Output().Return([]byte(""), nil)
+			},
+			versionOutput: TestFinchVersion{
+				"Client:",
+				"Version:       v1.0.0",
+				"OS/Arch:       linux/amd64",
+				"Git commit:    8ua9sdyas7dg0a78sdbasdasdasd78asudh980b",
```
pendo324, line 123: I don't see `tc.versionOutput` being used anywhere currently. Also, not sure if we want to rely on these exact git commit SHAs in our tests. That would mean we have to update these pretty regularly, right? Maybe for these you can just check that they aren't empty unless there's a way to mock them easily?

ahsan-z-khan: These are randomly generated SHA, not the actual ones. But follows the format.

For the structs, I was thinking doing `assert.Contains(t, tc.versionOutput, stdOut.String()` , but errors out. looking into it

e2e/version_test.go
```diff
@@ -4,20 +4,32 @@
...
-
-	"github.com/runfinch/finch/pkg/version"
 )
 
-var testVersion = func(o *option.Option) {
-	ginkgo.Context("Version", func() {
-		ginkgo.Specify("Test version", func() {
-			gomega.Expect(command.StdoutStr(o, "version")).To(gomega.Equal(fmt.Sprintf("Finch version: %s", version.Version)))
+// Checks finch version
+var testVersion = func(o *option.Option, m *option.Option) {
```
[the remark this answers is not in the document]
davidhsingyuchen: > Other have dependencies on nerdctl output/format.

That's true, but I don't think they will change frequently if ever. At the end of the day, we're printing them to our users, so we may want to verify them.

> 500 characters which fails in the golangci for lll

We can either disable `lll` for that line ([example](https://github.com/runfinch/common-tests/blob/1d5ec0db09b523f47f9825ef7237ed1d9c51401a/tests/run.go#L437)) or [use `(?x)` to span the regex itself across multiple lines](https://superuser.com/a/1748758/730265).

cmd/finch/version_test.go
ahsan-z-khan: I’ve been trying to assert through this,
```var output bytes.Buffer
			newVersionAction(lcc, logger, &output).run()
			assert.Contains(t, output.String(), tc.versionOutput.Client)```
But during debug, i can see that `&output`  buffer is not getting filled through the `io.Writer`

File: cmd/finch/version_test.go [hunk 12]
```diff
@@ -5,44 +5,152 @@ package main
...
+
+func TestVersionAction_run(t *testing.T) {
+	t.Parallel()
+
+	testCases := []struct {
+		name          string
+		wantErr       error
+		cmd           *cobra.Command
+		mockSvc       func(*mocks.LimaCmdCreator, *mocks.Logger, *gomock.Controller)
+		versionOutput TestFinchVersion
```
davidhsingyuchen, line 101: We probably want to replace this with a `postRunCheck` ([example](https://github.com/runfinch/finch/blob/52ce4e77aeb679a380f7b580bb13ddd5c8a45b01/pkg/config/nerdctl_config_applier_test.go#L38)). In `postRunCheck`, we can assert each line of the stdout when we expect there's no error and just return when an error is expected (`wantErr` is not `nil`).

```suggestion
		postRunCheck func(t *testing.T, stdout []byte)
```

cmd/finch/version.go
```diff
@@ -4,27 +4,107 @@
...
+
+func (va *versionAction) runAdapter(cmd *cobra.Command, args []string) error {
+	return va.run()
+}
+
+func (va *versionAction) run() error {
+	status, err := lima.GetVMStatus(va.creator, va.logger, limaInstanceName)
+	if (status != lima.Running) || (err != nil) {
+		fmt.Println("Finch version:", version.Version)
+		fmt.Println("Please run finch vm start to start the virtual machine")
```
davidhsingyuchen, line 78: Also, we can print different things depending on the status. If it's nonexistent, then "vm init"; if it's stopped, then "vm start".

```suggestion
		fmt.Println(`Please run "finch vm start" to start the virtual machine to get the version of other components (e.g., nerdctl)`)
```

cmd/finch/version.go
```diff
@@ -4,27 +4,107 @@
...
+	} else {
+		limaArgs := []string{"shell", limaInstanceName, "nerdctl", "version", "--format", "json"}
+		out, _ := va.creator.CreateWithoutStdio(limaArgs...).Output()
+
+		var nerdctlVersion NerdctlVersionOutput
+		err = json.Unmarshal(out, &nerdctlVersion)
+
+		if err == nil {
```
davidhsingyuchen, line 86: if the error is not nil, we may want to return it instead of a silent fail.

e2e/version_test.go
```diff
@@ -4,20 +4,25 @@
...
+		ginkgo.BeforeEach(func() {
+			command.RemoveAll(o)
+		})
+		ginkgo.AfterEach(func() {
+			command.RemoveAll(o)
+		})
+
+		ginkgo.It("Should print finch version information", func() {
+			//nolint: lll // Version output format is larger than 500character
+			gomega.Expect(command.StdoutStr(o, "version")).Should(gomega.MatchRegexp("Client:\\s+Version:\\s+v[0-9]+\\.[0-9]+\\.[0-9]+-[0-9]+-[A-Za-z0-9]+\\s+OS/Arch:\\s+[A-Za-z0-9]+/[A-Za-z0-9]+\\s+Git commit:\\s+[A-Za-z0-9]+\\s+nerdctl:\\s+Version:\\s+v([0-9]+(\\.[0-9]+)+)\\s+Git commit:\\s+[A-Za-z0-9]+\\s+buildctl:\\s+Version:\\s+v([0-9]+(\\.[0-9]+)+)\\s+Git commit:\\s+[A-Za-z0-9]+\\s+Server:\\s+containerd:\\s+Version:\\s+v(\\d(\\.\\d)+)\\s+Git commit:\\s+[A-Za-z0-9]+\\s+runc:\\s+Version:\\s+([A-Za-z0-9]+(\\.[A-Za-z0-9]+)+)\\s+Git commit:\\s+v([0-9]+(\\.[0-9]+)+)-[0-9]+-[A-Za-z0-9]+"))
```
ningziwen, line 25: Should we inject Version.version here?

ahsan-z-khan: Yeah we can do that. 
```
versionInfo := command.StdoutStr(o, "version")
gomega.Expect(versionInfo).Should(gomega.ContainSubstring(version.Version))
gomega.Expect(versionInfo).Should(gomega.ContainSubstring(version.GitCommit))
```

cmd/finch/version.go
ahsan-z-khan: I would rather not complicate the command and print different status here, which is already being implement in `finch vm status`. I will just remove this line

.golangci.yaml
```diff
@@ -12,7 +12,7 @@ linters-settings:
   lll:
     # 145 is just a lax value that does not require too much work to add this check,
     # and we don't want this to be too strict anyway.
-    line-length: 145
+    line-length: 160
```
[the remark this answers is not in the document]
davidhsingyuchen: If we cannot break it into multiple lines for some reason, we can add `nolint:lll` and explain the reason behind.

ahsan-z-khan:
Note: Lint passed locally for `nolint` but failing in CI

e2e/version_test.go
ahsan-z-khan: added

cmd/finch/version_test.go
```diff
@@ -5,44 +5,127 @@ package main
...
-	assert.Equal(t, cmd.Name(), "version")
-}
+	testCases := []struct {
+		name         string
+		wantErr      error
+		cmd          *cobra.Command
+		mockSvc      func(*mocks.LimaCmdCreator, *mocks.Logger, *gomock.Controller)
+		postRunCheck func(t *testing.T, stdout []byte)
+	}{
+		{
```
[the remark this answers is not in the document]
pendo324: I think we're still missing the error test case, but everything else looks good in the latest revision

pendo324:
The linting error looks pretty weird. Let's retry it after the other workflows finish

cmd/finch/version_test.go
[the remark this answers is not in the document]
ahsan-z-khan: Not sure why Linter fails in CI during environment preparation. 

Example:
```
  /usr/bin/tar: ../../../go/pkg/mod/inet.af/tcpproxy@v0.0.0-20220326234310-be3ee21c9fa0/.travis.yml: Cannot open: File exists
  Error: /usr/bin/tar: ../../../go/pkg/mod/inet.af/tcpproxy@v0.0.0-20220326234310-be3ee21c9fa0/tcpproxy.go: Cannot open: File exists
```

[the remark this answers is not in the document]
pendo324: Yeah the linter error does look really weird, but I think this comment is about 
having a mock where lima.GetVMStatus returns an error, instead of nil, like

```golang
...
    getVMStatusC.EXPECT().Output().Return(nil, errors.new("Some error"))
...
```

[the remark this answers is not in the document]
ahsan-z-khan: added in new push

cmd/finch/version_test.go
```diff
@@ -5,44 +5,177 @@ package main
...
+	"github.com/golang/mock/gomock"
 	"github.com/stretchr/testify/assert"
-	"github.com/stretchr/testify/require"
 )
 
-func captureStdout(t *testing.T, f func()) string {
-	temp := os.Stdout
-	read, write, _ := os.Pipe()
-	os.Stdout = write
+func TestVersionCommand(t *testing.T) {
```
davidhsingyuchen, line 20: To make the name of the testing function align with the name of the tested function.

```suggestion
func TestNewVersionCommand(t *testing.T) {
```

cmd/finch/version.go
```diff
@@ -4,27 +4,117 @@
...
+
+func (va *versionAction) run() error {
+	status, err := lima.GetVMStatus(va.creator, va.logger, limaInstanceName)
+	//nolint: gocritic // will change to switch statement later
+	if err != nil {
+		fmt.Fprintf(va.stdOut, "Finch version:\t%s", version.Version)
+		return err
+	} else if status != lima.Running {
+		fmt.Fprintf(va.stdOut, "Finch version:\t%s", version.Version)
+	} else {
```
davidhsingyuchen, line 82: - We can use early return here to reduce the nesting levels of the code below.
- We can create a wrapper function that will call `fmt.Fprintf(va.stdOut, "Finch version:\t%s", version.Version)` if the returned error is not nil so that we can remove the duplicated statements (there're 2 more in the code below).
- Return an error when the VM is not running: Checked nerdctl's code, and it also returns an error if some steps fail (after printing whatever is available).

```suggestion
	if err != nil {
		return fmt.Error("failed to get VM status: %w", err)
	}
	if status != lima.Running {
		return errors.New("detailed version info is unavailable because VM is not running")
	}
```

ahsan-z-khan:
> > Note: Lint passed locally for `nolint` but failing in CI
> 
> The linting error looks pretty weird. Let's try retrying it after the other workflows finish

I think I got it. It was `go.sum`. I need to run `go mod tidy' and force pushe

cmd/finch/version.go
```diff
@@ -4,27 +4,117 @@
...
+			fmt.Fprintf(va.stdOut, "Finch version:\t%s", version.Version)
+			return err
+		}
+
+		var nerdctlVersion NerdctlVersionOutput
+		err = json.Unmarshal(out, &nerdctlVersion)
+
+		if err != nil {
+			fmt.Fprintf(va.stdOut, "Finch version:\t%s", version.Version)
+			return err
```
davidhsingyuchen, line 95: Could we prefix the errors?

```suggestion
			return fmt.Errorf("failed to create the nerdctl version command: %w", err)
		}

		var nerdctlVersion NerdctlVersionOutput
		err = json.Unmarshal(out, &nerdctlVersion)
		if err != nil {
			fmt.Fprintf(va.stdOut, "Finch version:\t%s", version.Version)
			return fmt.Errorf("failed to JSON-unmarshal the nerdctl version output: %w", err)
```

cmd/finch/version.go
ahsan-z-khan: updated

cmd/finch/version.go
```diff
@@ -4,27 +4,107 @@
...
+	return &versionAction{creator: creator, logger: logger, stdOut: stdOut}
+}
+
+func (va *versionAction) runAdapter(cmd *cobra.Command, args []string) error {
+	return va.run()
+}
+
+func (va *versionAction) run() error {
+	status, err := lima.GetVMStatus(va.creator, va.logger, limaInstanceName)
+	if (status != lima.Running) || (err != nil) {
```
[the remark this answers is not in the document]
ahsan-z-khan: updated

ahsan-z-khan:
> > Note: Lint passed locally for `nolint` but failing in CI
> 
> The linting error looks pretty weird. Let's try retrying it after the other workflows finish

It's weird that it was passing before the latest push on `main`. Not sure if that has something to do

cmd/finch/version.go
```diff
@@ -4,27 +4,117 @@
...
+		fmt.Fprintf(va.stdOut, " Version:\t%s\n", version.Version)
+		fmt.Fprintf(va.stdOut, " OS/Arch:\t%s/%s\n", nerdctlVersion.Client.Os, nerdctlVersion.Client.Arch)
+		fmt.Fprintf(va.stdOut, " GitCommit:\t%s\n", version.GitCommit)
+		fmt.Fprintf(va.stdOut, "nerdctl:\n")
+		fmt.Fprintf(va.stdOut, " Version:\t%s\n", nerdctlVersion.Client.Version)
+		fmt.Fprintf(va.stdOut, " GitCommit:\t%s\n", nerdctlVersion.Client.GitCommit)
+		for _, compo := range nerdctlVersion.Client.Components {
+			fmt.Fprintf(va.stdOut, "%s:\n", compo.Name)
+			fmt.Fprintf(va.stdOut, "  Version:\t%s\n", compo.Version)
+			fmt.Fprintf(va.stdOut, "  GitCommit:\t%s\n", compo.Details.GitCommit)
```
davidhsingyuchen, line 108: To make it conform to the thing specified in https://github.com/runfinch/finch/issues/12#issuecomment-1347252973

```suggestion
		fmt.Fprintf(va.stdOut, " nerdctl:\n")
		fmt.Fprintf(va.stdOut, "  Version:\t%s\n", nerdctlVersion.Client.Version)
		fmt.Fprintf(va.stdOut, "  GitCommit:\t%s\n", nerdctlVersion.Client.GitCommit)
		for _, compo := range nerdctlVersion.Client.Components {
			fmt.Fprintf(va.stdOut, " %s:\n", compo.Name)
			fmt.Fprintf(va.stdOut, "  Version:\t%s\n", compo.Version)
			fmt.Fprintf(va.stdOut, "  GitCommit:\t%s\n", compo.Details.GitCommit)
```

cmd/finch/version_test.go
```diff
@@ -5,44 +5,177 @@ package main
...
+				logger.EXPECT().Debugf("Status of virtual machine: %s", "Running")
+
+				command := mocks.NewCommand(ctrl)
+				lcc.EXPECT().CreateWithoutStdio("shell", limaInstanceName, "nerdctl", "version", "--format", "json").Return(command)
+				//nolint: lll	// Version output format is larger than 500
+				command.EXPECT().Output().Return([]byte(`{"Client":{"Version":"v1.0.0","GitCommit":"c00780a1f5b905b09812722459c54936c9e070e6","GoVersion":"go1.19.2","Os":"linux","Arch":"arm64","Components":[{"Name":"buildctl","Version":"v0.10.5","Details":{"GitCommit":"bc26045116045516ff2427201abd299043eaf8f7"}}]},"Server":{"Components":[{"Name":"containerd","Version":"v1.6.8","Details":{"GitCommit":"9cd3357b7fd7218e4aec3eae239db1f68a5a6ec6"}},{"Name":"runc","Version":"1.1.4","Details":{"GitCommit":"v1.1.4-0-g5fd4c4d1"}}]}}`), nil)
+			},
+			postRunCheck: func(t *testing.T, stdout []byte) {
+				lines := strings.SplitAfter(string(stdout), "\n")
+
```
davidhsingyuchen, line 105: nit: prevent panic when there's a length mismatch

```suggestion
require.Len(t, lines, 18)
```

cmd/finch/version_test.go
```diff
@@ -5,44 +5,177 @@ package main
...
+			},
+			postRunCheck: func(t *testing.T, stdout []byte) {
+				lines := strings.SplitAfter(string(stdout), "\n")
 
-//nolint:paralleltest // TODO: Add t.Parallel after using dependency injection to pass a io.Writer to newVersionCommand.
-func TestVersionAction(t *testing.T) {
-	mockCmd := newVersionCommand()
-	versionActionFunc := func() {
-		require.NoError(t, versionAction(mockCmd, []string{}))
+				assert.Equal(t, lines[0], "Finch version:\t")
```
davidhsingyuchen, line 141: `lines := strings.SplitAfter(string(stdout), "\n")` can be removed.

```suggestion
				assert.Equal(t, string(stdout), "Finch version:\t\n")
```

pendo324 (dismissed):
The implementation and testing LGTM in the latest revision, but I'd prefer to wait for @davidhsingyuchen's review too before merging.

Also, the linter is still failing with some weird errors. I found some resources related to this:

https://github.com/golangci/golangci-lint-action/issues/23
https://github.com/golangci/golangci-lint-action/issues/244

It might just be a rare issue that happened to get triggered this time, but I haven't really looked into it yet

davidhsingyuchen (commented):
https://github.com/runfinch/finch/blob/main/CONTRIBUTING.md#pull-request-lifecycle: Please avoid resolving the comments left by a PR reviewer yourself, thanks!

ahsan-z-khan:
> > Note: Lint passed locally for `nolint` but failing in CI
> 
> The linting error looks pretty weird. Let's try retrying it after the other workflows finish

According to this: https://github.com/golangci/golangci-lint-action/issues/23

Setting 'skip-pkg-cache: true' on each run of the golangci-lint action seems to fix it. We have it setup as 'false'. Another solution is to config the ` go-linter`  in a separate action.

davidhsingyuchen:
> > > Note: Lint passed locally for `nolint` but failing in CI
> > 
> > 
> > The linting error looks pretty weird. Let's try retrying it after the other workflows finish
> 
> According to this: [golangci/golangci-lint-action#23](https://github.com/golangci/golangci-lint-action/issues/23)
> 
> Setting 'skip-pkg-cache: true' on each run of the golangci-lint action seems to fix it. We have it setup as 'false'. Another solution is to config the ` go-linter` in a separate action.

If we run `make lint` locally, some fixes are triggered, which means that it won't pass in CI. That line will be removed after addressing https://github.com/runfinch/finch/pull/108#discussion_r1066452161 anyway.

```sh
➜  finch git:(version_extension) make lint
golangci-lint run
➜  finch git:(version_extension) ✗ git --no-pager diff
diff --git a/cmd/finch/main.go b/cmd/finch/main.go
index 5b0f274..83c6889 100644
--- a/cmd/finch/main.go
+++ b/cmd/finch/main.go
@@ -39,7 +39,7 @@ func main() {
 	}
 }

-//nolint: lll //Line is more than 150 character
+// nolint: lll //Line is more than 150 character
 func xmain(logger flog.Logger, ffd path.FinchFinderDeps, fs afero.Fs, loadCfgDeps config.LoadSystemDeps, mem fmemory.Memory, stdOut io.Writer) error {
 	fp, err := path.FindFinch(ffd)
 	if err != nil {
```

cmd/finch/version_test.go
```diff
@@ -5,44 +5,179 @@ package main
...
+				assert.Equal(t, lines[8], "  Version:\tv0.10.5\n")
+				assert.Equal(t, lines[9], "  GitCommit:\tbc26045116045516ff2427201abd299043eaf8f7\n")
+				assert.Equal(t, lines[10], "\n")
+				assert.Equal(t, lines[11], "Server:\n")
+				assert.Equal(t, lines[12], " containerd:\n")
+				assert.Equal(t, lines[13], "  Version:\tv1.6.8\n")
+				assert.Equal(t, lines[14], "  GitCommit:\t9cd3357b7fd7218e4aec3eae239db1f68a5a6ec6\n")
+				assert.Equal(t, lines[15], " runc:\n")
+				assert.Equal(t, lines[16], "  Version:\t1.1.4\n")
+				assert.Equal(t, lines[17], "  GitCommit:\tv1.1.4-0-g5fd4c4d1\n")
```
davidhsingyuchen, line 126: nit, ensure that nothing else is on that line

```suggestion
				assert.Equal(t, lines[17], "  GitCommit:\tv1.1.4-0-g5fd4c4d1\n")
				assert.Equal(t, lines[18], "")
```

cmd/finch/version.go
davidhsingyuchen: My bad, I meant the following:

```go
func (va *versionAction) run() error {
	if err := va.printVersion(); err != nil {
		// Now we only need this statement once, which is clearer and easier to maintain.
		fmt.Fprintf(va.stdOut, "Finch version:\t%s\n", version.Version)
		return err
	}
	return nil
}

func (va *versionAction) printVersion() error {
	// Move the main logic to this function
	// ...
}
```

davidhsingyuchen (dismissed):
Feel free to do the suggested changes in a follow-up PR.

cmd/finch/version_test.go
ahsan-z-khan: okay sure. lets merge this and get the release ready. in the meantime I'll open follow up PR once releases

cmd/finch/version.go
davidhsingyuchen: Please create a follow-up PR for this. Resolving this now.

davidhsingyuchen (approved):

ahsan-z-khan:
6 `e2e` tests are failing on `arm64` which are not related to the PR. They demonstrate the behavior of Issue #140 .

cmd/finch/version.go
davidhsingyuchen: https://github.com/runfinch/finch/pull/147

Issue #107, opened 2022-12-08
Option to support volume mounts outside user $HOME


mharwani:
**What is the problem you're trying to solve?.**
Currently, finch containers can only access resources inside user's home directory. I think there should be more flexibility in mount locations exposed by the virtual machine.

**Describe the feature you'd like**
Allow users to configure additional mount locations during VM start. These locations will be accessible to containers through commands such as `finch run` or `finch cp`.

**Additional context**
Adding an option to configure new mount locations in `${HOME}/.finch/finch.yaml` seems like a good approach. And perhaps through a command line flag, for eg:
```
finch vm start -v /tmp/data
```

PR #123, opened 2022-12-15
docs: design of adding config to support additional directories

ningziwen:
Signed-off-by: Ziwen Ning <ningziwe@amazon.com>

Issue #, if available:
https://github.com/runfinch/finch/issues/107

*Description of changes:*
Doc

*Testing done:*

#### License Acceptance

### DIFF
--- a/docs/config_to_support_additional_directories.md
+++ b/docs/config_to_support_additional_directories.md
@@ -54,9 +54,8 @@ Pros:
 
 Cons:
 
-* Sometimes the error log in Lima context may not be descriptive in Finch context
-  * Mitigation: 
-    * Add Finch's own error message to wrap Lima's error message. We are already doing this in [init](https://github.com/runfinch/finch/blob/d8174ff773f0f92ec94d6d97c753a872a98f74a0/cmd/finch/virtual_machine_init.go#L84) and [start](https://github.com/runfinch/finch/blob/d8174ff773f0f92ec94d6d97c753a872a98f74a0/cmd/finch/virtual_machine_start.go#L75) today so no additional change is needed.
+* Sometimes the error log in Lima context may not be descriptive in Finch context.
+  * However, we are having finch error messages wrapping Lima error messages today in [init](https://github.com/runfinch/finch/blob/d8174ff773f0f92ec94d6d97c753a872a98f74a0/cmd/finch/virtual_machine_init.go#L84) and [start](https://github.com/runfinch/finch/blob/d8174ff773f0f92ec94d6d97c753a872a98f74a0/cmd/finch/virtual_machine_start.go#L75). This can mitigate this concern.
 * Depends on Lima error handing.
   * However, other commands also depend on Lima/Nerdctl error handling today and Finch can still catch it in e2e tests.)
 
@@ -91,7 +90,7 @@ Finch currently only supports macOS but may support other platforms in the futur
 
 * additional_workspace_locations
 * additional_workspace_directories
-* **additional_directories (recommended)**
+* additional_directories (recommended)
 * additional_locations
 * additional_work_directories
 * additional_work_locations
--- a/docs/config_to_support_additional_directories.md
+++ b/docs/config_to_support_additional_directories.md
@@ -6,7 +6,7 @@
 
 ## Approach
 
-The [finch.yaml](https://github.com/runfinch/finch/blob/main/finch.yaml#L35) file which is used to boot in Lima has Mounts field to handle the mount points. However, changing it in finch.yaml would make the configs only applied in `vm init`, and finch.yaml is expected to be the place to keep Finch's default config without being messed up with user's customized configs. So instead of adding to finch.yaml, I recommended adding additional_directories to Lima’s override.yaml file. Both `vm init` and `vm start` can apply the configs in override.yaml. This is same to how Finch applies cpu and memory configs today.
+The [finch.yaml](https://github.com/runfinch/finch/blob/d8174ff773f0f92ec94d6d97c753a872a98f74a0/finch.yaml#L35) file which is used to boot in Lima has Mounts field to handle the mount points. However, changing it in finch.yaml would make the configs only applied in `vm init`, and finch.yaml is expected to be the place to keep Finch's default config without being messed up with user's customized configs. So instead of adding to finch.yaml, I recommended adding additional_directories to Lima’s override.yaml file. Both `vm init` and `vm start` can apply the configs in override.yaml. This is same to how Finch applies cpu and memory configs today.
 
 For example, for Finch config:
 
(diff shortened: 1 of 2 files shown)


docs/config_to_support_additional_workspace_directories.md
```diff
@@ -0,0 +1,101 @@
...
+```
+images: []
+cpus: 2
+memory: 4GiB
+mounts:
+- location: /Volumes/mydir
+mountPoint: /Volumes/mydir
+writable: true
+sshfs:
+cache: true
```
pendo324, line 29: nit: looks like the indentation got messed up here?

docs/config_to_support_additional_workspace_directories.md
```diff
@@ -0,0 +1,101 @@
...
+
+### Platform consistency consideration
+
+Finch currently only supports macOS but may support other platforms in the future. As one field of the user facing finch config, having a consistent meaning can help users easy to use across different platforms. So we can make naming and description of the new field generic.
+
+#### Naming
+
+* additional_workspace_locations
+* **additional_workspace_directories (recommended)**
+* additional_directories
```
pendo324, line 87: I prefer this over `workspace_directories`, but I think both work

docs/config_to_support_additional_workspace_directories.md
```diff
@@ -0,0 +1,101 @@
+## Config to support additional workspace directories
+
+### Issue
+https://github.com/runfinch/finch/issues/107
+
+### Approach
+
+Lima yaml file has Mounts field to handle the mount points. Similar to other Finch config like cpu and memory, it is straightforward to write additional_workspace_directories to Lima’s override.yaml file.
```
vsiravar, line 8: [finch.yaml](https://github.com/runfinch/finch/blob/main/finch.yaml#L35) also has config for mounts.

docs/config_to_support_additional_workspace_directories.md
```diff
@@ -0,0 +1,101 @@
...
+
+*Option 1 Directly pass the Lima and return whatever error Lima returns. Contribute to Lima to enhance error handling if needed. (Recommended)*
+
+Pros:
+
+* Easy to implement
+
+Cons:
+
+* Sometimes the error log in Lima context may not be descriptive in Finch context
```
pendo324, line 55: We'd probably just need to improve Lima's error handling to mitigate this, right? Do you have any examples of a "bad error" that Lima would currently emit, and any ideas on how to fix it?

mharwani, line 55: Another option is to wrap Lima error into a more descriptive finch error

sam-berning: When I try to mount `/`, I get

```
FATA[0031] the YAML is invalid, saved the buffer as "lima.REJECTED.yaml": field `mounts[2].location` must not be a system path such as /etc or /usr
```

Which doesn't make sense if the user is only aware of `finch.yaml`

docs/config_to_support_additional_workspace_directories.md
```diff
@@ -0,0 +1,101 @@
...
+* Finch can’t catch it in e2e tests when a new location is allowed in Lima.
+
+*Option 3 Blocklist specific locations in Finch with different error messages.*
+Pros:
+
+* The error messages will be specific.
+
+Cons:
+
+* Finch can’t catch it in e2e tests when a new location is blocked in Lima.
```
vsiravar, line 76: Is this a con for option 1 as well?

docs/config_to_support_additional_workspace_directories.md
ahsan-z-khan: Can we catch Lima errors and modify it for Finch ?

docs/config_to_support_additional_workspace_directories.md
```diff
@@ -0,0 +1,101 @@
...
+Finch config applier will make override.yaml like this:
+
+```
+images: []
+cpus: 2
+memory: 4GiB
+mounts:
+- location: /Volumes/mydir
+mountPoint: /Volumes/mydir
+writable: true
```
mharwani, line 27: Option to make it read-only?

docs/config_to_support_additional_workspace_directories.md
monirul: i prefer `additional_directories` as it is concise and more generic.

docs/config_to_support_additional_workspace_directories.md
ningziwen: Added explanation about why using override.yaml instead of finch.yaml.

docs/config_to_support_additional_workspace_directories.md
ningziwen: Fixed.

docs/config_to_support_additional_workspace_directories.md
ningziwen: +1. Modified to additional_directories.

docs/config_to_support_additional_workspace_directories.md
ningziwen: I noticed in vm init and start, we already wrap Lima error messages with Finch error messages so should be sufficient for now. Let me know if you think this should be more descriptive.

ningziwen: We used buffer to do a simple replacement in nerdctl [command](https://github.com/runfinch/finch/blob/d8174ff773f0f92ec94d6d97c753a872a98f74a0/cmd/finch/nerdctl.go#L76). However it will become very tricky when doing complicated replacement. It may mess up the messages when there is overlap or there is a message change in upstream. It would also make any commands that require stdin exit. I tend to not do message replacing unless there is no other alternative.

docs/config_to_support_additional_workspace_directories.md
ningziwen: It is out of scope of this feature and we have not identified the use case of read-only option so far. We can evaluate it based on community's input later.

pendo324 (dismissed):

docs/config_to_support_additional_directories.md
```diff
@@ -0,0 +1,104 @@
...
+Pros:
+
+* Easy to implement
+
+Cons:
+
+* Sometimes the error log in Lima context may not be descriptive in Finch context.
+  * However, we are having finch error messages wrapping Lima error messages today in [init](https://github.com/runfinch/finch/blob/d8174ff773f0f92ec94d6d97c753a872a98f74a0/cmd/finch/virtual_machine_init.go#L84) and [start](https://github.com/runfinch/finch/blob/d8174ff773f0f92ec94d6d97c753a872a98f74a0/cmd/finch/virtual_machine_start.go#L75). This can mitigate this concern.
+* Depends on Lima error handing.
+  * However, other commands also depend on Lima/Nerdctl error handling today and Finch can still catch it in e2e tests.)
```
mharwani, line 60:
```suggestion
  * However, other commands also depend on Lima/Nerdctl error handling today and Finch can still catch it in e2e tests.
```

mharwani (approved):

pendo324 (approved):

mharwani (approved):

PR #127, opened 2022-12-20
fix: add stricter check for DOCKER_CONFIG var

### LINKED ISSUE #126: finch can not login
I use finch on Apple M1 (Ventura)
When i try to use command 

$ finch login 

system  show prompt for login 
Enter Username: nancom
Enter Password:
WARNING: Your password will be stored unencrypted in /Users/panupongchantaklang/.finch/config.json.
Configure a credential helper to remove this warning. See
https://docs.docker.com/engine/reference/commandline/login/#credentials-store

FATA[0005] error saving credentials: mkdir /Users/panupongchantaklang: permission denied

** 
because i change my default user folder from  panupongchantaklang -> nancom  when i new install my operating system

any config for change /Users/panupongchantaklang/ --> /Users/nancom/

pendo324:
Signed-off-by: Justin Alvarez <alvajus@amazon.com>

Issue #, if available: Closes #126 

*Description of changes:*
- Previously, this code was just checking for any substring containing `export DOCKER_CONFIG`, now we check for a line starting with the fully configured DOCKER_CONFIG variable (like `export DOCKER_CONFIG="/Users/justin/.finch"`)

*Testing done:*
- added unit test for this scenario
- I want to test this practically, but its difficult to change the user on my mac and keep the same VM. Thinking of ways to test before merging

#### License Acceptance

### DIFF
--- a/pkg/config/nerdctl_config_applier.go
+++ b/pkg/config/nerdctl_config_applier.go
@@ -69,7 +69,16 @@ func updateEnvironment(fs afero.Fs, user string) error {
 	}
 
 	profStr := string(profBuf)
-	if !strings.Contains(profStr, "export DOCKER_CONFIG") {
+	proStrs := strings.Split(profStr, "\n")
+	var hasVar = false
+	for _, str := range proStrs {
+		if strings.HasPrefix(str, fmt.Sprintf("export DOCKER_CONFIG=\"/Users/%s/.finch\"", user)) {
+			hasVar = true
+			break
+		}
+	}
+
+	if !hasVar {
 		profBufWithDockerCfg := fmt.Sprintf("%s\nexport DOCKER_CONFIG=\"/Users/%s/.finch\"\n", profStr, user)
 		if err := afero.WriteFile(fs, profileFilePath, []byte(profBufWithDockerCfg), 0o644); err != nil {
 			return fmt.Errorf("failed to write to profile file: %w", err)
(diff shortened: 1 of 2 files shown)


ollypom:
@pendo324 I was wondering if it would simplify things if the DOCKER_CONFIG variable was passed in as a [Env variable in the lima config](https://github.com/lima-vm/lima/blob/master/examples/default.yaml#L345)? Similar to https://github.com/runfinch/finch/pull/129

pendo324:
> @pendo324 I was wondering if it would simplify things if the DOCKER_CONFIG variable was passed in as a [Env variable in the lima config](https://github.com/lima-vm/lima/blob/master/examples/default.yaml#L345)? Similar to #129

Sorry for the late response @ollypom, I was out for the holidays and I'm still catching up on things. I believe I tried this before but wasn't able to get it working because of the fact that the DOCKER_CONFIG is dependent on the `$USER` (since its set to `/Users/$USER/.finch`).

Some of this complication is due to the fact that the Lima `env` map adds to `/etc/environment`, which _can't_ evaluate variables, by definition.

It's definitely possible that I overlooked something though, open to changing this behavior as its kinda hacky the way it is currently. The other advantage of the way it is right now is that its dynamically evaluated on VM init, which means it will get updated in the rare case that the `$USER` changes (which is pretty hard to test actually, which is the reason I haven't tried to merge this yet).

pendo324:
Deduping for all strings added to bashrc was already added by a separate PR ([see here](https://github.com/runfinch/finch/blob/main/pkg/config/nerdctl_config_applier.go#L60-L62)), closing

PR #128, opened 2022-12-20
feat: add config to support additional directories

ningziwen:
Signed-off-by: Ziwen Ning <ningziwe@amazon.com>

Issue #, if available:
https://github.com/runfinch/finch/issues/107

*Description of changes:*
add config to support additional directories according to the doc added in #123

*Testing done:*

#### License Acceptance

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,20 +94,20 @@ The `run` command has a `-v` option for volume mounts. See `Volume flags` under 
 
 Finch has a simple and extensible configuration. A configuration file at `${HOME}/.finch/finch.yaml` will be generated on first run. Currently, this config file has options for system resource limits for the underlying virtual machine. These default limits are generated dynamically based on the resources available on the host system, but can be changed by manually editing the config file.
 
-Currently, the options are:
-
-* CPUs [int] (required): the amount of vCPU to dedicate to the virtual machine
-* Memory [string] (required): the amount of memory to dedicate to the virtual machine
-* AdditionalDirectories [AdditionalDirectory{path [string]}] (optional): the work directories that are not supported by default. In macOS, only home directory is supported by default. For example, if you want to mount a directory into a container, and that directory is not under your home directory, then you'll need to specify this field to add that directory or any ascendant of it as a work directory.
-
 For a full list of configuration options, check [the struct here](pkg/config/config.go#L25).
 
 An example `finch.yaml` looks like this:
 
 ```yaml
+# CPUs: the amount of vCPU to dedicate to the virtual machine. (required)
 cpus: 4
+# Memory: the amount of memory to dedicate to the virtual machine. (required)
 memory: 4GiB
+# AdditionalDirectories: the work directories that are not supported by default. In macOS, only home directory is supported by default. 
+# For example, if you want to mount a directory into a container, and that directory is not under your home directory, 
+# then you'll need to specify this field to add that directory or any ascendant of it as a work directory. (optional)
 additional_directories:
+  # the path of each additional directory.
   - path: /Volumes
 ```
 
--- a/README.md
+++ b/README.md
@@ -96,22 +96,13 @@ Finch has a simple and extensible configuration. A configuration file at `${HOME
 
 Currently, the options are:
 
-* CPUs [int]: the amount of vCPU to dedicate to the virtual machine
-* Memory [string]: the amount of memory to dedicate to the virtual machine
-* AdditionalDirectories [AdditionalDirectory]: an optional field that states the work directories that are not supported by default. In macOS, only home directory is supported by default.
+* CPUs [int] (required): the amount of vCPU to dedicate to the virtual machine
+* Memory [string] (required): the amount of memory to dedicate to the virtual machine
+* AdditionalDirectories [AdditionalDirectory{path [string]}] (optional): the work directories that are not supported by default. In macOS, only home directory is supported by default.
 
 For a full list of configuration options, check [the struct here](pkg/config/config.go#L25).
 
-#### Examples
-
-1. `finch.yaml` without additional directories looks like this:
-
-```yaml
-cpus: 4
-memory: 4GiB
-```
-
-1. `finch.yaml` with additional directories that include one directory "/Volumes" looks like this:
+An example `finch.yaml` looks like this:
 
 ```yaml
 cpus: 4
--- a/README.md
+++ b/README.md
@@ -98,7 +98,7 @@ Currently, the options are:
 
 * CPUs [int] (required): the amount of vCPU to dedicate to the virtual machine
 * Memory [string] (required): the amount of memory to dedicate to the virtual machine
-* AdditionalDirectories [AdditionalDirectory{path [string]}] (optional): the work directories that are not supported by default. In macOS, only home directory is supported by default.
+* AdditionalDirectories [AdditionalDirectory{path [string]}] (optional): the work directories that are not supported by default. In macOS, only home directory is supported by default. For example, if you want to mount a directory into a container, and that directory is not under your home directory, then you'll need to specify this field to add that directory or any ascendant of it as a work directory.
 
 For a full list of configuration options, check [the struct here](pkg/config/config.go#L25).
 
--- a/README.md
+++ b/README.md
@@ -98,14 +98,26 @@ Currently, the options are:
 
 * CPUs [int]: the amount of vCPU to dedicate to the virtual machine
 * Memory [string]: the amount of memory to dedicate to the virtual machine
+* AdditionalDirectories [AdditionalDirectory]: an optional field that states the work directories that are not supported by default. In macOS, only home directory is supported by default.
 
 For a full list of configuration options, check [the struct here](pkg/config/config.go#L25).
 
-An example `finch.yaml` looks like this:
+#### Examples
+
+1. `finch.yaml` without additional directories looks like this:
 
 ```yaml
 cpus: 4
 memory: 4GiB
+```
+
+1. `finch.yaml` with additional directories that include one directory "/Volumes" looks like this:
+
+```yaml
+cpus: 4
+memory: 4GiB
+additional_directories:
+  - path: /Volumes
 ```
 
 ## What's next?
--- a/e2e/config_test.go
+++ b/e2e/config_test.go
@@ -108,7 +108,7 @@ var testConfig = func(o *option.Option, installed bool) {
 			var limaCfg limayaml.LimaYAML
 			err = yaml.Unmarshal(cfgBuf, &limaCfg)
 			gomega.Expect(err).ShouldNot(gomega.HaveOccurred())
-			gomega.Expect(limaCfg.CPUs).Should(gomega.BeNil())
+			gomega.Expect(limaCfg.CPUs).ShouldNot(gomega.BeNil())
 			gomega.Expect(*limaCfg.Memory).Should(gomega.Equal("6GiB"))
 		})
 
@@ -123,7 +123,7 @@ var testConfig = func(o *option.Option, installed bool) {
 			var limaCfg limayaml.LimaYAML
 			err = yaml.Unmarshal(cfgBuf, &limaCfg)
 			gomega.Expect(err).ShouldNot(gomega.HaveOccurred())
-			gomega.Expect(limaCfg.CPUs).Should(gomega.BeNil())
+			gomega.Expect(limaCfg.CPUs).ShouldNot(gomega.BeNil())
 			gomega.Expect(*limaCfg.Memory).Should(gomega.MatchRegexp(`\dGiB`))
 		})
 
--- a/e2e/config_test.go
+++ b/e2e/config_test.go
@@ -10,11 +10,14 @@ import (
 	"os/exec"
 	"path/filepath"
 
+	"github.com/lima-vm/lima/pkg/limayaml"
 	"github.com/onsi/ginkgo/v2"
 	"github.com/onsi/gomega"
 	"github.com/onsi/gomega/gexec"
 	"github.com/runfinch/common-tests/command"
 	"github.com/runfinch/common-tests/option"
+	"github.com/xorcare/pointer"
+	"gopkg.in/yaml.v3"
 )
 
 var finchConfigFilePath = os.Getenv("HOME") + "/.finch/finch.yaml"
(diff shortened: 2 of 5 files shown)


pkg/config/config.go
```diff
@@ -24,8 +24,9 @@ import (
 
 // Finch represents the configuration file for Finch CLI.
 type Finch struct {
-	CPUs   *int    `yaml:"cpus"`
-	Memory *string `yaml:"memory"`
+	CPUs                  *int     `yaml:"cpus"`
+	Memory                *string  `yaml:"memory"`
+	AdditionalDirectories []string `yaml:"additional_directories,omitempty"`
```
pendo324, line 29: Might be better to make this a slice of a struct, even if the struct just has one entry, like path, that way its non-breaking to add more fields to the struct in the future
```golang
type AdditionalDirectory struct {
    Path string
}
```

```suggestion
	AdditionalDirectories []AdditionalDirectory `yaml:"additional_directories,omitempty"`
```

pkg/config/lima_config_applier.go
```diff
@@ -43,6 +44,23 @@ func (lca *limaConfigApplier) Apply() error {
 
 	limaCfg.CPUs = lca.cfg.CPUs
 	limaCfg.Memory = lca.cfg.Memory
+	limaCfg.Mounts = []limayaml.Mount{}
+	for _, ad := range lca.cfg.AdditionalDirectories {
```
pendo324, line 48: This LGTM, just need to add some tests 👍

pkg/config/lima_config_applier.go
```diff
@@ -43,6 +44,23 @@ func (lca *limaConfigApplier) Apply() error {
...
+			Location: *ad.Path, MountPoint: *ad.Path, Writable: pointer.Bool(true),
+			SSHFS: limayaml.SSHFS{
+				Cache: pointer.Bool(true), FollowSymlinks: pointer.Bool(false),
+				SFTPDriver: pointer.String(""),
+			},
+			NineP: limayaml.NineP{
+				SecurityModel:   pointer.String(limayaml.Default9pSecurityModel),
+				ProtocolVersion: pointer.String(limayaml.Default9pProtocolVersion),
+				Msize:           pointer.String(limayaml.Default9pMsize), Cache: pointer.String(limayaml.Default9pCacheForRW),
+			},
```
davidhsingyuchen, line 61: It seems that we may only need to specify `Location` and `Writable` and omit all the other fields (i.e., effectively leaving them `null`) according to https://github.com/lima-vm/lima/blob/master/examples/default.yaml? We then don't have to check those fields in tests. We're just specifying the default values here anyway.

pkg/config/lima_config_applier.go
```diff
@@ -6,8 +6,9 @@ package config
 import (
 	"fmt"
 
-	"github.com/lima-vm/lima/pkg/limayaml"
+	limayaml "github.com/lima-vm/lima/pkg/limayaml"
```
davidhsingyuchen, line 9: Maybe unnecessary?

pkg/config/lima_config_applier.go
```diff
@@ -43,6 +44,23 @@ func (lca *limaConfigApplier) Apply() error {
 
 	limaCfg.CPUs = lca.cfg.CPUs
 	limaCfg.Memory = lca.cfg.Memory
+	limaCfg.Mounts = []limayaml.Mount{}
```
davidhsingyuchen, line 47: Won't this statement erase the home directory mount in finch.yaml?

README.md
```diff
@@ -98,14 +98,26 @@ Currently, the options are:
 
 * CPUs [int]: the amount of vCPU to dedicate to the virtual machine
 * Memory [string]: the amount of memory to dedicate to the virtual machine
+* AdditionalDirectories [AdditionalDirectory]: an optional field that states the work directories that are not supported by default. In macOS, only home directory is supported by default.
```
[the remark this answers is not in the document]
ningziwen: https://github.com/runfinch/finch/pull/123#discussion_r1050038581

My thought process of not using the naming in Finch config including "Mounts".

- I prefer not to include any implementation detail keyword in user facing features(commands/configs) so the user facing could be easily consistent across all the platforms as we don't know if the implementation detail may be different across different platforms. Even for same platform, keeping the naming generic can make the door open for the future re-architecture
- Virtualization is the implementation detail that users don't have to be aware.
- Finch config is user facing config.
- "Mounts" is the word specifically for virtualization in our context.

I'm aware the vm lifecycle command is an exception for now but still want to minimize this.

e2e/config_test.go
```diff
@@ -124,5 +124,20 @@ var testConfig = func(o *option.Option, installed bool) {
...
+			gomega.Expect(limaConfigFilePath).Should(gomega.BeARegularFile())
+			cfgBuf, err := os.ReadFile(filepath.Clean(limaConfigFilePath))
+			gomega.Expect(err).ShouldNot(gomega.HaveOccurred())
+			gomega.Expect(cfgBuf).Should(gomega.SatisfyAll(gomega.ContainSubstring("cpus: 6"),
+				gomega.ContainSubstring("memory: 4GiB"),
+				gomega.ContainSubstring("mounts:\n    - location: /Volumes\n      "+
+					"mountPoint: /Volumes\n      writable: true\n      sshfs:\n        cache: true\n        "+
+					"followSymlinks: false\n        sftpDriver: \"\"\n      9p:\n        securityModel: none\n        "+
+					"protocolVersion: 9p2000.L\n        msize: 128KiB\n        cache: mmap")))
+		})
```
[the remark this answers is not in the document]
ningziwen: > checking fields like msize seems to be a bit distracting

The code includes applying default values so we need to test the results have the default values to cover the code.

pkg/config/lima_config_applier.go
ningziwen: No. I mentioned in the design doc. https://github.com/runfinch/finch/blob/1bac92a99f2e031792018dedfb6425488af51f1e/docs/config_to_support_additional_directories.md?plain=1#L43

pkg/config/lima_config_applier.go
ningziwen: Yeah my bad. I didn't notice the final mount is applied the default values after merging the override config.

README.md
```diff
@@ -98,14 +98,26 @@ Currently, the options are:
...
 
 For a full list of configuration options, check [the struct here](pkg/config/config.go#L25).
 
-An example `finch.yaml` looks like this:
+#### Examples
```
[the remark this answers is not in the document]
ningziwen: Good call. I think we don't need a separated example with addressing this comment. https://github.com/runfinch/finch/pull/128/files#r1054778870

e2e/config_test.go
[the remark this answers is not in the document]
mharwani: Maybe we can compare against the default values from limayaml instead of literals?

e2e/config_test.go
```diff
@@ -124,5 +124,20 @@ var testConfig = func(o *option.Option, installed bool) {
 			startCmdSession := updateAndApplyConfig(o, []byte("memory: 0GiB"))
 			gomega.Expect(startCmdSession).Should(gexec.Exit(1))
 		})
+
+		ginkgo.It("updates config values when a config file is present with additional directories", func() {
```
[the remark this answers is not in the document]
ningziwen: Maybe we can only keep the 2 directories case as I don't think there is anything special for single directory case.

e2e/config_test.go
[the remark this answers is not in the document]
ningziwen: Yeah we can compare default values to catch the issues caused by upstream change.

mharwani (approved):

pkg/config/lima_config_applier.go
davidhsingyuchen: Thanks. Added the doc PR link in the description of this PR.

e2e/config_test.go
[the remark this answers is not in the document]
ningziwen: As we removed the default value from finch.yaml, override.yaml also doesn't have it now.

README.md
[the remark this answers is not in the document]
davidhsingyuchen: Thanks a lot for the explanation. Wasn't aware of the design doc. Maybe we can add an example to make it easier for users to know when they may want to use it? Like:

"For example, if you want to mount a directory into a container, and that directory is not under your home directory, then you'll need to specify this field to add that directory or any ascendant of it as a work directory."

e2e/config_test.go
```diff
@@ -124,5 +127,28 @@ var testConfig = func(o *option.Option, installed bool) {
...
+cpus: 6
+additional_directories:
+    - path: /Volumes
+    - path: /tmp/workspace`))
+			gomega.Expect(startCmdSession).Should(gexec.Exit(0))
+
+			gomega.Expect(limaConfigFilePath).Should(gomega.BeARegularFile())
+			cfgBuf, err := os.ReadFile(filepath.Clean(limaConfigFilePath))
+			gomega.Expect(err).ShouldNot(gomega.HaveOccurred())
+			var limaCfg limayaml.LimaYAML
```
davidhsingyuchen, line 142: For the sake of consistency, could we update all the other test cases to use this approach? Fine to do it in a follow-up PR.

pkg/config/config.go
```diff
@@ -22,10 +22,16 @@ import (
 	"github.com/runfinch/finch/pkg/system"
 )
 
+// AdditionalDirectory represents the additional directory used in Finch config.
```
davidhsingyuchen, line 25: We can probably refine the comment according to the description of it in README as it may be a bit too vague now.

pkg/config/lima_config_applier_test.go
```diff
@@ -81,6 +85,34 @@ func TestDiskLimaConfigApplier_Apply(t *testing.T) {
...
+			mockSvc: func(fs afero.Fs, l *mocks.Logger) {
+				err := afero.WriteFile(fs, "/lima.yaml", []byte("memory: 4GiB\ncpus: 8"), 0o600)
+				require.NoError(t, err)
+			},
+			postRunCheck: func(t *testing.T, fs afero.Fs) {
+				buf, err := afero.ReadFile(fs, "/lima.yaml")
+				require.NoError(t, err)
+
+				// limayaml.LimaYAML has a required "images" field which will also get marshaled
+				wantYaml := `images: []
```
davidhsingyuchen, line 105: We can also get rid of the irrelevant `images: []`.

README.md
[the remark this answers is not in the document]
ningziwen: Makes sense. Added.

README.md
```diff
@@ -96,8 +96,9 @@ Finch has a simple and extensible configuration. A configuration file at `${HOME
 
 Currently, the options are:
 
-* CPUs [int]: the amount of vCPU to dedicate to the virtual machine
-* Memory [string]: the amount of memory to dedicate to the virtual machine
+* CPUs [int] (required): the amount of vCPU to dedicate to the virtual machine
+* Memory [string] (required): the amount of memory to dedicate to the virtual machine
+* AdditionalDirectories [AdditionalDirectory{path [string]}] (optional): the work directories that are not supported by default. In macOS, only home directory is supported by default.
```
davidhsingyuchen: Maybe we can follow [what lima is doing](https://github.com/lima-vm/lima/blob/master/examples/default.yaml) to put descriptions and example values of the fields together in a single `.yaml` file and provide an [entry point in README](https://github.com/lima-vm/lima#configuration):

- For each field, reading the description of it together with an example value may make it easier to follow.
- `[AdditionalDirectory{path [string]}]` implies that we may need to add a new field here whenever a new field is added to the `AdditionalDirectory` struct, which can be a bit hard to read in the end, and maybe directly looking at the example values is easier to comprehend the structure of it:

```
additional_directories:
  - path: /Volumes
```

pkg/config/config.go
ningziwen: AdditionalDirectory is just the element of the whole AdditionalDirectories instead of serving the complicated use case by itself. Does it really make sense of adding the use case of AdditionalDirectories here?

pkg/config/lima_config_applier_test.go
ningziwen: image: [] is a required field so it will be anyways marshaled even if we don't input it.

README.md
ningziwen: The README already has an example yaml inline. Having a separated yaml file is a good idea. Current config is still very simple with 3 fields so I think inline yaml should be sufficient. Maybe we can have a separated yaml when there are more fields coming.

davidhsingyuchen: > The README already has an example yaml inline. 

Yes, and I'm suggesting to merge that thing and the descriptions here into a single YAML file as I feel that separating them may decrease the readability according to the reasons stated above. Types like `[AdditionalDirectory{path [string]}]` don't really add much value, but we feel that we should provide it because currently the description and the example value are separated.

pkg/config/config.go
davidhsingyuchen: Then maybe at least add some comments to `AdditionalDirectories`? Currently looking at the field definitions basically give no info what they are used for, while we shouldn't need to trace their references only to get a high-level understanding of them.

pkg/config/lima_config_applier_test.go
davidhsingyuchen: My bad, I meant right now we're asserting the whole file, so we need to mention it, but if we unmarshal the  bytes into a struct and only check the fields we care about like what we're doing in e2e tests, then we don't have to mention it and explain why it's there as it's irrelevant to us in the first place:

```go
			err = yaml.Unmarshal(cfgBuf, &limaCfg)
			gomega.Expect(err).ShouldNot(gomega.HaveOccurred())
			gomega.Expect(*limaCfg.CPUs).Should(gomega.Equal(6))
			gomega.Expect(*limaCfg.Memory).Should(gomega.Equal("4GiB"))
```

e2e/config_test.go
```diff
@@ -107,7 +120,12 @@ var testConfig = func(o *option.Option, installed bool) {
 			gomega.Expect(limaConfigFilePath).Should(gomega.BeARegularFile())
 			cfgBuf, err := os.ReadFile(filepath.Clean(limaConfigFilePath))
 			gomega.Expect(err).ShouldNot(gomega.HaveOccurred())
-			gomega.Expect(cfgBuf).Should(gomega.SatisfyAll(gomega.MatchRegexp(`cpus: \d`), gomega.MatchRegexp(`memory: \dGiB`)))
+
+			var limaCfg limayaml.LimaYAML
+			err = yaml.Unmarshal(cfgBuf, &limaCfg)
+			gomega.Expect(err).ShouldNot(gomega.HaveOccurred())
+			gomega.Expect(reflect.ValueOf(limaCfg).FieldByName("CPUs").IsValid()).Should(gomega.BeTrue())
```
davidhsingyuchen, line 127: Just wondering, does the simplified version fail somehow?

```suggestion
			gomega.Expect(reflect.ValueOf(*limaCfg.CPUs).IsValid()).Should(gomega.BeTrue())
```

davidhsingyuchen (dismissed):

davidhsingyuchen (approved):

Issue #77, opened 2022-12-01
Persistent volume support


pendo324:
**What is the problem you're trying to solve?.**
Lima recently merged support for persistent data volumes, we should enable it in Finch by default so data can persist between different VM initializations.

**Describe the feature you'd like**
Add persistent volume creation and management to Finch, along with scripts that mount the containerd directories onto the volume.

**Additional context**

Issue #11, opened 2022-11-22
Suggestion: move `/Applications/Finch/lima/data` to the home directory (for multi-user setup)


AkihiroSuda:
Hi 👋, nice to see the release of Finch.
I'd suggest moving the `/Applications/Finch/lima/data` to the home directory for allowing multi-user setup.

e.g., `~/.finch-lima`

Note that the directory path should not be long, as the UNIX socket path is limited to 104 chars.
(So , `~/Library/Application Support/...` is not an appropriate path.)

PR #133, opened 2022-12-27
feat: saves containerd user data to a persistent disk

sam-berning:
this allows users to retain downloaded images, containers, etc. across new installations of finch

Issue #, if available: https://github.com/runfinch/finch/issues/77

*Description of changes:*

- Adds a new package for managing the user data persistent disk
- Checks disk configuration & creates disk if necessary on `finch vm init`
- Attaches disk whenever the VM starts

*Testing done:*
- Unit testing on `disk` and `cmd/finch`
- Manual testing

*Open questions:*
- Should we also check the configuration on `finch vm start`?

#### License Acceptance


ningziwen:
Could you add e2e tests?

ningziwen:
Could you this resolve this issue https://github.com/runfinch/finch/issues/11?

pkg/disk/disk.go
```diff
@@ -0,0 +1,137 @@
...
+	"io/fs"
+
+	"github.com/spf13/afero"
+
+	"github.com/runfinch/finch/pkg/command"
+	"github.com/runfinch/finch/pkg/path"
+)
+
+const (
+	diskName = "finch"
```
ningziwen, line 20: Maybe add a comment in both here and yaml file to say the disk name must be same in the two places.

sam-berning:
> Could this change resolve this issue #11?

I don't think so. We'd need to move the whole `.../lima/data` directory under the user's home directory which I think is outside the scope of this PR.

This current implementation _is_ a little awkward though, because saving the `datadisk` via a symlink into a specific user's home directory necessarily pins the installation to a single user. Overall I'm not sure if this makes things any worse, since many of the files under `.../lima/data` are already owned by the installing user, but it's interesting to note.

pkg/disk/disk.go
```diff
@@ -0,0 +1,138 @@
...
+	return &userDataDiskManager{
+		lcc:     lcc,
+		fs:      fs,
+		finch:   finch,
+		homeDir: homeDir,
+	}
+}
+
+// InitializeUserDataDisk checks the current disk configuration and fixes it if needed.
+func (m *userDataDiskManager) InitializeUserDataDisk() error {
```
pendo324, line 59: nit: might want to change this name to something that is more descriptive of the behavior. Without reading the code, I would expect that this just does first time setup rather than also fixing existing disks/configurations. If you agree, maybe something like `EnsureUserDataDisk` would be more appropriate?

pkg/disk/disk.go
```diff
@@ -0,0 +1,138 @@
...
+	if m.limaDiskExists() {
+		limaPath := fmt.Sprintf("%s/_disks/%s/datadisk", m.finch.LimaHomePath(), diskName)
+		loc, err := m.fs.ReadlinkIfPossible(limaPath)
+		if err != nil {
+			return err
+		}
+		// if the file is not m symlink, loc will be an empty string
```
pendo324, line 66: nit: typo? Should this read: if the file is not ~~m~~ **a** symlink? I might just be understanding it wrong

pkg/disk/limactl_disk.go
```diff
@@ -0,0 +1,13 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+package disk
+
+type limactlDiskListOutput struct {
```
pendo324, line 6: Maybe we can add a comment with some reasoning why we can't import this from [lima-vm/pkg/store/disk](https://github.com/lima-vm/lima/blob/97ac8caa9a0b5e7b3d664d5e1c73b0e0b710ac75/pkg/store/disk.go#L14-L21)

pendo324 (commented):
LGTM for the most part, just left some questions / nits

pkg/disk/disk.go
sam-berning: yeah, should be **a** symlink. I had previously used `a` as a temporary variable name and I guess my IDE got a little overzealous when I renamed the variable 😆

pkg/disk/limactl_disk.go
sam-berning: good point, I actually overlooked this option. I think importing it from the original lima struct is better since we won't need to update it each time a field is added upstream in lima

pendo324:
Very nice writeup on the investigation, a lot of people would just dismiss the integration test failures as intermittent. I tried to respond, roughly in order of importance:

> The second takeaway also goes back to the open question on this PR -- should we be ensuring the disk only on init or on both init and start? I'm leaning more towards both, just to make it as painless as possible if a user were to manually make some change to the user data disk setup.
I think I agree with this. It's more work for the CLI, but much or all of the code can probably be reused. It should in theory reduce the chance of errors for Finch users.

> On that topic, is the current location of the datadisk good? Currently, the only other file under ~/.finch is finch.yaml which we teach users to edit. Since it's best for users to leave datadisk alone, should we put it somewhere else? Perhaps under another .directory in ~/.finch?

I think I'm good with moving it under a new hidden folder under `~/.finch`, just to avoid accidental edits. Maybe just `~/.finch/.disks` or something?

> In Lima, should a stopped instance holding a lock for a disk fail to start up when trying to attach the same disk? It feels like no, but this is a very strange edge case.

Maybe we can open an issue on the Lima repo that describes this edge case to see if they want to leave it as is, or add some unlocking mechanism when the same instance is re-using a disk from a "dirty shutdown"?

> Of course, there's the question as to how it can take >60 seconds for the M1 11.7 runner to stop the VM. Are we running
too many tests in parallel? Or is finch vm stop really that slow on M1 macs?

This was probably a rare occurence that will likely eventually happen again, given the volume that we run the integration tests at. We can probably add some more [cleanup code here](https://github.com/runfinch/finch/blob/main/.github/workflows/ci.yaml#L93-L103), and remove it after you implement the init/start check, or it gets changed in Lima itself?

davidhsingyuchen:
> Of course, there's the question as to how it can take >60 seconds for the M1 11.7 runner to stop the VM. Are we running
too many tests in parallel? Or is finch vm stop really that slow on M1 macs?

FWIW, currently e2e tests are run sequentially: https://github.com/onsi/ginkgo/blob/e7e3db7a3ee2fa40d21931f4f98f0c71c34ed3eb/types/config.go#L46

cmd/finch/main.go
```diff
@@ -108,6 +110,7 @@ func virtualMachineCommands(
 		config.NewNerdctlApplier(fssh.NewDialer(), fs, fp.LimaSSHPrivateKeyPath(), system.NewStdLib()),
 		fp,
 		fs,
+		disk.NewUserDataDiskManager(lcc, &afero.OsFs{}, fp, system.NewStdLib().Env("HOME")),
```
pendo324, line 113: nit: any reason to not just pass `fs` instead of a new `afero.OsFs{}`?

pendo324 (commented):
LGTM, just one more nit

cmd/finch/main.go
sam-berning: Because the persistent disk code needs to manage symlinks, but the generic `Fs` interface doesn't provide that. I made the `diskFS` interface in `pkg/disk/disk.go` for this purpose, but most other parts of finch can't use that because we want to be able to unit test them without interacting with the filesystem so we use `MemMapFs` which does not have symlinks implemented.

I could export the `diskFS` interface and pass that from main instead of an `Fs` but since nowhere else in finch currently needs it I felt it made the most sense to just instantiate a new `OsFs` here (especially since it doesn't store any state)

pendo324: Makes sense, thanks for the explanation!

pendo324 (approved):

PR #136, opened 2023-01-05
ci: automate step to update finch-core dependency in makefile

vsiravar:
Signed-off-by: Vishwas Siravara <siravara@amazon.com>

Issue #, if available:

*Description of changes:*
Historically releasing `finch` requires us to manually update the `finch-core` reference in the [Makefile](https://github.com/runfinch/finch/blob/main/Makefile#L24). This process is error prone and time consuming. This change automates this step to a certain degree and does not require the user to modify the Makefile manually. This change will eventually be triggered by a release in finch-core, current trigger is `workflow_dispatch` and the action `Update dependencies` has to be manually triggered before the release of finch. 
*Testing done:*
Manually executing the steps in the workflow on a remote server. 

#### License Acceptance


Makefile
```diff
@@ -20,8 +20,7 @@ VDE_INSTALL ?= /opt/finch
 UNAME := $(shell uname -m)
 ARCH ?= $(UNAME)
 SUPPORTED_ARCH = false
-CORE_VERSION ?= 0.1.1
-CORE_URL ?= https://artifact.runfinch.com/finch-core-$(CORE_VERSION).tar.gz
+CORE_URL ?= https://artifact.runfinch.com/finch-core-0.1.1.tar.gz
```
ningziwen, line 23: We can still keep this env for easy core version bump for local development and test?

vsiravar: https://github.com/runfinch/finch/blob/ba12834325006e60f37939a0f819bfe3686b5461/.github/bin/update-lima-dep.sh#L38 fetches the latest `finch-core` from s3 which would be of the format finch-core-<version>.tar.gz . It's possible to extract out the version from here but it may be cumbersome in case the format of the filename changes or renamed in s3. If the user wants to change the version it's still changing only in one place although not as convenient as having an env for just the version.  wdyt?

pendo324: Users can still pass their own `CORE_URL` variable, instead of just `CORE_VERSION`. This is the only place `CORE_VERSION` was used anyway, so the only difference is now users have to specify the entire URL instead of just the version. That's how I usually used CORE_URL anyway, since it works with `file://` URLs (its usually easier to use a local file URL in-dev than an internet/S3 URL anyway).

vsiravar: Yeah so this way users can override the `CORE_URL` by `make <target> CORE_URL=<custom url>`. So we don't really need the `CORE_VERSION` variable.

pendo324 (approved):
LGTM, but please resolve the comments before merging.

PR #146, opened 2023-01-11
refactor: break json into multiple lines in version unit test

davidhsingyuchen:
## Summary

PR makes the JSON object more readable.

## License Acceptance


ahsan-z-khan (approved):

ahsan-z-khan:
I think #146 #147 and #145 be same refactor PR. Why do we need 3?

davidhsingyuchen:
> I think #146 #147 and #145 be same refactor PR. Why do we need 3?

It's better to make each PR focused (i.e., each PR should only do one thing). Will be easier to review and revert if needed.

PR #150, opened 2023-01-12
chore: split e2e tests into 2 packages, vm and container

davidhsingyuchen:
## Summary

PR fixes https://github.com/runfinch/finch/pull/142#issuecomment-1379866441 by splitting container tests and vm tests into different binaries (i.e., different packages) to make sure that the former is run before the latter.

The package-level comment of the `e2e` package provides more details, so reading it first may make reviewing this PR easier.

## License Acceptance

### DIFF
--- a/e2e/container/container_test.go
+++ b/e2e/container/container_test.go
@@ -1,56 +1,35 @@
 // Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 // SPDX-License-Identifier: Apache-2.0
 
-package e2e
+// Package container runs tests related to container development.
+package container
 
 import (
-	"flag"
-	"os"
-	"path/filepath"
 	"testing"
 
 	"github.com/onsi/ginkgo/v2"
 	"github.com/onsi/gomega"
 	"github.com/runfinch/common-tests/command"
-	"github.com/runfinch/common-tests/option"
 	"github.com/runfinch/common-tests/tests"
+
+	"github.com/runfinch/finch/e2e"
 )
 
-const (
-	virtualMachineRootCmd = "vm"
-	installedTestSubject  = "finch"
-)
+//nolint:paralleltest // TestContainer is like TestMain for the container-related tests.
+func TestContainer(t *testing.T) {
+	const description = "Finch Container Development E2E Tests"
 
-var installed = flag.Bool("installed", false, "the flag to show whether the tests are ran against installed application")
-
-//nolint:paralleltest // TestE2e is like TestMain for our e2e tests.
-func TestE2e(t *testing.T) {
-	description := "Finch CLI e2e Tests"
-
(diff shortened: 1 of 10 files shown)


pendo324 (commented):
The split makes a lot of sense after reading the package level comments, you did a great job citing the Ginkgo docs. But I wonder if `vm` is the best name for the package, since it also contains test for config and the version command. Does something like `container` and `app` make sense? Or maybe `container` and `cli` or `common` and `cli`? Definitely open to discussion here

e2e/container/container_test.go
```diff
@@ -1,57 +1,36 @@
...
+	"github.com/runfinch/finch/e2e"
 )
 
-var installed = flag.Bool("installed", false, "the flag to show whether the tests are ran against installed application")
-
-//nolint:paralleltest // TestE2e is like TestMain for our e2e tests.
-func TestE2e(t *testing.T) {
-	description := "Finch CLI e2e Tests"
+//nolint:paralleltest // TestContainer is like TestMain for the container-related tests.
+func TestContainer(t *testing.T) {
```
azhouwd, line 19: How would we guarantee `TestContainer` runs before `TestVM`?

davidhsingyuchen:
> The split makes a lot of sense after reading the package level comments, you did a great job citing the Ginkgo docs.

Thanks. Updated the PR description to recommend reviewers to read the package-level comments first.

> But I wonder if vm is the best name for the package, since it also contains test for config and the version command. Does something like container and app make sense? Or maybe container and cli or common and cli? Definitely open to discussion here

I named it `vm` because the tests there are more concerned about the VM itself instead of the container development functionalities it provides. Open to other names though.

e2e/container/container_test.go
davidhsingyuchen: @pendo324's understanding is correct. They will be separated into 2 binaries, so the execution order is guaranteed.

azhouwd: Sorry still a bit confused why the execution order is guaranteed? Like `go test ./e2e/container` then `go test ./e2e/vm`

davidhsingyuchen: > why the execution order is guaranteed?

In `Makefile`:

1. We first run `go test ./e2e/container`, which only compiles and runs the tests defined in the `e2e/container` package, and that means only the container tests are included.
3. We then run `go test ./e2e/vm` to do the same for the VM tests.

azhouwd: oh sorry my bad, didn't realize the change in the Makefile.

azhouwd: Actually the name makes me think in a way that unit tests are triggered

azhouwd (approved):

pendo324 (approved):
I'm good with approving it with these names, we can change them later if we come up with something better

Issue #140, opened 2023-01-09
e2e tests failing intermittently


sam-berning:
**Describe the bug**
During the process of reviewing and merging https://github.com/runfinch/finch/pull/133, I noticed intermittent and unpredictable failures in the e2e tests. I addressed some of these issues [in the PR itself](https://github.com/runfinch/finch/pull/133#issuecomment-1371516639), but some of the failures persist. In the repository's current state, I've seen two types of failures:

1. **`finch vm stop` timeouts**
Occasionally, `finch vm stop` takes longer than 60 seconds to execute. I see such a timeout happen in about every other run of the e2e tests. The tests run this command ~20 times, and we run the e2e tests on 4 different hosts each test run. So it seems to happen on about 1% of stop commands.
Because the tests assume that the vm is in a particular state before the test starts, this can also cause carry-over effects to other tests, which results in multiple tests failing.
1. **issues with BuildKit starting up**
I've only seen this one time so far, but it didn't seem caused by a `finch vm stop` timeout. Below are the relevant logs:
```
time="2023-01-07T01:17:37Z" level=error msg="`buildctl` needs to be installed and `buildkitd` needs to be running, see https://github.com/moby/buildkit , and `containerd-rootless-setuptool.sh install-buildkit` for OCI worker or `containerd-rootless-setuptool.sh install-buildkit-containerd` for containerd worker" error="3 errors occurred:\n\t* failed to ping to host unix:///run/user/501/buildkit-finch/buildkitd.sock: exit status 1\n\t* failed to ping to host unix:///run/user/501/buildkit-default/buildkitd.sock: exit status 1\n\t* failed to ping to host unix:///run/user/501/buildkit/buildkitd.sock: exit status 1\n\n"
  time="2023-01-07T01:17:37Z" level=fatal msg="no buildkit host is available, tried 3 candidates: 3 errors occurred:\n\t* failed to ping to host unix:///run/user/501/buildkit-finch/buildkitd.sock: exit status 1\n\t* failed to ping to host unix:///run/user/501/buildkit-default/buildkitd.sock: exit status 1\n\t* failed to ping to host unix:///run/user/501/buildkit/buildkitd.sock: exit status 1\n\n"
```

The test failures seem to happen more often on ARM hosts than x86 hosts. The ARM hosts running OSX 11.7 in particular seem to have issues with `finch vm stop` timeouts.

**Steps to reproduce**
Run e2e tests

**Expected behavior**
Tests should succeed consistently

[...]

PR #154, opened 2023-01-12
fix: changes persistent disk path to prevent collisions

sam-berning:

Issue #, if available:

*Description of changes:*

It was possible for two different finch VMs in different locations to use the same persistent disk, causing the disk to corrupt. This changes the path to a sha256 sum of the installation location.

*Testing done:*

Unit testing, manual testing to confirm the disk is created in the right location.

#### License Acceptance

### DIFF
--- a/pkg/disk/disk.go
+++ b/pkg/disk/disk.go
@@ -9,6 +9,7 @@ import (
 	"errors"
 	"fmt"
 	"io/fs"
+	"os"
 	"path"
 
 	limaStore "github.com/lima-vm/lima/pkg/store"
@@ -121,7 +122,15 @@ func (m *userDataDiskManager) createLimaDisk() error {
 func (m *userDataDiskManager) attachPersistentDiskToLimaDisk() error {
 	limaPath := fmt.Sprintf("%s/_disks/%s/datadisk", m.finch.LimaHomePath(), diskName)
 	if !m.persistentDiskExists() {
-		err := m.fs.Rename(limaPath, m.finch.UserDataDiskPath(m.homeDir))
+		disksDir := path.Dir(m.finch.UserDataDiskPath(m.homeDir))
+		_, err := m.fs.Stat(disksDir)
+		if os.IsNotExist(err) {
+			err := m.fs.MkdirAll(disksDir, 0o755)
+			if err != nil {
+				return fmt.Errorf("could not create persistent disk directory: %w", err)
+			}
+		}
+		err = m.fs.Rename(limaPath, m.finch.UserDataDiskPath(m.homeDir))
 		if err != nil {
 			return err
 		}
--- a/pkg/path/finch.go
+++ b/pkg/path/finch.go
@@ -5,6 +5,7 @@
 package path
 
 import (
+	"crypto/sha256"
 	"fmt"
 
 	"github.com/runfinch/finch/pkg/system"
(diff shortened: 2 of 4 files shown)


weikequ:
Seems like we're still getting a vm stop timeout failure. Did you end up changing the timeout?

sam-berning:
> Seems like we're still getting a vm stop timeout failure. Did you end up changing the timeout?

Yeah, we upped it to 60 seconds. Not sure if continuing to increase the timeout makes sense, especially since for the most part `finch vm stop` takes around 5 seconds to complete. We're tracking the issue here: https://github.com/runfinch/finch/issues/140

weikequ:
Yeah overall the code LGTM, but still getting some test failures for some (hopefully) unrelated issues.

weikequ (dismissed):

pkg/path/finch.go
```diff
@@ -60,6 +61,11 @@ func (w Finch) LimaSSHPrivateKeyPath() string {
 	return fmt.Sprintf("%s/lima/data/_config/user", w)
 }
 
+func (w Finch) generatePathSum() string {
```
pendo324, line 64: So the install location in this case is the directory two levels up from the finch executable, right? I think this works for now, but we might have to revisit this to somehow include the instance name if we ever see a usecase for multiple finch instances (unless we want them to share their data disks?)

pkg/disk/disk.go
```diff
@@ -121,7 +122,15 @@ func (m *userDataDiskManager) createLimaDisk() error {
 func (m *userDataDiskManager) attachPersistentDiskToLimaDisk() error {
 	limaPath := fmt.Sprintf("%s/_disks/%s/datadisk", m.finch.LimaHomePath(), diskName)
 	if !m.persistentDiskExists() {
-		err := m.fs.Rename(limaPath, m.finch.UserDataDiskPath(m.homeDir))
+		disksDir := path.Dir(m.finch.UserDataDiskPath(m.homeDir))
```
pendo324, line 125: This new logic should handle the case where the dir doesn't exist and migrating an old datadisk to the new name right?

pkg/path/finch.go
sam-berning: > we might have to revisit this to somehow include the instance name if we ever see a usecase for multiple finch instances

yeah, I figure we'll need to overhaul the `path` package pretty significantly in order to do that anyway. the disk name is pretty arbitrary though so I might as well do it based off of `.../lima/data/finch` to "future-proof" it

> unless we want them to share their data disks?

this won't be possible based on the current implementation of persistent disks, if two instances try to write to the same qcow2 disk at the same time the disk will corrupt and very strange things happen. probably the best we could do is copy the other instance's disk at `finch vm init`

pkg/disk/disk.go
sam-berning: it does handle the case where the dir doesn't exist, yeah.

> migrating an old datadisk to the new name

this isn't included currently. I'm a little torn on whether to include it. perhaps we can put a notice on the release that you will lose your data upgrading from 0.2.0 to 0.2.1, with manual steps on how to restore it?

pendo324: Yeah, I think that makes sense. We don't really want to carry that migration code. Not sure if that means it needs to be 0.3.0 instead of 0.2.1, maybe @ningziwen can chime in?

ningziwen: I agree. It should be sufficient if we can put a notice in the release note. 

As this is a fix change, I guess it should be 0.2.1?

pendo324 (approved):

PR #158, opened 2023-01-14
fix: correctly handle local environment value pass-through

### LINKED ISSUE #35: Env variables passing not compatible with docker command
**What is the problem you're trying to solve?.**
I use finch as a docker alias
```
alias docker=finch
```

Env variables in Docker can be passed just with the name and then the value will be passed automatically
```
❯ export FOO=bar
❯ docker run -e FOO ubuntu env
HOME=/root
FOO=bar
HOSTNAME=809ed84fc40f
```

But with finch I have to pass the value explicitly. In the output the FOO variable is missing

```
❯ alias docker=finch
❯ export FOO=bar
❯ docker run -e FOO ubuntu env
HOME=/root
```

**Describe the feature you'd like**
Allow passing variables with the name only, so that `finch run` is compatible to the `docker run` command and can be used as a drop-in replacement

estesp:
The `-e`,`--env` flag as well as values within an `--env-file` may contain only a variable name with no "={VALUE}" portion. These entries provide a shorthand to say "pass the existing environment value of this variable into the container". Because of the VM boundary we need to extrapolate these values on the Finch side and pass them as discrete values into the Lima VM on the nerdctl command line. Also the file referenced by `--env-file` may not be accessible inside the VM, so we translate each entry in the file into `-e` entries on the command line rather than fail on the VM side, unable to locate the file being referenced.

Signed-off-by: Phil Estes <estesp@amazon.com>

Issue #, if available: Fixes #35

#### License Acceptance


QuinnyPig:
Confirmed that this works (once #154 is cherry-picked into the PR branch) in my testing.

estesp:
Rebased on `main` to make it easier for testing since this commit was between the persistent disk feature and the fix for unique persistent disk naming.

ningziwen:
The tests can be added here:

[Unit tests](https://github.com/runfinch/finch/blob/main/cmd/finch/nerdctl_test.go#L151)

[e2e tests](https://github.com/runfinch/common-tests/blob/main/tests/run.go)

ningziwen:
The existing e2e tests expect the local env value is missing in the result. https://github.com/runfinch/common-tests/blame/main/tests/run.go#L130

So this may be a feat instead of fix as this is an expected unimplemented feature based on e2e tests, assuming e2e tests can be the workaround to check if one behaviour is expected given the situation where we don't have unimplement feature documentation.

estesp:
> The existing e2e tests expect the local env value is missing in the result. https://github.com/runfinch/common-tests/blame/main/tests/run.go#L130

That test passes even though the core nerdctl problem (of `-eVAR` being unimplemented) [was fixed](https://github.com/containerd/nerdctl/pull/1524) since Finch was launched. It is almost an inadvertent pass, though, because that fix (since nerdctl v1.1.0) was not carried into the Finch side because of the VM boundary issue. To align with nerdctl function, we should have fixed the same bug in Finch when we included v1.1.0.

Of course, the other test framework issue is that to really test the exact behavior you need -eVAR1 where VAR1 is unset in the environment and -eVAR2 where VAR2 **is** set in the environment. I will update this PR with a modification in the test to validate both variations as that is the correct behavior from both Docker, and now nerdctl, since v1.1.0.

> 
> So this may be a feat instead of fix as this is an expected unimplemented feature based on e2e tests, assuming e2e tests can be the workaround to check if one behaviour is expected given the situation where we don't have unimplement feature documentation.

This seems like an expectation that the e2e tests are based on concrete list of known functionality (versus just how nerdctl worked when the test was written). I'm OK if we would rather call this a feature, but I don't think the e2e tests are the defacto standard on specific capabilities that were inadvertently broken at the time.

ningziwen:
Released the new version of common-tests. You can bump the version of common-tests and use it in this PR. https://github.com/runfinch/common-tests/releases

estesp:
Updated common-tests; ran locally to verify e2e tests:
```
Ran 274 of 274 Specs in 610.032 seconds
SUCCESS! -- 274 Passed | 0 Failed | 0 Pending | 0 Skipped
```

ningziwen:
@estesp Thanks. Could you please also add unit tests? https://github.com/runfinch/finch/pull/158#issuecomment-1386172980

`make coverage` can be used to self-validate unit test coverage.

estesp:
@ningziwen added unit tests; testing env-file gets a bit messy as we need an input file to get coverage over the file reading/parsing code blocks. Please advise if you want to use another way other than tmpfile creation. Also had to break it out of the test table as the flag name won't be known until the tmpfile is created.

cmd/finch/nerdctl.go
```diff
@@ -59,18 +63,59 @@ func (nc *nerdctlCommand) run(cmdName string, args []string) error {
...
+	for _, e := range fileEnvs {
+		evar, eval, _ := strings.Cut(e, "=")
+		envVars[evar] = eval
+	}
+	for _, e := range envs {
+		evar, eval, _ := strings.Cut(e, "=")
+		envVars[evar] = eval
+	}
+
+	var finalArgs []string
```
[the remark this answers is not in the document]
estesp: we can't append to nerdctl args as this command ends with an image reference and optionally a command to run; given we don't know the shape of the rest of the flags, we are essentially prepending `nerdctlArgs` with all the `-e ` entries to not break the existing command structure. While we could keep re-creating the array each time through the loop it seems simpler to just build up our own array and then append the existing `nerdctlArgs` on the end.

cmd/finch/nerdctl.go
```diff
@@ -122,6 +167,93 @@ func (nc *nerdctlCommand) shouldReplaceForHelp(cmdName string, args []string) bo
...
 }
 
+func argIsEnv(arg string) bool {
+	if strings.HasPrefix(arg, "-e") || (strings.HasPrefix(arg, "--env") && !strings.HasPrefix(arg, "--env-file")) {
+		return true
+	}
+	return false
+}
+
+func handleEnv(arg, arg2 string) (bool, []string) {
```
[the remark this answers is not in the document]
estesp: Fixed

cmd/finch/nerdctl.go
```diff
@@ -59,18 +63,59 @@ func (nc *nerdctlCommand) run(cmdName string, args []string) error {
...
+				return err
+			}
+			skip = shouldSkip
+			fileEnvs = append(fileEnvs, addEnvs...)
+		default:
+			nerdctlArgs = append(nerdctlArgs, arg)
 		}
-		nerdctlArgs = append(nerdctlArgs, arg)
 	}
+	// to handle environment variables properly, we add all entries found via
```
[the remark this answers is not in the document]
estesp: Done

cmd/finch/nerdctl.go
```diff
@@ -122,6 +167,93 @@ func (nc *nerdctlCommand) shouldReplaceForHelp(cmdName string, args []string) bo
...
+		// flag and value are in the same string
+		if strings.HasPrefix(arg, "-e") {
+			envVar = arg[2:]
+		} else {
+			// only other case is "--env="; skip that prefix
+			envVar = arg[6:]
+		}
+	}
+
+	if !strings.Contains(envVar, "=") {
```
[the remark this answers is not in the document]
estesp: No longer need the array; reworked this as we are only returning whether we have a fully formed env. variable and value.

cmd/finch/nerdctl.go
```diff
@@ -122,6 +167,93 @@ func (nc *nerdctlCommand) shouldReplaceForHelp(cmdName string, args []string) bo
...
+	}
+
+	file, err := os.Open(filepath.Clean(filename))
+	if err != nil {
+		return false, []string{}, err
+	}
+	defer file.Close() //nolint:errcheck,gosec // close of a file in O_RDONLY mode has no gosec issue
+
+	stdlib := system.NewStdLib()
+	scanner := bufio.NewScanner(file)
```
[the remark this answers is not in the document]
estesp: Done

cmd/finch/nerdctl.go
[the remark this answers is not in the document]
davidhsingyuchen: Ah my bad, thanks for the explanation!

pkg/system/stdlib.go
```diff
@@ -38,6 +38,10 @@ func (s *StdLib) Env(key string) string {
 	return os.Getenv(key)
 }
 
+func (s *StdLib) LookupEnv(key string) (string, bool) {
```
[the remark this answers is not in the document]
estesp: Added a mock for the systemdeps; added a unit test which tests a return value from `LookupEnv()`; this should be complete; let me know if it looks reasonable.

pkg/command/nerdctl.go
```diff
@@ -0,0 +1,13 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+package command
+
+import "github.com/runfinch/finch/pkg/system"
+
+// NerdctlCommandSystemDeps contains the system dependencies for newNerdctlCommand.
+//
+//go:generate mockgen -copyright_file=../../copyright_header -destination=../../pkg/mocks/nerdctl_cmd_system_deps.go -package=mocks -mock_names NerdctlCommandSystemDeps=NerdctlCommandSystemDeps . NerdctlCommandSystemDeps
+type NerdctlCommandSystemDeps interface {
```
davidhsingyuchen, line 11: Maybe we can move the type definition to `cmd/finch/nerdctl.go` (i.e., where it's used) ([example](https://github.com/runfinch/finch/blob/ff8a35f4ec238abeafb8a1ea7050e844b1996fba/pkg/path/finch.go#L77))

pkg/system/system.go
```diff
@@ -47,6 +47,11 @@ type EnvGetter interface {
 	Env(key string) string
 }
 
+// LookupEnvGetter mocks out os.Getenv.
+type LookupEnvGetter interface {
```
davidhsingyuchen, line 51: nit: `LookupEnvGetter` may sound like `LookupEnv` is a thing, and this interface can `Get` it, but it's not the case

```suggestion
type EnvLookuper interface {
```

cmd/finch/nerdctl.go
```diff
@@ -122,6 +174,92 @@ func (nc *nerdctlCommand) shouldReplaceForHelp(cmdName string, args []string) bo
...
+		}
+		switch {
+		case strings.HasPrefix(line, "#"):
+			// ignore comments
+			continue
+		case !strings.Contains(line, "="):
+			// only has the variable name; need to lookup value
+			if val, ok := systemDeps.LookupEnv(line); ok {
+				envs = append(envs, fmt.Sprintf("%s=%s", line, val))
+			}
```
davidhsingyuchen, line 251: Maybe we can also update the `with --env-file flag replacement` test case to cover this block.

pkg/command/nerdctl.go
estesp: that was my original idea, but `go generate` + `mockgen` errors out if operating on non-importable (`package main`) files; e.g. outputs an error about not working with programs, only modules/libraries. If I misunderstood the `go generate`/mockgen errors and someone knows the trick, let me know.

davidhsingyuchen: We can use the [source mode](https://github.com/golang/mock#source-mode) of `mockgen`.

After moving it to `cmd/finch/nerdctl.go`, we should only need to update the `go generate` comment to be:

```go
mockgen -copyright_file=../../copyright_header -destination=../../pkg/mocks/nerdctl_cmd_system_deps.go -package=mocks -mock_names NerdctlCommandSystemDeps=NerdctlCommandSystemDeps -source=nerdctl.go NerdctlCommandSystemDeps
```

To be specific, `.` is replaced with `-source=nerdctl.go`. Tested locally, and please let me know if it does not work for you, thanks!

cmd/finch/nerdctl_test.go
```diff
@@ -80,14 +82,14 @@ func TestNerdctlCommand_run(t *testing.T) {
...
-		mockSvc func(*mocks.LimaCmdCreator, *mocks.Logger, *gomock.Controller)
+		mockSvc func(*mocks.LimaCmdCreator, *mocks.NerdctlCommandSystemDeps, *mocks.Logger, *gomock.Controller)
 	}{
 		{
 			name:    "happy path",
 			cmdName: "build",
 			args:    []string{"-t", "demo", "."},
 			wantErr: nil,
-			mockSvc: func(lcc *mocks.LimaCmdCreator, logger *mocks.Logger, ctrl *gomock.Controller) {
+			mockSvc: func(lcc *mocks.LimaCmdCreator, nsd *mocks.NerdctlCommandSystemDeps, logger *mocks.Logger, ctrl *gomock.Controller) {
```
davidhsingyuchen, line 92: nit: maybe ncsd to make it consistent with its full name

```suggestion
			mockSvc: func(lcc *mocks.LimaCmdCreator, ncsd *mocks.NerdctlCommandSystemDeps, logger *mocks.Logger, ctrl *gomock.Controller) {
```

cmd/finch/nerdctl.go
estesp: Added a test for this branch now; note that I can't do checks for existing in host environment and command-line set in the same test because they are resolved into a map which doesn't preserve ordering; so the expected final command line can't be predicted. I don't think that's a problem, but if anyone reviewing wonders why they are separated into different tests.

pkg/system/system.go
estesp: Given "Lookuper" isn't a word I chose a different name, but is updated now.

pkg/command/nerdctl.go
estesp: Thanks! Updated

cmd/finch/nerdctl_test.go
estesp: Updated.

cmd/finch/nerdctl.go
```diff
@@ -122,6 +182,92 @@ func (nc *nerdctlCommand) shouldReplaceForHelp(cmdName string, args []string) bo
...
+		filename string
+		skip     bool
+	)
+
+	switch arg {
+	case "--env-file":
+		skip = true
+		filename = arg2
+	default:
+		filename = arg[11:]
```
davidhsingyuchen, line 234: Maybe we can also test this block. Usually we want one test case to only test one aspect, so ideally we would add a new test case for this, but I'm also fine with updating either of the existing 2 test cases for `--env-file`, up to you :)

estesp: Done

cmd/finch/nerdctl.go
```diff
@@ -122,6 +167,93 @@ func (nc *nerdctlCommand) shouldReplaceForHelp(cmdName string, args []string) bo
...
+	}
+
+	file, err := os.Open(filepath.Clean(filename))
```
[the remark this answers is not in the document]
estesp: personally I don't see how this adds any value at this point as test coverage is solid and there is no added value to this capability working or not working with more mocking around the handling of a simple file. If the maintainers feel strongly about it maybe it can be done in a follow-on improvement?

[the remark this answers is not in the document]
davidhsingyuchen: > how this adds any value at this point

Mainly mitigate https://github.com/runfinch/finch/pull/158#issuecomment-1413129951 and make the file-handling code consistent.

> maybe it can be done in a follow-on improvement?

Sure, I can do a follow-up PR.

davidhsingyuchen (approved):

Issue #157, opened 2023-01-13
Show VM creation progress or warn user it could take a while


otterley:
It took 3 minutes to launch the initial VM on my Macbook Pro with M1 Max (32GB):

```
INFO[0000] Using default values due to missing config file at "/Users/xxxx/.finch/finch.yaml" 
INFO[0000] "/Users/xxxx/.finch" directory doesn't exist, attempting to create it 
INFO[0000] binaries directory doesn't exist             
INFO[0000] Requesting root access to finish network dependency configuration 
Password:
INFO[0006] sudoers file not found: open /etc/sudoers.d/finch-lima: no such file or directory 
INFO[0008] Initializing and starting Finch virtual machine... 

<hmm, what is happening here...>

INFO[0187] Finch virtual machine started successfully   
```

You might want to give the user a some sort of warning about that - or, even better, display some sort of progress indicator during the creation process.

PR #161, opened 2023-01-16
fix: Show in-progress spinner when creating VM

hariohmprasath:
Issue: https://github.com/runfinch/finch/issues/157

Signed-off-by: Hari Ohm Prasth Rajagopal harrajag@amazon.com

Issue #, if available:
#157 

*Description of changes:*
Show spinner when VM creation is in progress

*Testing done:*
Ran local tests by building the code using make command and tested it by creating a new VM finch vm init command

```
INFO[0000] Initializing and starting Finch virtual machine...
⣾VM creation in progress
```

#### License Acceptance

### DIFF
--- a/cmd/finch/virtual_machine_init.go
+++ b/cmd/finch/virtual_machine_init.go
@@ -5,6 +5,9 @@ package main
 
 import (
 	"fmt"
+	"time"
+
+	"github.com/briandowns/spinner"
 
 	"github.com/runfinch/finch/pkg/disk"
 
@@ -94,6 +97,11 @@ func (iva *initVMAction) run() error {
 	instanceName := fmt.Sprintf("--name=%v", limaInstanceName)
 	limaCmd := iva.creator.CreateWithoutStdio("start", instanceName, iva.baseYamlFilePath, "--tty=false")
 	iva.logger.Info("Initializing and starting Finch virtual machine...")
+	// Show spinner
+	s := spinner.New(spinner.CharSets[11], 100*time.Millisecond)
+	s.Suffix = "VM creation in progress"
+	s.Start()
+	defer s.Stop()
 	logs, err := limaCmd.CombinedOutput()
 	if err != nil {
 		iva.logger.Errorf("Finch virtual machine failed to start, debug logs: %s", logs)
--- a/go.mod
+++ b/go.mod
@@ -3,6 +3,7 @@ module github.com/runfinch/finch
 go 1.18
 
 require (
+	github.com/briandowns/spinner v1.20.0
 	github.com/golang/mock v1.6.0
 	github.com/google/go-licenses v1.5.0
 	github.com/lima-vm/lima v0.14.2
@@ -35,6 +36,7 @@ require (
 	github.com/pkg/errors v0.9.1 // indirect
 	github.com/u-root/uio v0.0.0-20210528114334-82958018845c // indirect
 	golang.org/x/sync v0.1.0 // indirect
+	golang.org/x/term v0.4.0 // indirect
 	golang.org/x/time v0.0.0-20210723032227-1f47c861a9ac // indirect
 	google.golang.org/genproto v0.0.0-20220815135757-37a418bb8959 // indirect
 	google.golang.org/grpc v1.48.0 // indirect
@@ -58,7 +60,7 @@ require (
 	github.com/kevinburke/ssh_config v0.0.0-20190725054713-01f96b0aa0cd // indirect
 	github.com/kr/fs v0.1.0 // indirect
 	github.com/mattn/go-colorable v0.1.13 // indirect
-	github.com/mattn/go-isatty v0.0.16 // indirect
+	github.com/mattn/go-isatty v0.0.17 // indirect
 	github.com/mitchellh/go-homedir v1.1.0 // indirect
 	github.com/opencontainers/go-digest v1.0.0 // indirect
 	github.com/otiai10/copy v1.6.0 // indirect
--- a/go.sum
+++ b/go.sum
@@ -96,6 +96,8 @@ github.com/bazelbuild/rules_go v0.30.0/go.mod h1:MC23Dc/wkXEyk3Wpq6lCqz0ZAYOZDw2
 github.com/beorn7/perks v0.0.0-20180321164747-3a771d992973/go.mod h1:Dwedo/Wpr24TaqPxmxbtue+5NUziq4I4S80YR8gNf3Q=
 github.com/beorn7/perks v1.0.0/go.mod h1:KWe93zE9D1o94FZ5RNwFwVgaQK1VOXiVxmqh+CedLV8=
 github.com/bits-and-blooms/bitset v1.2.0/go.mod h1:gIdJ4wp64HaoK2YrL1Q5/N7Y16edYb8uY+O0FJTyyDA=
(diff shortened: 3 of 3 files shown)


davidhsingyuchen (commented):
Thanks for the PR! I think we can merge it as it's definitely an improvement, but I'm not 100% sure if #157 can be closed by it as strictly speaking, we're not showing the "progress" (i.e., 30%, 50%, etc.). @ningziwen WDYT?

LGTM overall, just please fix the failed checks. You can ignore the e2e ones as those failures are caused by flakiness and are tracked in #140.

cmd/finch/virtual_machine_init.go
```diff
@@ -94,11 +95,17 @@ func (iva *initVMAction) run() error {
...
+	s := spinner.New(spinner.CharSets[11], 100*time.Millisecond)
+	s.Suffix = "VM creation in progress"
+	s.Start()
 	logs, err := limaCmd.CombinedOutput()
 	if err != nil {
+		s.Stop()
 		iva.logger.Errorf("Finch virtual machine failed to start, debug logs: %s", logs)
 		return err
 	}
+	s.Stop()
```
otterley, line 108: You've got this twice - consider using a `defer s.Stop()` instead after `s.Start()`

hariohmprasath: Thanks, updated the code.

hariohmprasath:
> WDYT

Thanks @davidhsingyuchen, I have taken care of the comments.

**Regarding the progress:** I looked into this [code](https://github.com/runfinch/finch/blob/52ce4e77aeb679a380f7b580bb13ddd5c8a45b01/pkg/command/lima.go#L122) and seems like `lima` doesn't provide any async updates on the VM creation process. One option is to pipe the `stdout` from the creation progress to notify user that something is happening, but not sure whether thats the right thing to do. If you have any ideas let me know, happy to work on another PR. Thanks

cmd/finch/virtual_machine_init.go
davidhsingyuchen: Currently the logs are printed _before_ the spinner stops, while we may want them to be printed after instead.

```go
	if err != nil {
		iva.logger.Errorf("Finch virtual machine failed to start, debug logs: %s", logs)
		return err
	}
	iva.logger.Info("Finch virtual machine started successfully")
```

As a result, could you wrap the following statements in a func (`initVM`?) print the logs above after calling that func? Thanks!

```go
	instanceName := fmt.Sprintf("--name=%v", limaInstanceName)
	limaCmd := iva.creator.CreateWithoutStdio("start", instanceName, iva.baseYamlFilePath, "--tty=false")
	iva.logger.Info("Initializing and starting Finch virtual machine...")
	// Show spinner
	s := spinner.New(spinner.CharSets[11], 100*time.Millisecond)
	s.Suffix = "VM creation in progress"
	s.Start()
	defer s.Stop()
	logs, err := limaCmd.CombinedOutput()
```

davidhsingyuchen:
> Regarding the progress: I looked into this [code](https://github.com/runfinch/finch/blob/52ce4e77aeb679a380f7b580bb13ddd5c8a45b01/pkg/command/lima.go#L122) and seems like lima doesn't provide any async updates on the VM creation process. One option is to pipe the stdout from the creation progress to notify user that something is happening, but not sure whether thats the right thing to do. If you have any ideas let me know, happy to work on another PR. Thanks

@hariohmprasath Thanks for the investigation. I think the current form is good enough. We may not want to pipe the stdout from the creation progress as we cannot really control what's being printed out there, and it's sort of implementation details. /cc @ningziwen

h-rajagopal:
| Priority | Category    | Description                                                                                                                                | Suggestion                                                                                                                                                    | Code Reference                                                                                                                       |
|----------|-------------|--------------------------------------------------------------------------------------------------------------------------------------------|---------------------------------------------------------------------------------------------------------------------------------------------------------------|--------------------------------------------------------------------------------------------------------------------------------------|
| ⚠️       | Best Practice / Concurrency | The spinner is deferred to stop only when `run()` returns, so its goroutine may keep writing to the terminal while you collect and print `logs`/`err`, leading to interleaved output or race-like artifacts. | Stop the spinner immediately before handling or printing command output. For example, call `s.Stop()` right after `CombinedOutput()` returns and before logging results. | [virtual_machine_init.go#L97-L101](https://github.com/runfinch/finch/blob/main/cmd/finch/virtual_machine_init.go#L97-L101)         |
| 🟡       | Best Practice            | Indirect dependencies (e.g., `golang.org/x/term v0.4.0`) and transient module updates have been added to `go.mod` without ensuring they’re actually used. | Run `go mod tidy` to prune unused requirements and keep the module graph minimal, reducing build complexity and potential security surface.                     | [go.mod#L36](https://github.com/runfinch/finch/blob/main/go.mod#L36)                                                                  |
| 🟡       | Best Practice            | The new `go.sum` entries include `github.com/mattn/go-isatty v0.0.8/go.mod` without its corresponding module version line, suggesting stale or extraneous checksums. | Clean up `go.sum` (e.g., via `go mod tidy`) to remove orphaned entries and avoid checksum mismatches or confusion during CI builds.                            | [go.sum#L399-L403](https://github.com/runfinch/finch/blob/main/go.sum#L399-L403)                                                      |
| ℹ️       | Info                     | Introducing a spinner helps provide user feedback during potentially long VM creation, improving CLI UX.                                     | N/A                                                                                                                                                           | [virtual_machine_init.go#L97-L100](https://github.com/runfinch/finch/blob/main/cmd/finch/virtual_machine_init.go#L97-L100)         |

h-rajagopal:
| Priority | Category     | Description                                                                                                                                           | Suggestion                                                                                                                                                                                   | Code Reference                                                                                                                    |
|----------|--------------|-------------------------------------------------------------------------------------------------------------------------------------------------------|----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|-----------------------------------------------------------------------------------------------------------------------------------|
| 🔥       | Best Practice | Spinner is always started regardless of terminal capabilities, leading to stray control characters in non-TTY environments (e.g., CI logs).            | Guard spinner startup by detecting TTY (e.g., `golang.org/x/term.IsTerminal(os.Stdout.Fd())`) or use `spinner.NewWithWriter()` and disable it when not running interactively.               | [cmd/finch/virtual_machine_init.go#L98-L101](https://github.com/runfinch/finch/blob/main/cmd/finch/virtual_machine_init.go#L98-L101) |
| 🟡       | Code Quality  | Hard-coded `spinner.CharSets[11]` is a magic number—its purpose and visual style aren’t documented.                                                    | Define a named constant (e.g., `const vmSpinnerCharset = 11`) or add a comment explaining why charset 11 was chosen to improve clarity.                                                      | [cmd/finch/virtual_machine_init.go#L98](https://github.com/runfinch/finch/blob/main/cmd/finch/virtual_machine_init.go#L98)         |
| ℹ️       | Best Practice | Good use of `defer s.Stop()` ensures the spinner’s internal ticker goroutine is always stopped, avoiding resource leaks even on errors.               | —                                                                                                                                                                                            | [cmd/finch/virtual_machine_init.go#L100](https://github.com/runfinch/finch/blob/main/cmd/finch/virtual_machine_init.go#L100)       |

Shubhranshu153:
Closing this due to inactivity.

Issue #171, opened 2023-01-18
finch fails with "unrecognized system status"


otterley:
**Describe the bug**
finch can get into a state where, when running any command, it will fail and print

```
FATA[0000] unrecognized system status 
```

**Additional context**
Requested by engineers:

```
❯ LIMA_HOME=/Applications/Finch/lima/data /Applications/Finch/lima/bin/limactl ls
WARN[0000] instance "finch" has errors                   errors="[host agent is running but driver is not]"
NAME     STATUS    SSH                VMTYPE    ARCH       CPUS    MEMORY    DISK      DIR
finch    Broken    127.0.0.1:53899    qemu      aarch64    3       8GiB      100GiB    /Applications/Finch/lima/data/finch
```

sam-berning:
how did you install finch? via homebrew or the github release?

otterley:
Via the GitHub release (DMG file).

sam-berning:
Had you installed via homebrew previously by any chance?

otterley:
No, never.

sam-berning:
Ok, just wanted to rule out any chance that there were two running instances of Finch using the same disk. That's the only context where we've seen this bug before, but it seems like this might be something else.

In any case, https://github.com/runfinch/finch/pull/154 should help prevent this issue from happening. We'll have a release with that PR out soon.

sam-berning:
Just wanted to add a comment here on how to get back into a good state if you run into this issue. The first thing I would try is to manually delete the Lima instance directory, then re-init the finch VM.

```
LIMA_HOME=/Applications/Finch/lima/data/finch /Applications/Finch/lima/bin/limactl remove -f finch
finch vm init
```

If this doesn't fix the issue or if other issues arise with `finch images` or `finch ps`, it's possible that the persistent disk that stores containerd user data is corrupted. In this case, you should also delete the persistent disk. Note that this will cause any saved images or containers to be lost.

```
LIMA_HOME=/Applications/Finch/lima/data/finch /Applications/Finch/lima/bin/limactl remove -f finch
export FINCH_PERSISTENT_DISK=$(readlink /Applications/Finch/lima/data/_disks/finch/datadisk)
LIMA_HOME=/Applications/Finch/lima/data/finch /Applications/Finch/lima/bin/limactl disk delete finch
rm $FINCH_PERSISTENT_DISK
finch vm init
```

yamatatsu:
I get same error when I updated and restarted the Mac OS with finch vm running.

[...]

PR #178, opened 2023-01-20
feat: adds a --force flag to vm stop and remove

sam-berning:

Issue #, if available: https://github.com/runfinch/finch/issues/171

*Description of changes:*

Adds a force flag to `finch vm stop` and `finch vm remove`. This gives us a native way to recover from "unrecognized system status" as seen in https://github.com/runfinch/finch/issues/171.

*Testing done:*

Unit testing, E2E testing.

Force "unrecognized system status" by messing with the Lima instance
```
$ ./_output/bin/finch vm stop         
FATA[0000] unrecognized system status                   
$ ./_output/bin/finch vm stop --force
INFO[0000] Forcibly stopping Finch virtual machine...   
INFO[0001] Finch virtual machine stopped successfully   
$ ./_output/bin/finch vm start       
INFO[0000] Starting existing Finch virtual machine...   
INFO[0038] Finch virtual machine started successfully
```
```
$ ./_output/bin/finch vm stop         
FATA[0000] unrecognized system status                   
$ ./_output/bin/finch vm remove --force
INFO[0000] Forcibly removing Finch virtual machine...   
INFO[0000] Finch virtual machine removed successfully   
$ ./_output/bin/finch vm status        
Nonexistent
```

#### License Acceptance

### DIFF
--- a/cmd/finch/virtual_machine_remove.go
+++ b/cmd/finch/virtual_machine_remove.go
@@ -21,6 +21,8 @@ func newRemoveVMCommand(limaCmdCreator command.LimaCmdCreator, logger flog.Logge
 		RunE:  newRemoveVMAction(limaCmdCreator, logger).runAdapter,
 	}
 
+	removeVMCommand.Flags().BoolP("force", "f", false, "forcibly remove finch VM")
+
 	return removeVMCommand
 }
 
@@ -34,24 +36,24 @@ func newRemoveVMAction(creator command.LimaCmdCreator, logger flog.Logger) *remo
 }
 
 func (rva *removeVMAction) runAdapter(cmd *cobra.Command, args []string) error {
-	return rva.run()
+	force, err := cmd.Flags().GetBool("force")
+	if err != nil {
+		return err
+	}
+	return rva.run(force)
 }
 
-func (rva *removeVMAction) run() error {
+func (rva *removeVMAction) run(force bool) error {
+	if force {
+		return rva.removeVM(force)
+	}
+
 	err := rva.assertVMIsStopped(rva.creator, rva.logger)
 	if err != nil {
 		return err
 	}
 
-	limaCmd := rva.creator.CreateWithoutStdio("remove", limaInstanceName)
-	rva.logger.Info("Removing existing Finch virtual machine...")
-	logs, err := limaCmd.CombinedOutput()
-	if err != nil {
-		rva.logger.Errorf("Finch virtual machine failed to remove, debug logs: %s", logs)
-		return err
-	}
-	rva.logger.Info("Finch virtual machine removed successfully")
-	return nil
+	return rva.removeVM(false)
 }
 
 func (rva *removeVMAction) assertVMIsStopped(creator command.LimaCmdCreator, logger flog.Logger) error {
@@ -68,4 +70,27 @@ func (rva *removeVMAction) assertVMIsStopped(creator command.LimaCmdCreator, log
 	default:
 		return nil
 	}
+}
+
+func (rva *removeVMAction) removeVM(force bool) error {
+	limaCmd := rva.createVMRemoveCommand(force)
+	if force {
+		rva.logger.Info("Forcibly removing Finch virtual machine...")
+	} else {
+		rva.logger.Info("Removing existing Finch virtual machine...")
+	}
+	logs, err := limaCmd.CombinedOutput()
+	if err != nil {
+		rva.logger.Errorf("Finch virtual machine failed to remove, debug logs: %s", logs)
+		return err
+	}
+	rva.logger.Info("Finch virtual machine removed successfully")
+	return nil
+}
+
+func (rva *removeVMAction) createVMRemoveCommand(force bool) command.Command {
+	if force {
+		return rva.creator.CreateWithoutStdio("remove", "--force", limaInstanceName)
+	}
+	return rva.creator.CreateWithoutStdio("remove", limaInstanceName)
 }
--- a/cmd/finch/virtual_machine_remove_test.go
+++ b/cmd/finch/virtual_machine_remove_test.go
@@ -11,7 +11,6 @@ import (
 	"github.com/runfinch/finch/pkg/mocks"
 
 	"github.com/golang/mock/gomock"
-	"github.com/spf13/cobra"
 	"github.com/stretchr/testify/assert"
 )
 
@@ -28,14 +27,10 @@ func TestRemoveVMAction_runAdapter(t *testing.T) {
 	testCases := []struct {
 		name    string
 		mockSvc func(*mocks.Logger, *mocks.LimaCmdCreator, *gomock.Controller)
-		cmd     *cobra.Command
 		args    []string
 	}{
 		{
 			name: "should remove the instance",
-			cmd: &cobra.Command{
-				Use: "remove",
-			},
 			args: []string{},
 			mockSvc: func(logger *mocks.Logger, creator *mocks.LimaCmdCreator, ctrl *gomock.Controller) {
 				getVMStatusC := mocks.NewCommand(ctrl)
@@ -45,6 +40,18 @@ func TestRemoveVMAction_runAdapter(t *testing.T) {
 
 				command := mocks.NewCommand(ctrl)
 				creator.EXPECT().CreateWithoutStdio("remove", limaInstanceName).Return(command)
+				command.EXPECT().CombinedOutput()
+				logger.EXPECT().Info(gomock.Any()).AnyTimes()
+			},
+		},
+		{
+			name: "should forcibly remove the instance",
+			args: []string{
+				"--force",
+			},
+			mockSvc: func(logger *mocks.Logger, creator *mocks.LimaCmdCreator, ctrl *gomock.Controller) {
+				command := mocks.NewCommand(ctrl)
+				creator.EXPECT().CreateWithoutStdio("remove", "--force", limaInstanceName).Return(command)
 				command.EXPECT().CombinedOutput()
 				logger.EXPECT().Info(gomock.Any()).AnyTimes()
 			},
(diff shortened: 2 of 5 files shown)


e2e/vm/vm_test.go
```diff
@@ -27,7 +27,7 @@ func TestVM(t *testing.T) {
 		t.Fatal(err)
 	}
 
-	// The VM should be spined up in SynchronizedBeforeSuite of e2e/container.
+	// The VM should be spinned up in SynchronizedBeforeSuite of e2e/container.
```
estesp, line 30: minor nit: but if we're fixing some spelling, I think the tense is wrong: "The VM should be **spun** up.."

cmd/finch/virtual_machine_remove.go
```diff
@@ -34,24 +36,24 @@ func newRemoveVMAction(creator command.LimaCmdCreator, logger flog.Logger) *remo
...
+
+	err := rva.assertVMIsStopped(rva.creator, rva.logger)
 	if err != nil {
-		rva.logger.Errorf("Finch virtual machine failed to remove, debug logs: %s", logs)
 		return err
 	}
-	rva.logger.Info("Finch virtual machine removed successfully")
-	return nil
+
+	return rva.removeVM(force)
```
pendo324, line 56: Seeing this called with the force argument twice (even though the first time should handle the "forced" use case (since we just return on the first invocation) kinda makes me thing that we should just have two different functions, instead of overloading `rva.removeVM`. Maybe just `rva.removeVM` and `rva.forceRemoveVM`. Another option is to just call `rva.removeVM` with a hardcoded `false` argument the second time. Same feedback for the corresponding `vm stop` command. WDYT?

cmd/finch/virtual_machine_remove_test.go
```diff
@@ -49,19 +52,28 @@ func TestRemoveVMAction_runAdapter(t *testing.T) {
...
+			mockSvc: func(logger *mocks.Logger, creator *mocks.LimaCmdCreator, ctrl *gomock.Controller) {
+				command := mocks.NewCommand(ctrl)
+				creator.EXPECT().CreateWithoutStdio("remove", "--force", limaInstanceName).Return(command)
+				command.EXPECT().CombinedOutput()
+				logger.EXPECT().Info(gomock.Any()).AnyTimes()
+			},
+		},
 	}
 
+	//nolint:paralleltest // run in serial because tests with different args can interfere with each other
```
pendo324, line 70: Can you add more details into how tests with different arguments interfere with each other? Shouldn't that not be the case if we mock everything correctly?

e2e/vm/lifecycle_test.go
```diff
@@ -50,6 +57,12 @@ var testVMLifecycle = func(o *option.Option) {
 				command.New(o, virtualMachineRootCmd, "remove").WithTimeoutInSeconds(60).Run()
 				command.RunWithoutSuccessfulExit(o, "images")
 			})
+			ginkgo.It("should be able to force remove the virtual machine", func() {
```
pendo324, line 60: typo:
```suggestion
			ginkgo.It("should be able to force stop the virtual machine", func() {
```

cmd/finch/virtual_machine_remove_test.go
davidhsingyuchen, line 70: Could you elaborate more why they interfere with each other? There shouldn't be any state shared by unit tests.

cmd/finch/virtual_machine_remove_test.go
```diff
@@ -49,19 +52,28 @@ func TestRemoveVMAction_runAdapter(t *testing.T) {
...
+			name: "should forcibly remove the instance",
+			cmd:  removeCmd,
+			args: []string{
+				"--force",
+			},
+			mockSvc: func(logger *mocks.Logger, creator *mocks.LimaCmdCreator, ctrl *gomock.Controller) {
+				command := mocks.NewCommand(ctrl)
+				creator.EXPECT().CreateWithoutStdio("remove", "--force", limaInstanceName).Return(command)
+				command.EXPECT().CombinedOutput()
+				logger.EXPECT().Info(gomock.Any()).AnyTimes()
```
davidhsingyuchen, line 65: I understand that it was already `Info(gomock.Any()).AnyTimes()` before this PR, but maybe it's a good time for us to assert the exact statements as some changes are made to the logging statements in this PR anyway.

1. We can have more confidence.
2. We can have consistent logs assertions (i.e., most other test cases are doing exact matches if possible).

cmd/finch/virtual_machine_remove_test.go
```diff
@@ -49,19 +52,28 @@ func TestRemoveVMAction_runAdapter(t *testing.T) {
...
 		t.Run(tc.name, func(t *testing.T) {
-			t.Parallel()
-
-			ctrl := gomock.NewController(t)
-			logger := mocks.NewLogger(ctrl)
-			lcc := mocks.NewLimaCmdCreator(ctrl)
-
 			tc.mockSvc(logger, lcc, ctrl)
-			assert.NoError(t, newRemoveVMAction(lcc, logger).runAdapter(tc.cmd, tc.args))
+			tc.cmd.SetArgs(tc.args)
```
davidhsingyuchen, line 75: Maybe we can

1. Change `cmd *cobra.Command` to `cmd func() *cobra.Command` in the test case definition, and in that func, we can call `cmd.ParseFlags(...)` to set it up if needed.
2. Change what's inside `t.Run` back to the original flow.

Also, probably that's why the test cases cannot be run in parallel now?

cmd/finch/virtual_machine_remove.go
sam-berning: I was debating on this one when I wrote this. the reason I ended up going with this approach is that it's easier to keep feature parity between the two versions if we're not duplicating the code. (e.g. I plan to follow up with an option on `vm remove` to delete the finch persistent disk, which should work for both `--force` and regular)

I could go either way on this bc I think two separate functions is more readable overall. I think if we want to keep it this way, hardcoded `false` would be a good idea.

e2e/vm/lifecycle_test.go
```diff
@@ -50,6 +57,12 @@ var testVMLifecycle = func(o *option.Option) {
...
+				command.New(o, virtualMachineRootCmd, "init").WithTimeoutInSeconds(600).Run()
+				command.New(o, virtualMachineRootCmd, "stop").WithTimeoutInSeconds(60).Run()
+				command.New(o, virtualMachineRootCmd, "remove", "--force").WithTimeoutInSeconds(60).Run()
+				command.RunWithoutSuccessfulExit(o, "images")
+			})
```
davidhsingyuchen, line 65: Why do we need two `should be able to force remove the virtual machine`?

e2e/vm/lifecycle_test.go
```diff
@@ -23,7 +23,14 @@ var testVMLifecycle = func(o *option.Option) {
...
 			})
 
+			ginkgo.It("should be able to force remove the virtual machine", func() {
+				command.Run(o, "images")
+				command.New(o, virtualMachineRootCmd, "remove", "--force").WithTimeoutInSeconds(60).Run()
+				command.RunWithoutSuccessfulExit(o, "images")
+			})
+
 			ginkgo.It("should be able to stop the virtual machine", func() {
+				command.New(o, virtualMachineRootCmd, "init").WithTimeoutInSeconds(600).Run()
```
davidhsingyuchen, line 33: nit: maybe we can move this line to the last `It` so that for every `It` inside this `When` block (i.e., `"the virtual machine is in running status"`), the VM is running.

cmd/finch/virtual_machine_remove_test.go
sam-berning: I just realized that this is just because I use the same instance of the cobra command. will fix in next revision

davidhsingyuchen: Probably related: https://github.com/runfinch/finch/pull/178#discussion_r1083133815

davidhsingyuchen: race condition of https://github.com/runfinch/finch/pull/178#discussion_r1083129790 😓

cmd/finch/virtual_machine_remove.go
pendo324: I think I'm good with just hardcoding `false` on the second invocation

cmd/finch/virtual_machine_remove_test.go
sam-berning: the reason I chose to leave it like this is that these tests are meant to cover the `runAdapter` code, and so it's mainly testing if the `--force` argument is parsed correctly. I don't think changing the logging should impact these tests, as we already cover that in the `run` tests below.

cmd/finch/virtual_machine_remove_test.go
sam-berning: yeah, this should be addressed by https://github.com/runfinch/finch/pull/178#discussion_r1083145163

e2e/vm/lifecycle_test.go
sam-berning: force remove should work in any state (apart from nonexistent) so I wanted to cover both "Running" and "Stopped". perhaps it makes more sense for this test to just be force stop, though

davidhsingyuchen: Yeah, I feel we can just make it `stop --force` instead and move it to the "the virtual machine is in running status" `When` block.

> force remove should work in any state (apart from nonexistent) so I wanted to cover both "Running" and "Stopped".

I feel testing `remove --force` for `Stopped` may not be necessary as a stopped VM should be able to be removed with `--force` in the first place, but I'm also fine with having it.

pendo324 (approved):

davidhsingyuchen (dismissed):

pendo324 (approved):

sam-berning:
Rebased with the increased `finch vm stop timeout` from https://github.com/runfinch/finch/pull/189

pendo324 (approved):

davidhsingyuchen (approved):

PR #179, opened 2023-01-21
chore: allow e2e suites to run separately

sam-berning:

Issue #, if available:

*Description of changes:*

Had a discussion with @davidhsingyuchen about how to run tests in only one of the e2e test suites, and this is the approach we landed on.

- This will allow users to more easily test only the e2e test scenarios they are working on
- We can potentially run the two suites in parallel in CI on different runners

One drawback of this approach is that each run of the full e2e tests will take 3-5 minutes longer, due to running the `SynchronizedBeforeSuite` an additional time.

*Testing done:*

```
make test-e2e
make test-e2e-container
make test-e2e-vm
```

#### License Acceptance


Makefile
```diff
@@ -247,8 +247,18 @@ test-unit:
 # Container tests have to be run before VM tests according to the current code setup.
 # For more details, see the package-level comment of the e2e package.
```
davidhsingyuchen, line 248: Could you also update this? The order doesn't matter now, and they just need to be run sequentially.

e2e/vm/vm_test.go
```diff
@@ -27,7 +30,12 @@ func TestVM(t *testing.T) {
 		t.Fatal(err)
 	}
 
-	// The VM should be spined up in SynchronizedBeforeSuite of e2e/container.
+	ginkgo.SynchronizedBeforeSuite(func() []byte {
+		command.New(o, "vm", "init").WithTimeoutInSeconds(600).Run()
+		tests.SetupLocalRegistry(o)
```
davidhsingyuchen, line 35: We may not need this for VM tests.

```suggestion
```

davidhsingyuchen (approved):

PR #189, opened 2023-01-26
test: increase vm stop timeout to 90s

sam-berning:

Issue #, if available: https://github.com/runfinch/finch/issues/140

*Description of changes:*

Increases the `finch vm stop` timeout in e2e tests to 90 seconds.

In our CI/CD we've been seeing timeouts on most runs on ARM 11.7 instances, and occasionally on ARM 12.6 instances. Out of the 19 times that the `finch vm stop` command runs per run of the e2e tests, it tends to timeout only once or twice at most.

I've observed a couple of times on my local machine that `finch vm stop` can take 40-50 seconds, so it's reasonable to believe that increasing the timeout by another 30 seconds would eliminate most of these timeouts.

*Testing done:*

```
make test-e2e
```

#### License Acceptance

### DIFF
--- a/e2e/container/container_test.go
+++ b/e2e/container/container_test.go
@@ -31,7 +31,7 @@ func TestContainer(t *testing.T) {
 	}, func(bytes []byte) {})
 
 	ginkgo.SynchronizedAfterSuite(func() {
-		command.New(o, "vm", "stop").WithTimeoutInSeconds(60).Run()
+		command.New(o, "vm", "stop").WithTimeoutInSeconds(90).Run()
 		command.New(o, "vm", "remove").WithTimeoutInSeconds(60).Run()
 	}, func() {})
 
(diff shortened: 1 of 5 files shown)


pendo324 (approved):
Would be good to get to the root cause of why `vm stop` is taking so long to begin with (this never happens on my machine), but this is a good stop-gap. Thanks 👍

sam-berning:
> Would be good to get to the root cause of why vm stop is taking so long to begin with

yeah, I didn't end up root causing it when I saw it earlier. I've been trying to reproduce it on my machine today to take a look at the Lima hostagent logs, but no luck so far.

sam-berning:
Just an interesting note, we saw one instance in the e2e tests on this PR where the `finch vm stop` command took exactly 60s. From the ARM 11.7 logs:

```
time="2023-01-26T23:37:58Z" level=info msg="Stopping existing Finch virtual machine..."
time="2023-01-26T23:38:58Z" level=info msg="Finch virtual machine stopped successfully"
```

PR #202, opened 2023-01-31
feat: add release-homebrew workflow

AnqiPang:
Signed-off-by: Anqi Pang <anqipans@amazon.com>

Issue #, if available:

*Description of changes:*

*Testing done:*
Tested on self-hosted runners except PR to homebrew.

#### License Acceptance

### DIFF
--- a/.github/workflows/release-homebrew.yaml
+++ b/.github/workflows/release-homebrew.yaml
@@ -7,9 +7,6 @@ name: Release Homebrew
 on:
   workflow_dispatch:
 
-permissions:
-  contents: read # This is required for actions/checkout
-
 jobs:
   get-latest-tag:
     name: Get the latest release tag
@@ -93,6 +90,10 @@ jobs:
         run: |
           cd $(brew --repo homebrew/cask)
           brew uninstall --cask ./Casks/finch.rb
+          if [ -d "/Applications/Finch" ]; then
+            echo ERROR: Finch is not uninstalled
+            exit 1
+          fi
 
   macos-amd64-test-installer:
     needs: get-latest-tag
@@ -147,6 +148,10 @@ jobs:
         run: |
           cd $(brew --repo homebrew/cask)
           brew uninstall --cask ./Casks/finch.rb
+          if [ -d "/Applications/Finch" ]; then
+            echo ERROR: Finch is not uninstalled
+            exit 1
+          fi
 
   pr-to-homebrew:
     needs: [macos-arm64-test-installer, macos-amd64-test-installer]
--- a/.github/workflows/release-homebrew.yaml
+++ b/.github/workflows/release-homebrew.yaml
@@ -25,12 +25,14 @@ jobs:
       - name: 'Get the latest tag'
         id: latesttag
         uses: "WyriHaximus/github-action-get-previous-tag@v1"
-        with:
-          fallback: v0.1.0
       - name: 'Convert tag to version'
         id: latestversion
         run: |
           version=${{ steps.latesttag.outputs.tag }}
+          if [ -z $version ]; then
+            echo ERROR: Failed to get the latest tag
+            exit 1
+          fi
           if [[  $version == v* ]]; then
             version="${version:1}"
           fi
--- a/.github/workflows/release-homebrew.yaml
+++ b/.github/workflows/release-homebrew.yaml
@@ -1,0 +1,164 @@
+# When a third-party action is added (i.e., `uses`), please also add it to `download-licenses` in Makefile.
+# When the installers are ready in installer private bucket, run this workflow to test them.
+# TODO: Add job of uploading installers to Github release when installer tests are stable.
+# "release" runner tag is to make target hosts deterministic for easy clean up.
+# TODO: Remove the "release" runner tag when installer tests are stable.
+name: Release Homebrew
+on:
+  workflow_dispatch:
+
+permissions:
+  id-token: write # This is required for requesting the JWT
+  contents: read # This is required for actions/checkout
+
+jobs:
+  get-latest-tag:
+    name: Get the latest release tag
+    runs-on: ubuntu-latest
+    outputs:
+      tag: ${{ steps.latesttag.outputs.tag }}
+      version: ${{ steps.latestversion.outputs.version }}
+    steps:
+      - uses: actions/checkout@v3
+        with:
+            fetch-depth: 0
+      - name: 'Get the latest tag'
+        id: latesttag
+        uses: "WyriHaximus/github-action-get-previous-tag@v1"
+        with:
+          fallback: v0.1.0
+      - name: 'Convert tag to version'
+        id: latestversion
+        run: |
+          version=${{ steps.latesttag.outputs.tag }}
+          if [[  $version == v* ]]; then
+            version="${version:1}"
+          fi
+          echo "version=$version" >> ${GITHUB_OUTPUT}
+
+  # TODO: fix the arm64 test installer.
+  # Currently the first time of calling any Finch command in arm64 hosts triggered by Github action will fail by
+  # the error "Error: Process completed with exit code 137." So the arm64 job will fail.
+  # We temporarily use follow-up manual steps to complete and clean up it.
+  macos-arm64-test-installer:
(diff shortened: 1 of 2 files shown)


.github/workflows/release-homebrew.yaml
```diff
@@ -0,0 +1,164 @@
+# When a third-party action is added (i.e., `uses`), please also add it to `download-licenses` in Makefile.
+# When the installers are ready in installer private bucket, run this workflow to test them.
+# TODO: Add job of uploading installers to Github release when installer tests are stable.
```
KevinLiAWS, line 3: Seems homebrew only need to make a pr instead of uploading the pkg?

.github/workflows/release-homebrew.yaml
```diff
@@ -0,0 +1,164 @@
...
+      ACCESS_TOKEN: ${{ secrets.FINCH_BOT_TOKEN }}
+      FINCH_TAG: ${{ needs.get-latest-tag.outputs.tag }}
+      FINCH_VESRION: ${{ needs.get-latest-tag.outputs.version }}
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          ref: ${{ env.FINCH_TAG }}
+      - name: Clean up previous files
+        run: |
+          sudo rm -rf /opt/finch
```
KevinLiAWS, line 121: I think postinstall script is doing the clean up, better skip the clean up explicitly which will be more similar to actual use case?

.github/workflows/release-homebrew.yaml
```diff
@@ -0,0 +1,164 @@
...
+          cd $(brew --repo homebrew/cask)
+          echo "Update the local Finch cask with version ${FINCH_VERSION}"
+          brew bump-cask-pr --write-only -f --version=${FINCH_VERSION} finch
+      - name: Silently install
+        run: |
+          cd $(brew --repo homebrew/cask)
+          brew install --cask ./Casks/finch.rb
+      - name: Run e2e tests
+        run: INSTALLED=true make test-e2e
+      - name: Silently uninstall
```
KevinLiAWS, line 146: Not related with PR, but seems we didn't have installer test such as after uninstall, e.g., need to assert directory is removed after uninstall

.github/workflows/release-homebrew.yaml
```diff
@@ -0,0 +1,164 @@
...
+        os:
+          [
+            [self-hosted, macos, arm64, 11.7, release],
+            [self-hosted, macos, arm64, 12.6, release],
+            [self-hosted, macos, arm64, 13.0, release],
+          ]
+    runs-on: ${{ matrix.os }}
+    timeout-minutes: 60
+    env:
+      ACCESS_TOKEN: ${{ secrets.FINCH_BOT_TOKEN }}
```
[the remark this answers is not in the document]
AnqiPang: @ningziwen Could you confirm what the env var ACCESS_TOKEN is used for? The release-installer.yaml also contains this. https://github.com/runfinch/finch/blob/main/.github/workflows/release-installer.yaml#L34

.github/workflows/release-homebrew.yaml
AnqiPang: I can investigate whether we can add it to the workflow.

.github/workflows/release-homebrew.yaml
AnqiPang: Could you give more context about the postinstall script? What is it and how is it triggered?

.github/workflows/release-homebrew.yaml
```diff
@@ -0,0 +1,164 @@
...
+      version: ${{ steps.latestversion.outputs.version }}
+    steps:
+      - uses: actions/checkout@v3
+        with:
+            fetch-depth: 0
+      - name: 'Get the latest tag'
+        id: latesttag
+        uses: "WyriHaximus/github-action-get-previous-tag@v1"
+        with:
+          fallback: v0.1.0
```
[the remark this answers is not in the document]
AnqiPang: Good point, will update it. Do you have any concern for using the third-party github action `WyriHaximus/github-action-get-previous-tag@v1`? I'm not clear about the policy for that.

.github/workflows/release-homebrew.yaml
AnqiPang: Will do.

.github/workflows/release-homebrew.yaml
```diff
@@ -0,0 +1,164 @@
...
+# When the installers are ready in installer private bucket, run this workflow to test them.
+# TODO: Add job of uploading installers to Github release when installer tests are stable.
+# "release" runner tag is to make target hosts deterministic for easy clean up.
+# TODO: Remove the "release" runner tag when installer tests are stable.
+name: Release Homebrew
+on:
+  workflow_dispatch:
+
+permissions:
+  id-token: write # This is required for requesting the JWT
```
vsiravar, line 11: Do we need this token? Unsure if we are accessing any resource that requires an ODIC JWT token.

AnqiPang: I don't think so. Will remove it.

.github/workflows/release-homebrew.yaml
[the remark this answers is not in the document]
vsiravar: We use third-party actions in other workflows. Eg. `peter-evans/create-pull-request@v4` , so I think we should be good here. @ningziwen WDYT?

.github/workflows/release-homebrew.yaml
AnqiPang: > Seems homebrew only need to make a pr instead of uploading the pkg?
I think the TODO item means we need a workflow to upload the pkg installer from s3 bucket to the release assets after the e2e tests passes. It's not for homebrew release. Does it make sense to you?

.github/workflows/release-homebrew.yaml
vsiravar: Even for `contents` scope, the [default](https://docs.github.com/en/actions/security-guides/automatic-token-authentication#permissions-for-the-github_token) access is read for the GITHUB_TOKEN. Maybe we can just remove the `permissions` block?

.github/workflows/release-homebrew.yaml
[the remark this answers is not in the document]
ningziwen: This is needed for in your last step to raise the Homebrew PR but looks like you directly use it in line. So env is not needed. For installer test, it will be used to cut Github release. https://github.com/runfinch/finch/blob/main/.github/workflows/release-installer.yaml#L3

.github/workflows/release-homebrew.yaml
AnqiPang: Added check for uninstallation.

.github/workflows/release-homebrew.yaml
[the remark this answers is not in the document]
AnqiPang: Does the previous step refers to the release-installer workflow? We can trigger another workflow from workflow with personal access token of required permission or create a Github App. I think the former takes much less effort for implementation as long as we give right permission to the bot.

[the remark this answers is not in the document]
vsiravar: If we want to avoid using `WyriHaximus/github-action-get-previous-tag@v1`, we could get the latest tag by checking out `finch` and running tag=\`git describe --tags $(git rev-list --tags --max-count=1)`.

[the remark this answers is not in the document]
AnqiPang: I think it's okay to use the action for now. What do you think?

.github/workflows/release-homebrew.yaml
```diff
@@ -0,0 +1,163 @@
+# When a third-party action is added (i.e., `uses`), please also add it to `download-licenses` in Makefile.
+# When the installers are ready in installer private bucket, run this workflow to test them.
```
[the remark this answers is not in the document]
AnqiPang: Removed

KevinLiAWS (approved):

vsiravar (approved):

ningziwen:
nit: should this be "ci" instead of "feat?

PR #210, opened 2023-02-08
feat: adds support bundles

sam-berning:

Issue #, if available:

*Description of changes:*

Allow users to generate a support bundle, a collection of Finch-related logs and configs that will help debug issues.

*Testing done:*

Unit & manual testing

#### License Acceptance


e2e/vm/support_bundle_test.go
```diff
@@ -0,0 +1,37 @@
...
+	"os"
+	"strings"
+
+	"github.com/onsi/ginkgo/v2"
+	"github.com/onsi/gomega"
+	"github.com/runfinch/common-tests/command"
+	"github.com/runfinch/common-tests/option"
+)
+
+var testSupportBundle = func(o *option.Option) {
```
ningziwen, line 16: Any other scenarios?

cmd/finch/support_bundle.go
```diff
@@ -0,0 +1,99 @@
...
+		return err
+	}
+	a.logger.Info("Generating support bundle...")
+	bundleFile, err := a.builder.GenerateSupportBundle(additionalFiles, excludeFiles)
+	if err != nil {
+		return err
+	}
+	a.logger.Infof("Bundle created: %s", bundleFile)
+	a.logger.Info("Files posted on a Github issue can be read by anyone.")
+	a.logger.Info("Please ensure there is no sensitive information in the bundle before uploading.")
```
weikequ, line 83: Should we list all the files (or the types of files) that will be included to be transparent?

cmd/finch/support_bundle.go
```diff
@@ -0,0 +1,99 @@
...
+	lcc command.LimaCmdCreator,
+) *generateSupportBundleAction {
+	return &generateSupportBundleAction{
+		logger:  logger,
+		builder: builder,
+		lcc:     lcc,
+	}
+}
+
+func (a *generateSupportBundleAction) runAdapter(cmd *cobra.Command, args []string) error {
```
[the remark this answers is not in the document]
sam-berning: in the repo it's convention for these action variable names to be short abbreviations but not quite this short (e.g. `sva` for `startVMAction`). I'll change this one to `gsa`

cmd/finch/support_bundle.go
sam-berning: that's a good idea. I'll add a line for that

e2e/vm/support_bundle_test.go
ningziwen: I guess e2e tests are just the guard to prevent unexpected regression instead of something we believe it will never change?

cmd/finch/support_bundle.go
```diff
@@ -0,0 +1,99 @@
...
+func newSupportBundleGenerateCommand(logger flog.Logger, builder support.BundleBuilder, lcc command.LimaCmdCreator) *cobra.Command {
+	supportBundleGenerateCommand := &cobra.Command{
+		Use:   "generate",
+		Args:  cobra.NoArgs,
+		Short: "Generate support bundle",
+		RunE:  newGenerateSupportBundleAction(logger, builder, lcc).runAdapter,
+	}
+
+	supportBundleGenerateCommand.Flags().StringArray("include", []string{}, "additional files to include in the support bundle")
+	supportBundleGenerateCommand.Flags().StringArray("exclude", []string{}, "files to exclude from the support bundle")
```
ningziwen, line 37: I guess the files to be excluded are deterministric? Should we list all the possible strings or give some examples here? I don't know what format of string to put here only from this description.

cmd/finch/support_bundle.go
```diff
@@ -0,0 +1,99 @@
...
+
+func newSupportBundleGenerateCommand(logger flog.Logger, builder support.BundleBuilder, lcc command.LimaCmdCreator) *cobra.Command {
+	supportBundleGenerateCommand := &cobra.Command{
+		Use:   "generate",
+		Args:  cobra.NoArgs,
+		Short: "Generate support bundle",
+		RunE:  newGenerateSupportBundleAction(logger, builder, lcc).runAdapter,
+	}
+
+	supportBundleGenerateCommand.Flags().StringArray("include", []string{}, "additional files to include in the support bundle")
```
ningziwen, line 36: It is not clear what to put in --include and --exclude. Should it be absolute path or relative path?

e2e/vm/support_bundle_test.go
sam-berning: Yeah that's fair. I'll add the new scenarios.

sam-berning: > And also what if one file is in both --include and --exclude?

We currently favor `--exclude` in this scenario, i.e. the file will not be included, as we want to ensure customers can exclude a file if they want to.

cmd/finch/support_bundle.go
sam-berning: include can be either a relative or absolute path, exclude can be an absolute or relative path or the basename of a file. I'll add some more words here

e2e/vm/support_bundle_test.go
ningziwen: Yeah I think it is good to cover it in documentation/CLI description and e2e tests.

cmd/finch/support_bundle.go
ningziwen: Will it cause lots of complicated scenarios if we support both "basename" or "relative path"? What if there is a "finch.yaml" in the current relative path? Will the command exclude the one in current relative path or ~/.finch/finch.yaml in that case?

I think we need to either try to remove the uncertain branches to simply the scenarios or cover them in e2e tests and docuemntation/CLI description.

sam-berning: > Will the command exclude the one in current relative path or ~/.finch/finch.yaml in that case?

In that case if it's specified by basename, all files named `finch.yaml` will be excluded. If a user wants to exclude just one of the specific files, they can specify the full absolute or relative path to the file. I'll add some documentation on that to the CLI option.

e2e/vm/support_bundle_test.go
```diff
@@ -0,0 +1,159 @@
...
+					_, err = reader.Open(path.Join(zipPrefix, "logs", "serial.log"))
+					gomega.Expect(err).ShouldNot(gomega.BeNil())
+
+					err = os.Remove(dirEntry.Name())
+					gomega.Expect(err).Should(gomega.BeNil())
+				}
+			}
+			gomega.Expect(bundleExists).Should(gomega.BeTrue())
+		})
+		ginkgo.It("Should generate a support bundle with the same file included and excluded", func() {
```
ningziwen, line 100: Something like "Should generate a support bundle with the file excluded when it is both in --included and --exclude". Including both of the result and condition in the description can make the test log more readable.

e2e/vm/support_bundle_test.go
```diff
@@ -0,0 +1,159 @@
...
+						bundleExists = true
+					}
+
+					err = os.Remove(dirEntry.Name())
+					gomega.Expect(err).Should(gomega.BeNil())
+				}
+			}
+			gomega.Expect(bundleExists).Should(gomega.BeFalse())
+
+			command.New(o, "vm", "init").WithTimeoutInSeconds(600).Run()
```
ningziwen, line 156: Should we move to line 137 and add defer, to succeed other tests even if this fails?

e2e/vm/support_bundle_test.go
ningziwen: Same for other descriptions.

e2e/vm/support_bundle_test.go
```diff
@@ -0,0 +1,159 @@
...
+					gomega.Expect(err).Should(gomega.BeNil())
+
+					err = os.Remove(dirEntry.Name())
+					gomega.Expect(err).Should(gomega.BeNil())
+				}
+			}
+			gomega.Expect(bundleExists).Should(gomega.BeTrue())
+		})
+		ginkgo.It("Should generate a support bundle with a default file excluded", func() {
+			command.Run(o, "support-bundle", "generate", "--exclude", "serial.log")
```
ningziwen, line 75: Also add abstract and relative path for both include and exclude? It should be a quick change that only needs copy-paste

cmd/finch/support_bundle.go
```diff
@@ -0,0 +1,103 @@
...
+		Use:   "generate",
+		Args:  cobra.NoArgs,
+		Short: "Generate support bundle",
+		Long:  "Generates a collection of logs and configs that can be uploaded to a Github issue to help debug issues.",
+		RunE:  newGenerateSupportBundleAction(logger, builder, lcc).runAdapter,
+	}
+
+	supportBundleGenerateCommand.Flags().StringArray("include", []string{},
+		"additional files to include in the support bundle, specified by absolute or relative path")
+	supportBundleGenerateCommand.Flags().StringArray("exclude", []string{},
```
ningziwen, line 39: Where is the best place to state "exclude" has higher privilege than "include"? Maybe specifiy in both --exclude and --include?

cmd/finch/support_bundle.go
ningziwen: Will you add examples?

sam-berning: On second thought, I don't think we should. I think it may clutter up the description too much, and I think it's better to encourage trial and error.

e2e/vm/support_bundle_test.go
ningziwen: @sam-berning How about exclude by relative path?

ningziwen:
@sam-berning Need to resolve the conflict files.

e2e/vm/support_bundle_test.go
ningziwen: Also how about invalid path for include and exclude?

e2e/vm/support_bundle_test.go
```diff
@@ -0,0 +1,228 @@
...
+	"time"
+
+	"github.com/onsi/ginkgo/v2"
+	"github.com/onsi/gomega"
+	"github.com/runfinch/common-tests/command"
+	"github.com/runfinch/common-tests/option"
+)
+
+var testSupportBundle = func(o *option.Option) {
+	ginkgo.Describe("Support bundles", func() {
```
ningziwen, line 22: Is it possible to add couple of lines to validate the zip name and the unzipped folder name here?

ningziwen (approved):
.  o ..                  
     o . o o.o                
          ...oo               
            __[]__            
         __|_o_o_o\__         
         \""""""""""/         
          \. ..  . /          
     ^^^^^^^^^^^^^^^^^^^^

weikequ (approved):
Looks good!

Issue #180, opened 2023-01-23
Container and volumes data does not persist between different VM instances


mharwani:
**Describe the bug**
Cannot start a created container after a new installation of Finch, or after VM re-initialization.

**Steps to reproduce**
1. `finch run --name test-container -v ~/workplace:/workplace alpine ls /workplace`
2. `finch vm stop && finch vm remove && finch vm init`
```
INFO[0000] Stopping existing Finch virtual machine...   
INFO[0003] Finch virtual machine stopped successfully   
INFO[0000] Removing existing Finch virtual machine...   
INFO[0000] Finch virtual machine removed successfully   
INFO[0000] Initializing and starting Finch virtual machine... 
INFO[0103] Finch virtual machine started successfully
```
3. `finch ps -a`
```
CONTAINER ID    IMAGE                              COMMAND            CREATED          STATUS     PORTS    NAMES
6afe448dcc57    docker.io/library/alpine:latest    "ls /workplace"    4 minutes ago    Created             test-container
```
4. `finch start --attach test-container`
```
FATA[0000] failed to create shim task: OCI runtime create failed: runc create failed: unable to start container process: error during container init: error mounting "/home/mharwani.linux/.local/share/nerdctl/1935db59/containers/finch/6afe448dcc572d9dceec82d1e5d1c42ba70d46106de062eea277aa696fe5db43/hostname" to rootfs at "/etc/hostname": stat /home/mharwani.linux/.local/share/nerdctl/1935db59/containers/finch/6afe448dcc572d9dceec82d1e5d1c42ba70d46106de062eea277aa696fe5db43/hostname: no such file or directory: unknown 
FATA[0000] exit status 1
```

**Expected behavior**
`test-container` should start and display contents of `~/workplace` without error.

**Additional context**
`nerdctl` user data (located in `~/.local/share/nerdctl`) is not stored in the persistent disk. This directory stores important information such as volumes, container names, hostnames, etc. May want to add `~/.local/share/nerdctl` to [additionalDisks](https://github.com/runfinch/finch/blob/v0.3.0/finch.yaml#L152)

weikequ:
Can reproduce on my own machine running finch v0.3.0 running macOS 12.6.2 on arm64.

```bash
❯ finch start --attach test-container
[...]

PR #211, opened 2023-02-08
fix: persists network user data

sam-berning:

Issue #, if available: https://github.com/runfinch/finch/issues/180

*Description of changes:*

Persists the `cni` user data stored in `~/.local/share/cni` and in `~/.config/cni`

*Testing done:*

Manual testing

```
$ finch network create samtest --subnet 10.8.0.0/24
$ finch run --network=samtest --name hello public.ecr.aws/amazonlinux/amazonlinux:2 sleep 1
$ finch vm stop
$ finch vm remove
$ finch vm init
$ finch start hello
```

#### License Acceptance

### DIFF
--- a/finch.yaml
+++ b/finch.yaml
@@ -149,18 +149,25 @@ provision:
   script: |
     #!/bin/bash
     sudo chown $USER /mnt/lima-finch
-    # https://github.com/containerd/nerdctl/blob/main/extras/rootless/containerd-rootless.sh#L146
+
+    # https://github.com/containerd/nerdctl/blob/cffdf87ff4d648a5344eea1406bb95ca3ad7eaa4/extras/rootless/containerd-rootless.sh#L144-L146
+    # XDG_DATA_HOME & ~/.local/share: https://github.com/containerd/nerdctl/blob/cffdf87ff4d648a5344eea1406bb95ca3ad7eaa4/extras/rootless/containerd-rootless.sh#L51
     mkdir -p /mnt/lima-finch/containerd ~/.local/share/containerd
     sudo mount --bind /mnt/lima-finch/containerd ~/.local/share/containerd
+
     # https://github.com/containerd/nerdctl/blob/main/docs/dir.md#dataroot
     mkdir -p /mnt/lima-finch/nerdctl ~/.local/share/nerdctl
     sudo mount --bind /mnt/lima-finch/nerdctl ~/.local/share/nerdctl
+
     # https://github.com/containerd/nerdctl/blob/main/docs/dir.md#netconfpath
     mkdir -p /mnt/lima-finch/cni-config ~/.config/cni
     sudo mount --bind /mnt/lima-finch/cni-config ~/.config/cni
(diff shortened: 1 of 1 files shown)


finch.yaml
```diff
@@ -151,9 +151,14 @@ provision:
     sudo chown $USER /mnt/lima-finch
     mkdir -p /mnt/lima-finch/containerd
     mkdir -p /mnt/lima-finch/nerdctl
+    mkdir -p /mnt/lima-finch/cni-config
+    mkdir -p /mnt/lima-finch/cni-local
     mkdir -p ~/.local/share/nerdctl
+    mkdir -p ~/.config/cni
```
davidhsingyuchen, line 157: Why don't we need to `mkdir -p` for `~/.local/share/containerd` and `~/.local/share/cni`?

finch.yaml
```diff
@@ -151,9 +151,14 @@ provision:
...
     sudo mount --bind /mnt/lima-finch/containerd ~/.local/share/containerd
     sudo mount --bind /mnt/lima-finch/nerdctl ~/.local/share/nerdctl
+    sudo mount --bind /mnt/lima-finch/cni-config ~/.config/cni
+    sudo mount --bind /mnt/lima-finch/cni-local ~/.local/share/cni
```
davidhsingyuchen, line 161: Maybe we can add some reference links for each line. Take `~/.local/share/nerdctl` for example, the reference link would be https://github.com/containerd/nerdctl/blob/main/docs/dir.md#data.

finch.yaml
sam-berning: these both exist by default by the time this script is run, whereas the `cni` and `nerdctl` configs are only created when they're used. I suppose there's no harm in adding them to make sure they are created in case this changes later.

davidhsingyuchen: > these both exist by default by the time this script is run, whereas the cni and nerdctl configs are only created when they're used.

I see.

> I suppose there's no harm in adding them to make sure they are created in case this changes later.

Agree. Maybe we can add them then. It can also lower cognitive overhead (i.e., we won't need to know which ones are only created when they're used; instead, we just make sure everything is there).

finch.yaml
sam-berning: I like this in theory, but I'm not too sure there is a good reference for the others. "~/.local/share/cni" turns up no results on Google, for example. I can add them on a best-effort basis.

davidhsingyuchen: That's true. If we don't add `~/.local/share/cni`, what errors will be there? Maybe we start from there to see which part of code is using it. The link doesn't have to be a user-facing doc; it could be a permalink to the relevant code if that's what we can get.

sam-berning: I was able to find [this snippet of code in `containerd-rootless.sh`](https://github.com/containerd/nerdctl/blob/main/extras/rootless/containerd-rootless.sh#L145-L150) that references the cni and containerd user data directories, so that should work.

pendo324 (approved):
Thanks for adding the doc links

PR #213, opened 2023-02-09
refactor: replace os calls with afero.Fs ones in nerdctlCommand

davidhsingyuchen:
## Summary

PR fixes https://github.com/runfinch/finch/pull/158#discussion_r1101922922. Besides that, it also adds a new test case, namely `"with --env-file flag, but the specified file does not exist"`, which becomes easier to add after the refactoring.

## Notes

`gosec` is removed from `//nolint:errcheck,gosec` because `gosec` checks things like `os.Open` but not `fs.Open`, where `fs` is of type `afero.Fs`. Updated the `noling` comment accordingly.

## License Acceptance

### DIFF
--- a/cmd/finch/main.go
+++ b/cmd/finch/main.go
@@ -90,7 +90,7 @@ var newApp = func(logger flog.Logger, fp path.Finch, fs afero.Fs, fc *config.Fin
 	)
 
 	// append nerdctl commands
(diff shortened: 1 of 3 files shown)


pendo324 (approved):
LGTM. Thanks for following up on the comment from #158

PR #214, opened 2023-02-10
feat: adds option to remove persistent user data

sam-berning:
Provides a native way for Finch to delete a persistent user data disk. Calling `finch vm remove --user-data` will now cleanup both the VM and any persistent user data.

Issue #, if available:

*Description of changes:*

Provides a native way for Finch to delete a persistent user data disk. Calling `finch vm remove --user-data` will now cleanup both the VM and any persistent user data.

*Testing done:*

E2E, Unit, and manual testing

```
$ ./_output/bin/finch vm init
$ ./_output/bin/finch vm stop
$ ./_output/bin/finch vm remove --user-data
$ ls ./_output/lima/data/_disks/finch
ls: ./_output/lima/data/_disks/finch: No such file or directory
```

#### License Acceptance


pendo324:
I like the idea behind this PR, but it got me wondering if we need an entire disk lifecycle management command instead of just tacking it on to the vm commands. Like, do we need `finch disk remove` or `finch disk recreate`? Or I guess a better question: is there a use-case for deleting a disk _without_ removing the entire VM?

sam-berning:
> Or I guess a better question: is there a use-case for deleting a disk _without_ removing the entire VM?

Theoretically, users could also remove the disk on `finch vm stop` safely, and they might want to do that. Another case would be if a user runs `finch vm remove` without `--user-data`, they might later want to delete user data.

The reason I chose to attach this to `finch vm remove` was because persistent disk is invisible to users by design, and having a separate management command for the disk felt kind of counter to that. I think you're right that adding a separate management command would make this more flexible, perhaps as a `vm` subcommand though, such as `finch vm clear-user-data`?

The other option would be to also add a similar flag to `finch vm stop` and changing the logic such that the user data cleanup will still happen even if the VM is already in the state it's supposed to transition to.

Not sure which one I like better, WDYT?

Shubhranshu153:
Stale PR. Closing it

PR #227, opened 2023-02-17
ci: Change build artifact name for upload build to s3

KevinLiAWS:
Issue #, if available:

*Description of changes:*
Make github action upload build to S3 artifact's name with github tag or branch name, make it possible for the caller to get a specific build version of finch
*Testing done:*
Tested the action on this branch and result is expected

#### License Acceptance


.github/workflows/upload-build-to-s3.yaml
```diff
@@ -25,22 +25,22 @@ jobs:
...
 
-      - name: Make macos arm64 build
+      - name: Make macos aarch64 build
         run: |
           brew install go lz4 automake autoconf libtool
           make clean
           make download-licenses 
           make FINCH_OS_IMAGE_LOCATION_ROOT=/Applications/Finch.app/Contents/Resources/resources
-          tar -zcvf finch.arm64."$(date '+%s').tar.gz" _output
+          tar -zcvf finch.${GITHUB_REF_NAME}.aarch64.tar.gz _output
```
AnqiPang, line 34: How should this workflow be triggered manually? From main branch or a tag? Since we are using the variable $GITHUB_REF_NAME, which is the short ref name of the branch or tag that triggered the workflow run. I think we should make it clear how this workflow be triggered in the documentation.  The value of the $GITHUB_REF_NAME is not unique, will the file uploaded to s3 overwrite the existing file with same filename?

KevinLiAWS: Yes, when triggering this action, we need to feed with it with the version tag. It is expected that a newer trigger of the same version override the existing one in S3 bucket, since the build itself is a temp file and only newer triggered is consumed from the downstream. In other words, after the build artifact is consumed from the downstream, it can even be deleted to save space in S3 bucket.

KevinLiAWS: GITHUB_REF_NAME is the same usage in this action: https://github.com/runfinch/finch/blob/main/.github/workflows/release-installer.yaml

AnqiPang: What is the use case when trigger the workflow from a branch? I know triggering from a tag is for release purpose.

KevinLiAWS: @AnqiPang if we want to pre-test the a new feature with installer before a new release, we can trigger the build from the latest main, for example, the persistent disk bug bash installer was generated from this route.

KevinLiAWS: One more point is that, the previous naming only have the timestamp in the name, and when the signing tool to fetch the artifact from S3, it will look for the latest artifact, and can't distinguish the version from the artifact. With the new naming with the tag/branch, the signing tool can easily get the expected build by name instead of a meaningless timestamp.

AnqiPang (approved):

vsiravar (approved):
LGTM. We should probably have a clean up in a different PR to delete the older builds from the dependency bucket.

PR #228, opened 2023-02-18
ci: fix release homebrew workflow

AnqiPang:
Issue #, if available:

*Description of changes:*

*Testing done:*

#### License Acceptance

### DIFF
--- a/.github/workflows/release-homebrew.yaml
+++ b/.github/workflows/release-homebrew.yaml
@@ -50,8 +50,11 @@ jobs:
     timeout-minutes: 60
     env:
       FINCH_TAG: ${{ needs.get-latest-tag.outputs.tag }}
-      FINCH_VESRION: ${{ needs.get-latest-tag.outputs.version }}
+      FINCH_VERSION: ${{ needs.get-latest-tag.outputs.version }}
     steps:
+      - uses: actions/setup-go@v3
+        with:
+          go-version: 1.19.x
       - uses: actions/checkout@v3
         with:
           ref: ${{ env.FINCH_TAG }}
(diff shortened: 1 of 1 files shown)


.github/workflows/release-homebrew.yaml
```diff
@@ -50,8 +50,11 @@ jobs:
     timeout-minutes: 60
     env:
       FINCH_TAG: ${{ needs.get-latest-tag.outputs.tag }}
-      FINCH_VESRION: ${{ needs.get-latest-tag.outputs.version }}
+      FINCH_VERSION: ${{ needs.get-latest-tag.outputs.version }}
     steps:
+      - uses: actions/setup-go@v3
```
vsiravar, line 55: Do we need to setup `go` for `get-latest-tag` job?

vsiravar (commented):
The tests did not run for this PR for some reason, although they may not matter. Maybe closing and reopening the PR might help.

.github/workflows/release-homebrew.yaml
AnqiPang: No I don't think so. Here is an example of using the action `WyriHaximus/github-action-get-previous-tag` https://github.com/WyriHaximus/github-action-get-previous-tag#example

vsiravar: Okay, then maybe we can just remove the `- uses: actions/setup-go@v3` step.

AnqiPang: The setup-go action is used in the `macos-arm64-test-installer` and `macos-amd64-test-installer` jobs, which is same as the [`release-installer` workflow](https://github.com/runfinch/finch/blob/main/.github/workflows/release-installer.yaml#L42). It should work without the `setup-go` action because self-hosted runners have installed `go` via brew. But it might be used to set up go env when `go` is not installed.

vsiravar: My bad, I thought `uses: actions/setup-go@v3` was in `get-latest-tag` job :(. Yeah you are right, its needed for `macos-arm64-test-installer` and `macos-amd64-test-installer`.

vsiravar (approved):

Issue #196, opened 2023-01-30
Support rootful containers in finch


monirul:
**What is the problem you're trying to solve?.**
Currently finch runs containers in rootless mode using nerdctl and containerd. This prevents users to set `memlock` in compose yml.

**Describe the feature you'd like**
Finch should support rootful mode.

**Additional context**
Rootless container: A rootless container is a container that could be run without root privileges in the host. 
Rootful container: A rootful container is a container run by root in the host. Since the container is run by root, it has access to all the functionality that root has.

GitHub issue related to memlock:
https://github.com/runfinch/finch/issues/186

ningziwen:
We can either add an config to support rootful mode or switch to rootful mode by default. That needs us to weigh the pros and cons between rootful/rootless.

About the pros of rootful over rootless, other than https://github.com/runfinch/finch/issues/186, https://github.com/runfinch/finch/issues/176 may also be resolved by switching to rootful mode. They need to be further validated in new e2e tests.

ningziwen:
This should also be resolved. https://github.com/runfinch/finch/issues/32

PR #232, opened 2023-02-20
fix: switch from rootless containers to rootful containers

vsiravar:
Issue #, if available:
Fixes https://github.com/runfinch/finch/issues/196
*Description of changes:*
Change finch vm configuration to support rootful containers. Opening a draft PR to get early feedback while investigating  https://github.com/lima-vm/lima/issues/1376.
*Testing done:*

2 e2e tests are failing due to https://github.com/lima-vm/lima/issues/1376. 

#### Retains volumes after vm is restarted
 ```
 $ finch vm stop && finch vm remove && finch vm init
 $ finch run --name test-container -v ~/workplace:/workplace alpine ls /workplace
 testfiles
 $ finch vm stop && finch vm remove && finch vm init
 $ finch start --attach test-container              
testfiles
 ```

 #### Retains created network after vm is reinitialized 
 ```
 $ finch vm stop && finch vm remove && finch vm init
$ finch network create my-bridge-network-1
756c515526ed29831c59d9f7d1b1142ec3e440b19096dc224ee25e9a883d00c1
 $ finch network ls
NETWORK ID      NAME                   FILE
17f29b073143    bridge                 /etc/cni/net.d/nerdctl-bridge.conflist
d2f8cc31a256    my-bridge-network-1    /etc/cni/net.d/nerdctl-my-bridge-network-1.conflist
ce498ce9f1f0    my-bridge-network      /etc/cni/net.d/nerdctl-my-bridge-network.conflist

$ finch vm stop && finch vm remove && finch vm init
$ finch network ls                                 
NETWORK ID      NAME                   FILE
17f29b073143    bridge                 /etc/cni/net.d/nerdctl-bridge.conflist
d2f8cc31a256    my-bridge-network-1    /etc/cni/net.d/nerdctl-my-bridge-network-1.conflist
ce498ce9f1f0    my-bridge-network      /etc/cni/net.d/nerdctl-my-bridge-network.conflist

# Connect container to the network
$ finch run --network my-bridge-network-1 alpine
``` 

#### Test setup
1. Install finch 0.4.0 from Installer(rootless).
2. Create images, containers, networks and volumes.
```
$ finch pull alpine 
$ finch run alpine
$ finch network create rootless-network
$ finch run --name test-container -v ~/workplace:/workplace alpine ls /workplace
$ finch images
REPOSITORY    TAG       IMAGE ID        CREATED          PLATFORM          SIZE       BLOB SIZE
alpine        latest    69665d02cb32    2 minutes ago    linux/arm64/v8    7.8 MiB    3.1 MiB
$ finch ps -a
CONTAINER ID    IMAGE                              COMMAND            CREATED               STATUS                           PORTS    NAMES
1a1684bba87e    docker.io/library/alpine:latest    "/bin/sh"          2 minutes ago         Exited (0) 2 minutes ago                  alpine-1a168
924bd7b54549    docker.io/library/alpine:latest    "ls /workplace"    About a minute ago    Exited (0) About a minute ago 
```
3. Stop and remove vm.
4. Build finch from this branch(rootful) and replace the finch binary under /Applications/Finch/bin/.
5. Edit /Applications/Finch/os/finch.yaml with finch.yaml file from this branch.
6. Start the vm and check its running in rootful mode.
```
$ LIMA_HOME=/Applications/Finch/lima/data /Applications/Finch/lima/bin/limactl shell finch 
$ ps aux | grep containerd
root        1595  0.0  0.4 757868 33160 ?        Ssl  17:41   0:00 /usr/local/bin/containerd-stargz-grpc --log-level=debug --config=/etc/containerd-stargz-grpc/config.toml
root        1720  0.3  0.5 1419508 42832 ?       Ssl  17:41   0:00 /usr/local/bin/containerd
root        1911  0.0  0.1 720460  9640 ?        Sl   17:41   0:00 /usr/local/bin/containerd-shim-runc-v2 -namespace finch -id 1a1684bba87ea1d5a73285ae29ad890c91c27f74a9c7c43f236d34245fa4cc5c -address /run/containerd/containerd.sock
root        2142  0.0  0.1 720716 10284 ?        Sl   17:42   0:00 /usr/local/bin/containerd-shim-runc-v2 -namespace finch -id 
```
7. Run tests
```
$ finch images  
REPOSITORY    TAG       IMAGE ID        CREATED           PLATFORM          SIZE       BLOB SIZE
alpine        latest    69665d02cb32    10 minutes ago    linux/arm64/v8    7.8 MiB    3.1 MiB
$ finch ps -a   
CONTAINER ID    IMAGE                              COMMAND            CREATED           STATUS     PORTS    NAMES
1a1684bba87e    docker.io/library/alpine:latest    "/bin/sh"          10 minutes ago    Created             alpine-1a168
924bd7b54549    docker.io/library/alpine:latest    "ls /workplace"    10 minutes ago    Created             test-container
$ finch start 1a1684bba87e
1a1684bba87e

$ finch network ls
NETWORK ID      NAME                FILE
17f29b073143    bridge              /etc/cni/net.d/nerdctl-bridge.conflist
4b2aad24cab9    rootless-network    /etc/cni/net.d/nerdctl-rootless-network.conflist
                host                
                none                
```
$ finch start --attach test-container 
finch
```


#### License Acceptance


finch.yaml
```diff
@@ -148,27 +148,26 @@ provision:
 - mode: user
   script: |
     #!/bin/bash
+    # We need create /etc/nerdctl directory as root and make it writeable by non root user because the sftp client in https://github.com/runfinch/finch/blob/main/pkg/config/nerdctl_config_applier.go#L140 cannot run as root.
```
vsiravar, line 151: Adding some additional context, the private key used for authenticating the client to the server(lima instance) is created during `limactl start finch` and the key cannot be used by root user.

pkg/config/nerdctl_config_applier.go
```diff
@@ -145,7 +145,7 @@ func (nca *nerdctlConfigApplier) Apply(remoteAddr string) error {
 	sftpFs := sftpfs.New(sftpClient)
 
 	// rootless hardcoded to true for now to match our os.yaml file
```
ningziwen, line 147: The comment needs to change.

finch.yaml
davidhsingyuchen: > Also note thatlimactl [cannot be run as root ](https://github.com/lima-vm/lima/blob/master/cmd/limactl/main.go#L77).

This should refer to `sudo limactl ...` instead of `limactl shell finch sudo ...`, and the latter is what is we want here, so it's a bit unclear why it's relevant here.

> the key cannot be used by root user.

What if we replace these 4 lines with `sudo cp $USER/.ssh/authorized_keys /root/.ssh/authorized_keys` and change the SSH user to `root` [here](https://github.com/runfinch/finch/blob/4a33a2e63c72e687030c70b13dcc5c5748606b86/pkg/config/nerdctl_config_applier.go#L130)?

pkg/config/nerdctl_config_applier.go
vsiravar: Updated.

finch.yaml
```diff
@@ -148,27 +148,26 @@ provision:
...
+    sudo mount --bind /mnt/lima-finch/nerdctl /var/lib/nerdctl
 
     # https://github.com/containerd/nerdctl/blob/main/docs/dir.md#netconfpath
-    mkdir -p /mnt/lima-finch/cni-config ~/.config/cni
-    sudo mount --bind /mnt/lima-finch/cni-config ~/.config/cni
+    sudo mkdir -p /mnt/lima-finch/cni-config /etc/cni/
+    sudo mount --bind /mnt/lima-finch/cni-config /etc/cni/
+
+    sudo mkdir -p /var/lib/containerd
+    sudo mount --bind /mnt/lima-finch /var/lib/containerd
```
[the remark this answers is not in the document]
vsiravar: Added reference doc. `/etc/containerd` has containerd related config files, in finch's case we have  [/etc/containerd/config.toml](https://github.com/containerd/containerd/blob/main/docs/man/containerd-config.toml.5.md#synopsis)  which is generated when the vm is initialized, we can't change this state unless the user logs into the vm and makes changes to this config. I am not sure if we want to save this data.

finch.yaml
vsiravar: > Also note thatlimactl [cannot be run as root ](https://github.com/lima-vm/lima/blob/master/cmd/limactl/main.go#L77).
What I meant was the `~/.ssh/authorized_keys` is created in the home directory of the non root user as `limactl` can't be run as root. If we could run `sudo limactl` then `authorized_keys` would be placed under `/root/.ssh/`. The suggestion you made below fixes this.
>the key cannot be used by root user.
This is a great idea 👍 , made appropriate changes.

finch.yaml
```diff
@@ -148,27 +148,25 @@ provision:
 - mode: user
   script: |
     #!/bin/bash
-    sudo chown $USER /mnt/lima-finch
 
-    # https://github.com/containerd/nerdctl/blob/cffdf87ff4d648a5344eea1406bb95ca3ad7eaa4/extras/rootless/containerd-rootless.sh#L144-L146
-    # XDG_DATA_HOME & ~/.local/share: https://github.com/containerd/nerdctl/blob/cffdf87ff4d648a5344eea1406bb95ca3ad7eaa4/extras/rootless/containerd-rootless.sh#L51
-    mkdir -p /mnt/lima-finch/containerd ~/.local/share/containerd
-    sudo mount --bind /mnt/lima-finch/containerd ~/.local/share/containerd
+    # To allow root user to ssh into the vm.
```
davidhsingyuchen, line 152:
```suggestion
    # Enable SSHing into the VM as root (e.g., in `nerdctlConfigApplier.Apply`).
```

finch.yaml
davidhsingyuchen: > What I meant was the ~/.ssh/authorized_keys is created in the home directory of the non root user as limactl can't be run as root. If we could run sudo limactl then authorized_keys would be placed under /root/.ssh/.

Ah I see! Thank you for the explanation.

> This is a great idea 👍 , made appropriate changes.

Nice! cc @pendo324 as I'm not 100% sure if there's any side effect of doing so.

finch.yaml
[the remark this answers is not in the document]
vsiravar: > Could you do some manual verification for the test cases not yet covered in [additional_disk_test.go](https://github.com/runfinch/finch/blob/main/e2e/vm/additional_disk_test.go) and paste the test result into PR description just in case?

I manually added tests in the description. Lmk if this is sufficient. 

> Another option is to open another PR to add missing test cases

Sure I can do this, is it okay if I did it in a separate PR after this is merged?

> BTW, what about /var/lib/cni 

This directory stores config files and [results of interaction between](https://github.com/containernetworking/cni/blob/main/SPEC.md) containerd and cni plugins. `/var/lib/cni/networks` is where the config files are stored. For example `/var/lib/cni/networks/bridge/10.4.0.2` contains config for 10.4.0.2 which is the IP given by the CNI plugin to a container.  Each time a container stops are starts it gets a new IP, so may not be too useful to save this info. `/var/lib/cni/results` contains [results](https://github.com/containernetworking/cni/blob/main/SPEC.md#Section-5-Result-Types) when containerd passes parameters to the cni plugin, so maybe its okay to not persist this data.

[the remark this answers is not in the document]
davidhsingyuchen: > I manually added tests in the description. Lmk if this is sufficient.

LGTM

> Sure I can do this, is it okay if I did it in a separate PR after this is merged?

Sure.

> This directory stores config files and [results of interaction between](https://github.com/containernetworking/cni/blob/main/SPEC.md) containerd and cni plugins. /var/lib/cni/networks is where the config files are stored. For example /var/lib/cni/networks/bridge/10.4.0.2 contains config for 10.4.0.2 which is the IP given by the CNI plugin to a container. Each time a container stops are starts it gets a new IP, so may not be too useful to save this info. /var/lib/cni/results contains [results](https://github.com/containernetworking/cni/blob/main/SPEC.md#Section-5-Result-Types) when containerd passes parameters to the cni plugin, so maybe its okay to not persist this data.

I see. As long as we are 100% sure nothing there will persist across container-stop-and-start cycle, then we don't have to include that directory.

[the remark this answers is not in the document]
vsiravar: > Maybe we can keep this PR only changing to rootful with still persisting all the directories that were persisted, and then after making sure that everything works fine, we can create a follow-up PR to remove the directories that seem unnecessary?

Sounds good to me. Just making sure that we are not persisting anything under `~/.local` which is not used by rootful mode.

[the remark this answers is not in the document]
vsiravar: One thing to note though is `/var/lib/cni/results` can bloat as described [here](https://github.com/cri-o/cri-o/issues/4296) as a new file is created whenever a container is created.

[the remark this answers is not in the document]
davidhsingyuchen: > Just making sure that we are not persisting anything under ~/.local which is not used by rootful mode right?
 
Yes. I should have made it clear that I was referring to the counterpart, not the exact same directory.

finch.yaml
pendo324: As long as we _copy_ the `authorized_keys` file instead of _moving_ it (which is what your recommendation says), I don't see any issue with this approach. As far as I can tell, the only consequence is that you can SSH into your own VM as your own user, and as root. Since your user can already use `sudo` freely in the VM, I don't see a problem with that.

finch.yaml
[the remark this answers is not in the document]
vsiravar: Thanks! I updated the bind mounts to be consistent with the rootless config.

finch.yaml
```diff
@@ -148,27 +148,28 @@ provision:
...
 
     # https://github.com/containerd/nerdctl/blob/main/docs/dir.md#netconfpath
-    mkdir -p /mnt/lima-finch/cni-config ~/.config/cni
-    sudo mount --bind /mnt/lima-finch/cni-config ~/.config/cni
+    sudo mkdir -p /mnt/lima-finch/cni-config /etc/cni/
+    sudo mount --bind /mnt/lima-finch/cni-config /etc/cni/
+
+    # https://github.com/containerd/containerd/blob/main/docs/ops.md#base-configuration
+    sudo mkdir -p /var/lib/containerd
+    sudo mount --bind /mnt/lima-finch/containerd /var/lib/containerd
```
davidhsingyuchen, line 166: nit: move this block to be where the `~/.local/share/containerd` part was (i.e., right after `sudo chown $USER /mnt/lima-finch`) may make it easier to review.

davidhsingyuchen (commented):
LGTM re. the current code, thanks!

ningziwen:
Should this be feat! (bumps minor version) or fix (bumps patch version)?

Following semantic versioning, my thought is the switching implementation details but not impacting external features should be fix, even if rootful is a big switch.

However, this broke two existing e2e tests, but these 2 e2e tests are specifically to how we implement the tests. It only happens in using netcat to run a server in container and test port forwarding. The reason is when container starts in rootful mode, it will ping the netcat port for one time, and netcat will exit after being pinged for one time. We are [changing](https://github.com/runfinch/common-tests/commit/7a6a6c36d11796b2048d90353f06d25013b132a8) from netcat to nginx and they would all passed because nginx can be consistently pinged.

estesp:
I'm more inclined to `fix` over `feat!` by nature of the fact that end user interfaces and usage does not change at all for the user. Some user's experiences with finch will improve due to the removal of the underlying limitations of rootless, which definitely would be seen as a "fix" by users, not a new feature.

davidhsingyuchen (commented):
Maybe we can change the PR title, which will be commit msg later, to be something like `switch from rootless containers to rootful containers`. The current title may not be clear that all the containers are going to be rootful, not just supported.

pendo324:
If the change is transparent to users, meaning that users _before_ the change will not have their VM stop working, and users _after_ the change will still have their `vm init`/`vm start` working without any errors or changes needed, then I don't think it's a breaking change, and we can use fix!.

Yes, we had to refactor the test, but most users aren't running containers that only publish a message on a port once ("one-shot" containers), so I actually think the new test is more representative of the actual usage of the `-p` option, and we should treat the followup into why the behavior is different between rootless and rootfull as a separate issue.

However, in my testing, this does (or at least can) break users with existing VMs when they update and run `vm start` again. At least I hit an error case like this:

```
$ ./_output/bin/finch vm start
INFO[0000] Starting existing Finch virtual machine...
INFO[0028] Finch virtual machine started successfully
FATA[0028] failed to setup ssh client: ssh: handshake failed: ssh: unable to authenticate, attempted methods [none publickey], no supported methods remain
$ ./_output/bin/finch images
FATA[0000] cannot access containerd socket "/run/containerd/containerd.sock": no such file or directory
FATA[0000] exit status 1
```

Repro is:
1. init a new VM with the latest main
2. stop it
3. pull this PR
4. `make finch`
5. try to run `vm start`

davidhsingyuchen:
> this does (or at least can) break users with existing VMs when they update and run vm start again

A Finch user will use an installer to update their Finch, and the existing VM will be erased by the installer, so this problem may not happen for normal users? In other words, it should only impact Finch devs.

Wondering this:

1. A user has some containers and images.
2. Run Finch installer to upgrade their Finch to a version that contains this change.
3. `finch vm init`
4. Will the previous containers and images be usable?

Since we're mounting the rootful counterparts into the same location in the persistent disk, theoretically it should work, but maybe we should verify that because if it's not working, then it'll be a breaking change for our users.

vsiravar:
> Wondering this:
> 
> 1. A user has some containers and images.
> 2. Run Finch installer to upgrade their Finch to a version that contains this change.
> 3. `finch vm init`
> 4. Will the previous containers and images be usable?
> 
> Since we're mounting the rootful counterparts into the same location in the persistent disk, theoretically it should work, but maybe we should verify that because if it's not working, then it'll be a breaking change for our users.

I tried this test. The test was setup using

1. Install finch 0.4.0 from Installer(rootless).
2. Create images, containers, networks and volumes. 
3. Stop and remove vm. 
4. Build finch from this branch(rootful) and replace the finch binary under `/Applications/Finch/bin/`. 
5. Edit `/Applications/Finch/os/finch.yaml` with `finch.yaml` file from this branch. 
6. Start the vm. 

Observations. 
1. The images and containers persist between rootless and rootful. 
2. Images can be run without errors. However while starting the containers using `finch start` command, there is an error complaining about mounts. Error log below
```

**Comment by pendo324** [comment]
> > this does (or at least can) break users with existing VMs when they update and run vm start again
> 
> A Finch user will use an installer to update their Finch, and the existing VM will be erased by the installer, so this problem may not happen for normal users? In other words, it should only impact Finch devs.

That's a good point, the scenario I posted can only really happen to devs. We can include a mention in the release notes to say that it's expected and also list how to fix it.

> Wondering this:
> 
> 1. A user has some containers and images.
> 2. Run Finch installer to upgrade their Finch to a version that contains this change.
> 3. `finch vm init`
> 4. Will the previous containers and images be usable?
> 
> Since we're mounting the rootful counterparts into the same location in the persistent disk, theoretically it should work, but maybe we should verify that because if it's not working, then it'll be a breaking change for our users.

I just tested this (basically just removing my dev vm before re-init), and it worked fine fwiw. I think we already have e2e tests for persistent volume, but not across versions so this sounds like something we will have to manually verify.

**Comment by pendo324** [comment]
> Observations.
> 
> 1. The images and containers persist between rootless and rootful.
> 2. Images can be run without errors. However while starting the containers using `finch start` command, there is an error complaining about mounts. Error log below
> 
> ```
> % finch start  77a9e1787dff
> FATA[0000] failed to create shim task: OCI runtime create failed: runc create failed: unable to start container process: error during container init: error mounting "/home/siravara.linux/.local/share/nerdctl/1935db59/containers/finch/77a9e1787dffb48651a2a72204636d604af437aa63fd0d53fa69d65cd0291839/hostname" to rootfs at "/etc/hostname": stat /home/siravara.linux/.local/share/nerdctl/1935db59/containers/finch/77a9e1787dffb48651a2a72204636d604af437aa63fd0d53fa69d65cd0291839/hostname: no such file or directory: unknown 
> ```
> 
> This would be a breaking change for users who want to start their saved containers after upgrading to rootful.

Interesting, I guess my tests weren't as in depth as yours. I think we can actually fix that by mounting the persistent data volume to both the old rootless and new rootfull locations. Can you try that out?

File: finch.yaml [hunk 6]
```diff
@@ -148,27 +148,28 @@ provision:
...
+
+    # Enable SSHing into the VM as root (e.g., in `nerdctlConfigApplier.Apply`).
+    sudo cp ~/.ssh/authorized_keys /root/.ssh/
     sudo chown $USER /mnt/lima-finch
 
-    # https://github.com/containerd/nerdctl/blob/cffdf87ff4d648a5344eea1406bb95ca3ad7eaa4/extras/rootless/containerd-rootless.sh#L144-L146
-    # XDG_DATA_HOME & ~/.local/share: https://github.com/containerd/nerdctl/blob/cffdf87ff4d648a5344eea1406bb95ca3ad7eaa4/extras/rootless/containerd-rootless.sh#L51
-    mkdir -p /mnt/lima-finch/containerd ~/.local/share/containerd
-    sudo mount --bind /mnt/lima-finch/containerd ~/.local/share/containerd
+   # https://github.com/containerd/containerd/blob/main/docs/ops.md#base-configuration
```
pendo324, line 156: Missing one space at the beginning of this line
```suggestion
    # https://github.com/containerd/containerd/blob/main/docs/ops.md#base-configuration
```

vsiravar: Updated.

vsiravar: > Interesting, I guess my tests weren't as in depth as yours. I think we can actually fix that by mounting the persistent data volume to both the old rootless and new rootfull locations. Can you try that out?

This actually works. However should we make this change though just to support rootless containers which were created prior to this change?

vsiravar:
> Interesting, I guess my tests weren't as in depth as yours. I think we can actually fix that by mounting the persistent data volume to both the old rootless and new rootfull locations. Can you try that out?
This actually works. However should we make this change though just to support rootless containers which were created prior to this change?

pendo324:
> > Interesting, I guess my tests weren't as in depth as yours. I think we can actually fix that by mounting the persistent data volume to both the old rootless and new rootfull locations. Can you try that out?
> 
> This actually works. However should we make this change though just to support rootless containers which were created prior to this change?

Thanks for looking into it. I think its ok to mount both locations, it shouldn't hurt anything else 👍. Why we do it might be slightly confusing to more advanced users though, so just document it as a comment in the finch.yaml provisioning script.

davidhsingyuchen:
From https://github.com/runfinch/finch/pull/232#issuecomment-1448976237:

```sh
FATA[0000] failed to create shim task: OCI runtime create failed: runc create failed: unable to start container process: error during container init: error mounting "/home/siravara.linux/.local/share/nerdctl/1935db59/containers/finch/77a9e1787dffb48651a2a72204636d604af437aa63fd0d53fa69d65cd0291839/hostname" to rootfs at "/etc/hostname": stat /home/siravara.linux/.local/share/nerdctl/1935db59/containers/finch/77a9e1787dffb48651a2a72204636d604af437aa63fd0d53fa69d65cd0291839/hostname: no such file or directory: unknown 
```

In the error message, `/home/siravara.linux/.local/share/nerdctl/1935db59/containers/finch/77a9e1787dffb48651a2a72204636d604af437aa63fd0d53fa69d65cd0291839/hostname` should come from [`src`](https://github.com/containerd/nerdctl/blob/cc1b6e0da1c65e77f5f5f732ef34ea0c5eb3dcec/cmd/nerdctl/container_run_network.go#L97), and the path prefix (i.e., `/home/siravara.linux/.local/share/nerdctl`) comes from [`DataRoot()`](https://github.com/containerd/nerdctl/blob/cc1b6e0da1c65e77f5f5f732ef34ea0c5eb3dcec/pkg/defaults/defaults_linux.go#L34).

Question: when the error is returned, we should be already running rootful containerd, and why is the prefix still the rootless one? Or why does `rootlessutil.IsRootless()` return `true`?

pendo324:
> From [#232 (comment)](https://github.com/runfinch/finch/pull/232#issuecomment-1448976237):
> 
> ```shell
> FATA[0000] failed to create shim task: OCI runtime create failed: runc create failed: unable to start container process: error during container init: error mounting "/home/siravara.linux/.local/share/nerdctl/1935db59/containers/finch/77a9e1787dffb48651a2a72204636d604af437aa63fd0d53fa69d65cd0291839/hostname" to rootfs at "/etc/hostname": stat /home/siravara.linux/.local/share/nerdctl/1935db59/containers/finch/77a9e1787dffb48651a2a72204636d604af437aa63fd0d53fa69d65cd0291839/hostname: no such file or directory: unknown 
> ```
> 
> In the error message, `/home/siravara.linux/.local/share/nerdctl/1935db59/containers/finch/77a9e1787dffb48651a2a72204636d604af437aa63fd0d53fa69d65cd0291839/hostname` should come from [`src`](https://github.com/containerd/nerdctl/blob/cc1b6e0da1c65e77f5f5f732ef34ea0c5eb3dcec/cmd/nerdctl/container_run_network.go#L97), and the path prefix (i.e., `/home/siravara.linux/.local/share/nerdctl`) comes from [`DataRoot()`](https://github.com/containerd/nerdctl/blob/cc1b6e0da1c65e77f5f5f732ef34ea0c5eb3dcec/pkg/defaults/defaults_linux.go#L34).
> 
> Question: when the error is returned, we should be already running rootful containerd, and why is the prefix still the rootless one? Or why does `rootlessutil.IsRootless()` return `true`?

Isn't it because @vsiravar was trying to run a container created when rootless mode was enabled _after_ rootfull mode was in use?

It looks like for existing containers, nerdctl just gets the data from the state files / filesystem?

https://github.com/containerd/nerdctl/blob/cc1b6e0da1c65e77f5f5f732ef34ea0c5eb3dcec/cmd/nerdctl/container_run_network.go#L183-L240

vsiravar:
> 

Not entirely sure of this though, however I shelled into the vm and checked, `containerd` is running as root and I still get this error when I try `sudo nerdctl start ed678d167f70` . 

```
$ ps aux | grep containerd
root        1480  0.0  0.4 757868 33500 ?        Ssl  00:50   0:00 /usr/local/bin/containerd-stargz-grpc --log-level=debug --config=/etc/containerd-stargz-grpc/config.toml
root        1610  0.3  0.5 1345716 42144 ?       Ssl  00:50   0:01 /usr/local/bin/containerd
siravara    2041  0.0  0.0   6040  1824 pts/0    S+   00:57   0:00 grep --color=auto containerd
```
Also I can't run `nerdctl` as a non-root user. i get this error. 

```
$ nerdctl start ed678d167f70
FATA[0000] rootless containerd not running? (hint: use `containerd-rootless-setuptool.sh install` to start rootless containerd): stat /run/user/504/containerd-rootless: no such file or directory 
```
I can create another ticket to investigate it outside this PR if thats okay.

davidhsingyuchen:
> > From [#232 (comment)](https://github.com/runfinch/finch/pull/232#issuecomment-1448976237):
> > ```shell
> > FATA[0000] failed to create shim task: OCI runtime create failed: runc create failed: unable to start container process: error during container init: error mounting "/home/siravara.linux/.local/share/nerdctl/1935db59/containers/finch/77a9e1787dffb48651a2a72204636d604af437aa63fd0d53fa69d65cd0291839/hostname" to rootfs at "/etc/hostname": stat /home/siravara.linux/.local/share/nerdctl/1935db59/containers/finch/77a9e1787dffb48651a2a72204636d604af437aa63fd0d53fa69d65cd0291839/hostname: no such file or directory: unknown 
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > In the error message, `/home/siravara.linux/.local/share/nerdctl/1935db59/containers/finch/77a9e1787dffb48651a2a72204636d604af437aa63fd0d53fa69d65cd0291839/hostname` should come from [`src`](https://github.com/containerd/nerdctl/blob/cc1b6e0da1c65e77f5f5f732ef34ea0c5eb3dcec/cmd/nerdctl/container_run_network.go#L97), and the path prefix (i.e., `/home/siravara.linux/.local/share/nerdctl`) comes from [`DataRoot()`](https://github.com/containerd/nerdctl/blob/cc1b6e0da1c65e77f5f5f732ef34ea0c5eb3dcec/pkg/defaults/defaults_linux.go#L34).
> > Question: when the error is returned, we should be already running rootful containerd, and why is the prefix still the rootless one? Or why does `rootlessutil.IsRootless()` return `true`?
> 
> Isn't it because @vsiravar was trying to run a container created when rootless mode was enabled _after_ rootfull mode was in use?
> 
> It looks like for existing containers, nerdctl just gets the data from the state files / filesystem?
> 
> https://github.com/containerd/nerdctl/blob/cc1b6e0da1c65e77f5f5f732ef34ea0c5eb3dcec/cmd/nerdctl/container_run_network.go#L183-L240
> 
> New containers shouldn't have this problem from what I understand, just containers created before the switch to rootfull.

You're right. I was looking at the flow of `nerdctl run`, while I should be looking at that of `nerdctl start`. To be specific, the container was created with Finch v0.4.0, which means that in its "configuration", it's already told to mount that rootless hostname path, so basically the path was already decided before we tried to start that container with Finch built from this PR branch, hence the error.

davidhsingyuchen:
> > > Interesting, I guess my tests weren't as in depth as yours. I think we can actually fix that by mounting the persistent data volume to both the old rootless and new rootfull locations. Can you try that out?
> > 
> > 
> > This actually works. However should we make this change though just to support rootless containers which were created prior to this change?
> 
> Thanks for looking into it. I think its ok to mount both locations, it shouldn't hurt anything else 👍. Why we do it might be slightly confusing to more advanced users though, so just document it as a comment in the finch.yaml provisioning script.

I'm also good with this. However, my assumption is that eventually we still want to remove those rootless mounts to avoid confusion and reduce complexity of our code, and we're delaying that so that the blast radius of it can be minimized. To be specific, I suppose we'd like to add a TODO: in a future release, we will remove those mounts in a `BREAKING CHANGES` PR, and in the release notes, we will say something around this: "Containers created by Finch <= v0.4.0 will be malfunctional after upgrading to this version", and hopefully it will be already long enough since v0.4.0 was out so that the possibility of our users still having such containers running is low.

ningziwen:
> > > > Interesting, I guess my tests weren't as in depth as yours. I think we can actually fix that by mounting the persistent data volume to both the old rootless and new rootfull locations. Can you try that out?
> > > 
> > > 
> > > This actually works. However should we make this change though just to support rootless containers which were created prior to this change?
> > 
> > 
> > Thanks for looking into it. I think its ok to mount both locations, it shouldn't hurt anything else 👍. Why we do it might be slightly confusing to more advanced users though, so just document it as a comment in the finch.yaml provisioning script.
> 
> I'm also good with this. However, my assumption is that eventually we still want to remove those rootless mounts to avoid confusion and reduce complexity of our code, and we're delaying that, so the blast radius of it can be minimized. To be specific, I suppose we'd like to add a TODO: in a future release, we will remove those mounts in a `BREAKING CHANGES` PR, and in the release notes, we will say something around this: "Containers created by Finch <= v0.4.0 will be malfunctional after upgrading to this version", and hopefully it will be already long enough since the next version of v0.4.0 is out so that the possibility of our users still having such containers running is low.

Seems mounting both locations persistently don't add any performance burden or very minimum maintainence burden? Instead of a TODO to remove it as a breaking change, how about adding a comment to say mounting both locations gives the ability of switching between rootful and rootless without breaking change?

davidhsingyuchen:
> > > > > Interesting, I guess my tests weren't as in depth as yours. I think we can actually fix that by mounting the persistent data volume to both the old rootless and new rootfull locations. Can you try that out?
> > > > 
> > > > 
> > > > This actually works. However should we make this change though just to support rootless containers which were created prior to this change?
> > > 
> > > 
> > > Thanks for looking into it. I think its ok to mount both locations, it shouldn't hurt anything else 👍. Why we do it might be slightly confusing to more advanced users though, so just document it as a comment in the finch.yaml provisioning script.
> > 
> > 
> > I'm also good with this. However, my assumption is that eventually we still want to remove those rootless mounts to avoid confusion and reduce complexity of our code, and we're delaying that, so the blast radius of it can be minimized. To be specific, I suppose we'd like to add a TODO: in a future release, we will remove those mounts in a `BREAKING CHANGES` PR, and in the release notes, we will say something around this: "Containers created by Finch <= v0.4.0 will be malfunctional after upgrading to this version", and hopefully it will be already long enough since the next version of v0.4.0 is out so that the possibility of our users still having such containers running is low.
> 
> Seems mounting both locations persistently doesn't add any performance burden or very minimum maintainence burden? Instead of a TODO to remove it as a breaking change, how about adding a comment to say mounting both locations gives the ability of switching between rootful and rootless without breaking change?

Yeah that's fair. I don't have a strong opinion on this. By "confusion & complexity", I was mainly referring to the fact that yet another set of paths that need to be understood by Finch devs.

pendo324 (approved):

vsiravar:
Added tests in  PR description testing persistence of images, containers, volumes and networks between rootless and rootful.

finch.yaml
```diff
@@ -148,27 +148,39 @@ provision:
 - mode: user
   script: |
     #!/bin/bash
-    sudo chown $USER /mnt/lima-finch
 
-    # https://github.com/containerd/nerdctl/blob/cffdf87ff4d648a5344eea1406bb95ca3ad7eaa4/extras/rootless/containerd-rootless.sh#L144-L146
-    # XDG_DATA_HOME & ~/.local/share: https://github.com/containerd/nerdctl/blob/cffdf87ff4d648a5344eea1406bb95ca3ad7eaa4/extras/rootless/containerd-rootless.sh#L51
```
davidhsingyuchen, line 154: Could we add back the references for rootless directories?

vsiravar: Updated.

davidhsingyuchen (approved):

pendo324 (approved):

ningziwen (approved):

vsiravar:
Adding additional observations with forward compatibility. 

I was doing some more testing between rootful and rootless mode. If there is a situation where users who produce rootful containers will for some reason downgrade the version of finch to a rootless version(say 0.4.0) the persisted data will not run in the rootelss context and users might lose persisted data. I found that rootful containers can't run in rootless mode from my test.
Which makes it not forward compatible.
Maybe in the release notes we can mention that "Containers produced using this version and later versions may misbehave in versions <= 0.4.0"?

pendo324:
> Adding additional observations with forward compatibility.
> 
> I was doing some more testing between rootful and rootless mode. If there is a situation where users who produce rootful containers will for some reason downgrade the version of finch to a rootless version(say 0.4.0) the persisted data will not run in the rootelss context and users might lose persisted data. I found that rootful containers can't run in rootless mode from my test. Which makes it not forward compatible. Maybe in the release notes we can mention that "Containers produced using this version and later versions may misbehave in versions <= 0.4.0"?

Yeah, we should definitely mention that in the patch notes. Eventually, we may support both modes and need to come up with a more permanent fix for this (maybe just two different sets of mount points, or two different hosts), but for now, I think it makes sense as a note. We might even want to add it to a doc page/README too.

vsiravar:
> > Adding additional observations with forward compatibility.
> > I was doing some more testing between rootful and rootless mode. If there is a situation where users who produce rootful containers will for some reason downgrade the version of finch to a rootless version(say 0.4.0) the persisted data will not run in the rootelss context and users might lose persisted data. I found that rootful containers can't run in rootless mode from my test. Which makes it not forward compatible. Maybe in the release notes we can mention that "Containers produced using this version and later versions may misbehave in versions <= 0.4.0"?
> 
> Yeah, we should definitely mention that in the patch notes. Eventually, we may support both modes and need to come up with a more permanent fix for this (maybe just two different sets of mount points, or two different hosts), but for now, I think it makes sense as a note. We might even want to add it to a doc page/README too.

Cool will add the below note in the release notes. 
### Note on rootful containers
Containers created by Finch >= v0.4.1 will malfunction in versions <=0.4.0.

pendo324:
> ### Note on rootful containers
> Containers created by Finch >= v0.4.1 will malfunction in versions <=0.4.0.

Looks good, but is it possible to add some more context into why this would occur?

vsiravar:
> > ### Note on rootful containers
> > Containers created by Finch >= v0.4.1 will malfunction in versions <=0.4.0.
> 
> Looks good, but is it possible to add some more context into why this would occur?

Changed to

"Containers created using Finch version 0.4.1 or later may not function properly in versions 0.4.0 or earlier due to how the data of rootful containers is stored with root user permissions. This can cause issues when accessed by older versions of Finch that do not support rootful containers."

PR #255, opened 2023-03-02
build: update os image versions

AnqiPang:
Issue #, if available:

*Description of changes:*

*Testing done:*

- [x] I've reviewed the guidance in CONTRIBUTING.md

#### License Acceptance

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #271, opened 2023-03-07
ci: Add workflow to sync finch submodules

vsiravar:
Issue #, if available: N/A

*Description of changes:*
Add workflow to update submodules in `finch`. 

#### Details
1. Workflow runs daily at 9am UTC. 
2. Has a `workflow_dispatch` trigger for manually running the workflow in addition to the regular daily cadence. 
*Testing done:*
Yes. See https://github.com/vsiravar/finch-public/pull/2
Test with updating `FINCH_OS_BASENAME` : https://github.com/vsiravar/finch-public/pull/4/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52

#### License Acceptance

### DIFF
--- a/.github/bin/update-os-image.sh
+++ b/.github/bin/update-os-image.sh
@@ -1,5 +1,5 @@
 #!/bin/bash
-set -x
+set -euxo pipefail
 # Set OS hash directory in finch-core as base directory for searching latest OS images.
 OS_BASE_IMAGE_HASH_DIR="./deps/finch-core/hashes/"
 
--- a/.github/bin/update-os-image.sh
+++ b/.github/bin/update-os-image.sh
@@ -1,0 +1,24 @@
+#!/bin/bash
+set -x
+# Set OS hash directory in finch-core as base directory for searching latest OS images.
+OS_BASE_IMAGE_HASH_DIR="./deps/finch-core/hashes/"
+
+OS_AARCH64_FILENAME_PATTERN="Fedora-Cloud-Base-.*-.*.aarch64-.*.qcow2"
+
+OS_X86_64_FILENAME_PATTERN="Fedora-Cloud-Base-.*-.*.x86_64-.*.qcow2"
+
+
+# Use wildcard patterns to search for the two files and assign their paths to variables
+AARCH64_FILEPATH=$(find "$OS_BASE_IMAGE_HASH_DIR" -name "Fedora-Cloud-Base-*-*.aarch64-*.qcow2.sha512" -print -quit)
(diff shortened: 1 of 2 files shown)


davidhsingyuchen (commented):
From #255, it seems that `FINCH_OS_BASENAME` has to be in sync with what's under `finch-core/hashes`. I'm not very sure about this, but maybe it makes sense to also update `FINCH_OS_BASENAME` when updating submodules? /cc @pendo324

vsiravar:
> From #255, it seems that `FINCH_OS_BASENAME` has to be in sync with what's under `finch-core/hashes`. I'm not very sure about this, but maybe it makes sense to also update `FINCH_OS_BASENAME` when updating submodules? /cc @pendo324

Makes sense, `finch-core` uses [sha512 to check os base image file integrity ](https://github.com/runfinch/finch-core/blob/main/Makefile#L51) while in the [finch.yaml file we use sha256](https://github.com/runfinch/finch/blob/main/Makefile#L47). I guess that's the reason why we need  `FINCH_OS_BASENAME` in finch's Makefile. Added script to update `FINCH_OS_BASENAME` in the Makefile along with updating submodules. Added test in the description.  cc @davidhsingyuchen @pendo324

pendo324 (approved):
Yeah, the comments in the PR are correct; the name has to be in sync. I'm actually wondering if its possible to remove this condition, but for now, this action looks good 👍

.github/bin/update-os-image.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/bash
+set -x
```
davidhsingyuchen, line 2: More info: https://coderwall.com/p/fkfaqq/safer-bash-scripts-with-set-euxo-pipefail

```suggestion
set -euxo pipefail
```

.github/workflows/sync-submodules.yaml
```diff
@@ -0,0 +1,29 @@
...
+      - name: Update sub modules
+        run: |
+          git submodule update --remote
+          ./.github/bin/update-os-image.sh
+
+      - name: Create PR
+        uses: peter-evans/create-pull-request@v4
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+          commit-message: bump submodules
```
davidhsingyuchen, line 27: nit: Since it doesn't add additional info given the title below, maybe we can just drop it

```suggestion
```

davidhsingyuchen (approved):

PR #308, opened 2023-03-21
docs: Adding instructions of how to authenticate to GCR/AR in FAQ

azhouwd:
Issue #, if available:

*Description of changes:*
Users are not able to authenticate to GCR/AR either by using service account key or authentication plugin.

Authenticating through service account key will save the credentials to an invalid path in the host OS - `/root/.docker/config.json`.

Authenticating through auth plugin - `gcloud auth configure-docker` is not working now. It is supported for nerdctl but that would require us to install gcloud cli and docker in the VM. The implementation of auth plugin would need more analysis.

More info: https://cloud.google.com/artifact-registry/docs/docker/authentication

This PR is to provide an instruction that gives users a quick workaround way to authenticate to GCR/AR.   

*Testing done:*
Manual testing

#### License Acceptance

### DIFF
--- a/README.md
+++ b/README.md
@@ -117,6 +117,21 @@ This section contains frequently-asked questions regarding working with Finch.
 LIMA_HOME=/Applications/Finch/lima/data /Applications/Finch/lima/bin/limactl shell finch
 ```
 
+#### How to authenticate to GCR(Google Container Registry) or AR(Artifact Registry)?
+Create a [GCP Service Account](https://cloud.google.com/iam/docs/service-accounts-create#creating), grant `Owner` role, and download the key as a JSON file to the $HOME directory.
+
+Then shell into the VM by running the following command:
+```sh
+LIMA_HOME=/Applications/Finch/lima/data /Applications/Finch/lima/bin/limactl shell finch
+```
+
+In the VM, go to the directory where the above key JSON file exists and run command:
+```sh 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -117,6 +117,21 @@ This section contains frequently-asked questions regarding working with Finch.
 LIMA_HOME=/Applications/Finch/lima/data /Applications/Finch/lima/bin/limactl shell finch
 ```
 
+#### How to authenticate to GCR(Google Container Registry) or AR(Artifact Registry)?
+Create a [GCP Service Account](https://cloud.google.com/iam/docs/service-accounts-create#creating), grant `Owner` role, and download the key as a JSON file to the $HOME directory.
+
+Then shell into the VM by running the following command:
+```sh
+LIMA_HOME=/Applications/Finch/lima/data /Applications/Finch/lima/bin/limactl shell finch
```
ningziwen, line 125: I think we should not put `vm shell` as part of any user's workflow, unless it is a very temporary workaround. If we want to add this to doc, let's highlight it is workaround and link an issue for the right solution here.

azhouwd: Added link to the issue and specify it's a temp workaround way.

azhouwd:
Closed this PR as the fix was moved to https://github.com/runfinch/finch/pull/315

PR #315, opened 2023-03-23
fix: Correctly set and pick up environment variables

azhouwd:
Issue #, if available:
Currently the environment variable is set in `/root/.bashrc` so that it is [not easily accessible](https://superuser.com/questions/1254422/linux-sudo-env-does-not-have-environment-variable-while-root-user-has) by regular users even by specifying `sudo -E`. Normally we would access the VM by regular user. In order to pick up the environment variables, we need to store the vars in the /home/<user>.linux/.bashrc and added -E to sudo command as sudo itself won't preserve the environment variables.

*Description of changes:*
Added -E to sudo command when we call limactl command.
Update the profile path so that env vars are written in /home/<user>.linux/.bashrc.
Added e2e tests to make sure the content is written correctly into the ~/.finch/.config.json file in host system.
*Testing done:*
Manual testing and E2E tests

#### License Acceptance


ningziwen:
Could you add e2e tests? Can be simply "logout -> ~/.finch/config.json.assertNotContains() -> login -> ~/.finch/config.json.assertContains()", unless you know better ways.

cmd/finch/main.go
```diff
@@ -117,7 +117,7 @@ func virtualMachineCommands(
 		logger,
 		optionalDepGroups,
 		config.NewLimaApplier(fc, fs, fp.LimaOverrideConfigPath()),
-		config.NewNerdctlApplier(fssh.NewDialer(), fs, fp.LimaSSHPrivateKeyPath()),
+		config.NewNerdctlApplier(fssh.NewDialer(), fs, fp.LimaSSHPrivateKeyPath(), system.NewStdLib().Env("USER")),
```
ningziwen, line 120: Maybe refer to the previous rootful change. https://github.com/runfinch/finch/commit/ae43a0779c778cc3656724f84b5cdfd8ab087798#diff-531afb9f97a7384e0de895046f9e14befbebc49c896b080fce4c41c75f14f1b4

The user was populated by `user := nca.systemDeps.Env("USER")`

pkg/config/nerdctl_config_applier.go
```diff
@@ -139,7 +141,7 @@ func (nca *nerdctlConfigApplier) Apply(remoteAddr string) error {
 		return fmt.Errorf("failed to update the nerdctl config file: %w", err)
 	}
 
-	if err := updateEnvironment(sftpFs, user); err != nil {
+	if err := updateEnvironment(sftpFs, nca.hostUser); err != nil {
```
vsiravar, line 144: A comment explaining why we need to update the host users `.bashrc` file would be useful.

cmd/finch/nerdctl.go
```diff
@@ -146,7 +146,7 @@ func (nc *nerdctlCommand) run(cmdName string, args []string) error {
 	}
 	finalArgs = append(finalArgs, nerdctlArgs...)
 
-	limaArgs := append([]string{"shell", limaInstanceName, "sudo", nerdctlCmdName, cmdName}, finalArgs...)
+	limaArgs := append([]string{"shell", limaInstanceName, "sudo", "-E", nerdctlCmdName, cmdName}, finalArgs...)
```
vsiravar, line 149: A comment about why we need -E would be useful here.

cmd/finch/main.go
azhouwd: It's the same thing except we only pass the USER here as we only need it in NewNerdctlApplier

ningziwen: In the previous way, the variable is input from main.go.

azhouwd: Could you please elaborate a bit? This variable is also input from main.go

pkg/config/nerdctl_config_applier.go
azhouwd: I think the comment for `updateEnvironment` has already explained it well?

vsiravar: Yeah that part explains why we set `DOCKER_CONFIG` in `.bashrc`. I was referring to the fact that we need to update the host users `/home/<user>.linux/.bashrc` file vs root's  because `sudo -E` does not source `/root/.bashrc`.

cmd/finch/main.go
ningziwen: Yeah you are right.

pkg/config/nerdctl_config_applier.go
azhouwd: updated

cmd/finch/nerdctl.go
azhouwd: updated

ningziwen (dismissed):

vsiravar (changes requested):
We should add `sudo -E` in the [version command](https://github.com/runfinch/finch/blob/main/cmd/finch/version.go#L92) as well to keep it consistent with the rest of `nerdctl` commands.

azhouwd:
> We should add `sudo -E` in the [version command](https://github.com/runfinch/finch/blob/main/cmd/finch/version.go#L92) as well to keep it consistent with the rest of `nerdctl` commands.

Good catch, updated

ningziwen (dismissed):

vsiravar (commented):
Looks good. `golanci-lint` failed on `version_test.go` due to to max line length. We should fix that before merge.

ningziwen (approved):

vsiravar (dismissed):
LGTM

ningziwen:
> The newly added E2E tests are failing consistent on arm64, I haven't figured out the reasons. Maybe we can add them in separate PR to unblock the bug fix? @ningziwen @vsiravar

@azhouwd Did you run into this in local `make test-e2e`?

azhouwd:
> > The newly added E2E tests are failing consistent on arm64, I haven't figured out the reasons. Maybe we can add them in separate PR to unblock the bug fix? @ningziwen @vsiravar
> 
> @azhouwd Did you run into this in local `make test-e2e`?

No the whole suite works fine locally, I tried several times

vsiravar:
> > > The newly added E2E tests are failing consistent on arm64, I haven't figured out the reasons. Maybe we can add them in separate PR to unblock the bug fix? @ningziwen @vsiravar
> > 
> > 
> > @azhouwd Did you run into this in local `make test-e2e`?
> 
> No the whole suite works fine locally, I tried several times

Does the test work if you execute it manually on the CI runners by logging in?

azhouwd:
> > > > The newly added E2E tests are failing consistent on arm64, I haven't figured out the reasons. Maybe we can add them in separate PR to unblock the bug fix? @ningziwen @vsiravar
> > > 
> > > 
> > > @azhouwd Did you run into this in local `make test-e2e`?
> > 
> > 
> > No the whole suite works fine locally, I tried several times
> 
> Does the test work if you execute it manually on the CI runners by logging in? What if we put a sleep after
> 
> https://github.com/runfinch/finch/blob/7b0f62b5162cd89e67c1f4c4e0505afa6737189f/e2e/vm/finch_config_file_test.go#L39
> 
> since the registry is running in detached mode and might take a while to come up? WDYT

yeah you are right, I printed the images and containers trying to debug, but the issue happens to be fixed..

ningziwen (approved):

ningziwen (dismissed):

vsiravar (approved):

azhouwd:
Merged due to the e2e failure is a known issue with Rosetta

PR #323, opened 2023-03-28
ci: add "exp" PR type

pendo324:
Issue #, if available:

*Description of changes:*
  - Add a new "expr" PR type to track experimental features. An "experimental" feature is something that can be used, but is not guaranteed to be stable (such as the changes made by #282).

*Testing done:*
  - N/A

#### License Acceptance


.github/workflows/release-please.yaml
```diff
@@ -38,6 +38,11 @@ jobs:
                 "type":"build",
                 "section":"Build System or External Dependencies",
                 "hidden":false
+              },
+              {
+                "type":"expr",
+                "section":"Experimental",
```
ningziwen, line 44: Or "Experimental Features" to make it clearer?

.github/workflows/release-please.yaml
```diff
@@ -38,6 +38,11 @@ jobs:
                 "type":"build",
                 "section":"Build System or External Dependencies",
                 "hidden":false
+              },
+              {
+                "type":"expr",
```
ningziwen, line 43: Is "expr" the most intuitive short version of "experimental"? How about "exp"? https://www.abbreviations.com/abbreviation/experimental

ningziwen (commented):
Not language expert.. just want to raise the discussion to think more about the naming and messages.

ningziwen:
Should the pr title be "ci"?

.github/workflows/release-please.yaml
pendo324: Sure, we can switch it to "exp" 👍

.github/workflows/release-please.yaml
pendo324: I think we can do something like "exp(feat)", since not all exp changes necessarily have to be features. WDYT?

pendo324 (commented):
Also wondering if we really need to run the full e2e tests for any `.github/**.yaml` changes. Should I also create a new ci workflow to only lint github workflow yaml, like how we do it in [ci-docs](https://github.com/runfinch/finch/blob/main/.github/workflows/ci-docs.yaml)?

ningziwen:
> Also wondering if we really need to run the full e2e tests for any `.github/**.yaml` changes. Should I also create a new ci workflow to only lint github workflow yaml, like how we do it in [ci-docs](https://github.com/runfinch/finch/blob/main/.github/workflows/ci-docs.yaml)?

@pendo324 The e2e test itself can be added/changed in **.yaml so it is a bit different with .doc. Even specifying a subset of yamls, it can be risky.

pendo324:
> > Also wondering if we really need to run the full e2e tests for any `.github/**.yaml` changes. Should I also create a new ci workflow to only lint github workflow yaml, like how we do it in [ci-docs](https://github.com/runfinch/finch/blob/main/.github/workflows/ci-docs.yaml)?
> 
> @pendo324 The e2e test itself can be added/changed in **.yaml so it is a bit different with .doc. Even specifying a subset of yamls, it can be risky.

I added `ci-github-yaml.yaml` and it should only trigger for `.github/**.yaml` files. I agree that e.g. finch.yaml changes should run full e2e tests, but I'm not really sure if GitHub workflows/actions need to run the whole suite. I can remove it if you think it should be its own issue/PR though. WDYT?

pendo324:
> @pendo324 This will cause changing [ci-github-yaml.yaml](https://github.com/runfinch/finch/pull/323/files#diff-075b0082f413b37c55c96ca1cc98bc2be8d58861177639a2540f1e788f7acf65) itself will not run e2e tests, which is risky and require more manual effort to review the changes.
> 
> Let's discuss in a separated issue/PR if you still want to propose this.

I don't necessarily think that's risky, but I see how changing the "actual" e2e test (`ci.yaml`) could be risky. I'll remove it and create a new issue to track this because I do think running all of the e2e tests for a simple yaml change is wasteful.

ningziwen (approved):

pendo324:
Skipping CI / e2e-tests since this change doesn't impact functionality (just a GitHub workflow YAML change).

PR #351, opened 2023-04-10
build: add local-core Makefile target

pendo324:
Issue #, if available:

*Description of changes:*
- adds new Makefile target that uses the locally compiled submodule version of Lima instead of the version from the lima-and-qemu remote archive. This allows more easy testing of local submodule changes before building a new lima-and-qemu archive.

*Testing done:*
- local testing

#### License Acceptance

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -56,6 +56,9 @@ arch-test:
 .PHONY: all
 all: arch-test finch finch-core finch.yaml networks.yaml config.yaml lima-and-qemu
 
+.PHONY: all-local
+all-local: arch-test finch networks.yaml config.yaml lima-and-qemu finch-core finch.yaml
+
 .PHONY: finch-core
 finch-core:
 	cd deps/finch-core && \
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -56,6 +56,9 @@ arch-test:
 .PHONY: all
 all: arch-test finch finch-core finch.yaml networks.yaml config.yaml lima-and-qemu
 
+.PHONY: all-local
+all-local: arch-test finch networks.yaml config.yaml lima-and-qemu finch-core finch.yaml
```
ningziwen, line 60: Do you want to replace to local-core here?

pendo324: Yeah, typo. Good catch. I changed the names right before I made the PR

pendo324: Fixed in latest rev

ningziwen:
This can unblock local tests but the CI still uses `make` unless we change it to `make all-local`.

pendo324:
> This can unblock local tests but the CI still uses `make` unless we change it to `make all-local`.

My motivation was to make it possible to test locally. CI should use what is actually used by the release imo. Maybe we can add more ci targets in the future and have some use the all-local target and some use the lima-and-qemu target.

ningziwen (approved):

pendo324:
Compose test failed due to known issue. Merging.

Issue #345, opened 2023-04-05
Add benchmarking in Finch


ningziwen:
**What is the problem you're trying to solve?.**
Report the key metric benchmarking in CI and persistent diagram.

**Describe the feature you'd like**
`go test -bench .` could be the interface to start from.
[Continuous benchmark Github action](https://github.com/marketplace/actions/continuous-benchmark) may be integrated to report in CI and persistent diagram.

**Additional context**

ningziwen:
[PR](https://github.com/runfinch/finch/pull/344) to add barebone benchmarking for `vm init` and `vm start` only.

estesp:
Should this issue be closed now @ningziwen ?

ningziwen:
Closing the issue as the basic benchmarking result is already in Finch. https://runfinch.github.io/finch/dev/bench/

Code: https://github.com/runfinch/finch/tree/gh-pages/dev/bench

Further benchmarking enhancements will be tracked in separated issues.

PR #373, opened 2023-04-20
test: add cpu footprint metrics and disk usage to benchmarking

ningziwen:

*Description of changes:*

*Testing done:*
Tested locally with benchmark make targets

#### License Acceptance


benchmark/benchmark.go
```diff
@@ -30,3 +36,101 @@ func GetSubject() (string, error) {
...
+
+// Wrapper reports the benchmarking metrics of targetFunc to testing.B.
+func Wrapper(b *testing.B, targetFunc func(), cleanupFunc func()) {
+	var peakCPUSum, avgCPUSum float64
+	var totalCPUTimeSum time.Duration
+	var diskUsageDeltaSum int64
+	b.ResetTimer()
+	for n := 0; n < b.N; n++ {
+		b.StartTimer()
+		peakCPU, avgCPU, totalCPUTime, diskUsageDelta, err := measureMetrics(targetFunc)
```
monirul, line 48: Should we create a struct and return the struct instead of multiple variables?
eg.
```
mesuredMatrics, err := easureMetrics(targetFunc);
```

benchmark/benchmark.go
```diff
@@ -30,3 +36,101 @@ func GetSubject() (string, error) {
...
+		if err != nil {
+			b.Fatal("Error measuring metrics:", err)
+			return
+		}
+		b.StopTimer()
+		peakCPUSum += peakCPU
```
monirul, line 54: If we use the struct, we can move this to a separate method in the struct?
e.g.
```
var totalSumMetrics MeasuredMetrics;
totalSumMetrices.Add(mesuredMatrics)
```

pendo324:
Haven't looked at the changes in detail yet, but how confident are we that only one GitHub workflow is running on a host at a time? I could see getting incorrect data if multiple jobs were running simultaneously.

monirul (changes requested):

ningziwen:
@pendo324 Github runner only runs one job at one time. https://github.com/orgs/community/discussions/26769#discussioncomment-3253321

And our other tests also have this assumption. For example, multiple e2e tests may choose same Finch config. Or multiple installer test workflow may want to install and uninstall at the same time.

monirul:
> Haven't looked at the changes in detail yet, but how confident are we that only one GitHub workflow is running on a host at a time? I could see getting incorrect data if multiple jobs were running simultaneously.

I have the same concern, how do we ensure the environment is always same during the execution. 
Also, are we running benchmarking for different version of OS to find out if there is any impact based on OS?

ningziwen:
> > Haven't looked at the changes in detail yet, but how confident are we that only one GitHub workflow is running on a host at a time? I could see getting incorrect data if multiple jobs were running simultaneously.
> 
> I have the same concern, how do we ensure the environment is always same during multiple execution. Also, are we running benchmarking for different version of OS to find out if there is any impact based on OS?

@monirul Yes we should report the metrics by different OS and arch. The label can be like "<os-version>-<arch>-VMInit". It should be addressed in a separated PR to onboard the benchmarking Github action.

pendo324 (approved):

pendo324 (approved):
Looks a little cleaner using the struct

benchmark/benchmark.go
```diff
@@ -17,6 +23,24 @@ const InstalledTestSubject = "finch"
...
+// Metrics includes all the additional metrics that Finch benchmark reports,
+// except testing.B natively supported metrics like execution time and memory.
+type Metrics struct {
+	PeakCPUUsage    float64
+	AverageCPUUsage float64
+	TotalCPUTime    time.Duration
+	DiskUsageDelta  int64
+}
+
+func addMetrics(m1 Metrics, m2 Metrics) Metrics {
```
monirul, line 35: should we change it to method of Metrics struct?
e.g.
```
func (m1 *Metrics) addMetrics(m2 Metrics) {
	m1.PeakCPUUsage += m2.PeakCPUUsage,
	m1.AverageCPUUsage += m2.AverageCPUUsage,
	m1.TotalCPUTime += m2.TotalCPUTime,
	m1.DiskUsageDelta += m2.DiskUsageDelta,
}
```

benchmark/benchmark.go
```diff
@@ -30,3 +54,101 @@ func GetSubject() (string, error) {
...
+	b.ResetTimer()
+	for n := 0; n < b.N; n++ {
+		b.StartTimer()
+		metrics, err := measureMetrics(targetFunc)
+		if err != nil {
+			b.Fatal("Error measuring metrics:", err)
+			return
+		}
+		b.StopTimer()
+		metricsSum = addMetrics(metricsSum, metrics)
```
monirul, line 70: If you change the function as `func (m1 *Metrics) addMetrics(m2 Metrics) {`
you can simply write as `metricsSum.Add(metrics)`

benchmark/benchmark.go
```diff
@@ -30,3 +54,101 @@ func GetSubject() (string, error) {
...
+					ticker.Stop()
+					wg.Done()
+					return
+				}
+				cpuUsage = append(cpuUsage, percent[0])
+			case <-done:
+
+				after, err := disk.Usage("/")
+				if err != nil {
+					fmt.Printf("Error getting disk usage after: %v\n", err)
```
monirul, line 119: Instead of fmt.printf(), should we use logrus? that way we can print errors as Warning/Error. which would be easy to find?

monirul (changes requested):
Added few more comments.

benchmark/benchmark.go
```diff
@@ -30,3 +56,101 @@ func GetSubject() (string, error) {
...
+		metricsSum.Add(metrics)
+		cleanupFunc()
+	}
+	b.ReportMetric(metricsSum.PeakCPUUsage/float64(b.N), "%cpu_peak/op")
+	b.ReportMetric(metricsSum.AverageCPUUsage/float64(b.N), "%cpu_avg/op")
+	b.ReportMetric(metricsSum.TotalCPUTime.Seconds()/float64(b.N), "cpu_seconds/op")
+	b.ReportMetric(float64(metricsSum.DiskUsageDelta/int64(b.N)), "disk_bytes/op")
+}
+
+func measureMetrics(f func()) (Metrics, error) { //nolint:unparam // make it extensible for future error handling
```
monirul, line 81: nit: What is the responsibility of this `f()`? Can we give it a better name from which tells what is it for?

benchmark/benchmark.go
```diff
@@ -30,3 +56,101 @@ func GetSubject() (string, error) {
...
+	done := make(chan struct{})
+	var cpuUsage []float64
+	var startTime time.Time
+	var diskUsageBefore, diskUsageAfter uint64
+
+	wg := sync.WaitGroup{}
+	wg.Add(1)
```
monirul, line 88: nit: The intension of the wg.Add(1) is not clear here. Can you pls add a comment for this line at least?

monirul (changes requested):

monirul (approved):
Nice work Ziwen. Changes looks good to me!
Thanks.

PR #376, opened 2023-04-24
ci: Updated runner labels, removed minor os version

weikequ:
Issue #, if available:

*Description of changes:*
- Updated runner labels to be split into 2 categories: "release" and "test"
- Updated runner labels to only reference the major version of macOS

*Testing done:*

#### License Acceptance

### DIFF
--- a/.github/workflows/ci-docs.yaml
+++ b/.github/workflows/ci-docs.yaml
@@ -46,7 +46,13 @@ jobs:
   e2e-tests:
     strategy:
       matrix:
-        os: [[self-hosted, macos, amd64, 13], [self-hosted, macos, amd64, 12.6], [self-hosted, macos, arm64, 13], [self-hosted, macos, arm64, 12.6]]
+        os: 
+          [
+            [self-hosted, macos, amd64, 13, test], 
+            [self-hosted, macos, amd64, 12, test], 
+            [self-hosted, macos, arm64, 13, test], 
+            [self-hosted, macos, arm64, 12, test]
+          ]
     runs-on: ${{ matrix.os }} 
     steps:
       - run: echo "Skipping CI for docs & contrib files"
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -82,7 +82,13 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [[self-hosted, macos, amd64, 13], [self-hosted, macos, amd64, 12.6], [self-hosted, macos, arm64, 13], [self-hosted, macos, arm64, 12.6]]
+        os: 
+          [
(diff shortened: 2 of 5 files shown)


.github/workflows/ci-docs.yaml
```diff
@@ -46,7 +46,13 @@ jobs:
   e2e-tests:
     strategy:
       matrix:
-        os: [[self-hosted, macos, amd64, 13], [self-hosted, macos, amd64, 12.6], [self-hosted, macos, arm64, 13], [self-hosted, macos, arm64, 12.6]]
+        os: 
+          [
+            [self-hosted, macos, amd64, 13, test],
```
KevinLiAWS, line 51: Have the current runners already been tagged with "test"? Will the e2e test run for this PR?

weikequ: Not yet, as we don't know when this PR will be merged (don't want the runners to be unavailable), I will change the labels on the runners right before I merge it. Note that when this PR is merged, any current PRs will also need to update their branch to have the right labels/run CI.

ningziwen (approved):

KevinLiAWS (approved):

.github/workflows/ci-docs.yaml
```diff
@@ -46,7 +46,13 @@ jobs:
   e2e-tests:
     strategy:
       matrix:
-        os: [[self-hosted, macos, amd64, 13], [self-hosted, macos, amd64, 12.6], [self-hosted, macos, arm64, 13], [self-hosted, macos, arm64, 12.6]]
+        os: 
+          [
+            [self-hosted, macos, amd64, 13, test], 
+            [self-hosted, macos, amd64, 12, test],
```
monirul, line 52: What is the implication of the value of `test`/`release`? Is it possible to add one line of comment explaining what is the implication of these values?

monirul (changes requested):

.github/workflows/ci-docs.yaml
weikequ: The implication is that workflows that are not part of the release process should be labelled 'release'. This way, we can separate our testing and release infrastructure to increase security. Do you think we should add a comment to every file where this occurs? Building on your idea a bit, it might be better instead to have a central markdown file to explain all the labels.

monirul: Central markdown file is fine. As long as it is easy to find. 😄

monirul (dismissed):

weikequ:
Will merge without fully completing e2e tests as this change relates to them.

KevinLiAWS (approved):
LGTM

vsiravar (approved):
LGTM

PR #384, opened 2023-04-28
ci: add benchmarking workflow

ningziwen:

*Description of changes:*

*Testing done:*
https://runfinch.github.io/finch/dev/bench/macOS/12/arm64/
https://runfinch.github.io/finch/dev/bench/macOS/12/x86_64/
https://runfinch.github.io/finch/dev/bench/macOS/13/arm64/
https://runfinch.github.io/finch/dev/bench/macOS/13/x86_64/

#### License Acceptance

### DIFF
--- a/.github/workflows/benchmark.yaml
+++ b/.github/workflows/benchmark.yaml
@@ -1,0 +1,70 @@
+# - When a third-party action is added (i.e., `uses`), please also add it to `download-licenses` in Makefile.
+# - When a job is added/removed/renamed, please make corresponding changes in ci-docs.yaml.
+name: Benchmark
+on:
+  push:
+    branches:
+      - main
+    paths-ignore:
+      - '**.md'
+      - 'contrib/**'
+
+permissions:
+  # deployments permission to deploy GitHub pages website
+  deployments: write
+  # contents permission to update benchmark contents in gh-pages branch
+  contents: write
+
+jobs:
+  benchmark:
+    strategy:
+      fail-fast: false
+      matrix:
+        os:
+          [
+            [self-hosted, macos, amd64, 13, test],
+            [self-hosted, macos, amd64, 12, test],
+            [self-hosted, macos, arm64, 13, test],
+            [self-hosted, macos, arm64, 12, test]
+          ]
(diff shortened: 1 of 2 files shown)


.github/workflows/benchmark.yaml
```diff
@@ -0,0 +1,71 @@
+# - When a third-party action is added (i.e., `uses`), please also add it to `download-licenses` in Makefile.
+# - When a job is added/removed/renamed, please make corresponding changes in ci-docs.yaml.
```
weikequ, line 2: Will this be a required workflow as part of branch protections? If so, we need to create a docs version if we're ignoring the docs and contrib folder.

.github/workflows/benchmark.yaml
```diff
@@ -0,0 +1,71 @@
...
+          if pgrep '^socket_vmnet'; then
+            sudo pkill '^socket_vmnet'
+          fi
+      - name: Install Rosetta 2
+        run: echo "A" | softwareupdate --install-rosetta || true
+      - run: brew install go lz4 automake autoconf libtool
+      - name: Build project
+        run: |
+          export PATH="/opt/homebrew/opt/libtool/libexec/gnubin:$PATH"
+          which libtool
```
weikequ, line 55: nit: probably don't need this line

pendo324: +1, I think this is copy/pasted from other actions, we can probably remove it in general

.github/workflows/benchmark.yaml
ningziwen: I think for now workflow still needs to run the full CI.

weikequ: Ok, so no branch protection requirement to run this workflow, right (aka PRs don't need this workflow to pass in order to be merged)?

pendo324 (approved):

.github/workflows/benchmark.yaml
weikequ: Got it. That sounds good! Thanks for confirming

weikequ (approved):

PR #393, opened 2023-05-05
refactor: minor optimization to not fetch qemu packages if they are already installed

pendo324:
Issue #, if available: N/A

*Description of changes:*
  - Check if executables are already present before trying to install packages. Saves a potentially useless and costly dnf cache refresh.

*Testing done:*
  - local testing

#### License Acceptance

### DIFF
--- a/pkg/config/lima_config_applier.go
+++ b/pkg/config/lima_config_applier.go
@@ -154,7 +154,18 @@ func toggleUserModeEmulationInstallationScript(limaCfg *limayaml.LimaYAML, enabl
 			Mode: "system",
 			Script: fmt.Sprintf(`%s
 #!/bin/bash
-dnf install -y --setopt=install_weak_deps=False qemu-user-static-aarch64 qemu-user-static-arm qemu-user-static-x86
+qemu_pkgs=""
+if [ ! -f /usr/bin/qemu-aarch64-static ]; then
+  qemu_pkgs="$qemu_pkgs qemu-user-static-aarch64"
+elif [ ! -f /usr/bin/qemu-aarch64-static ]; then
+  qemu_pkgs="$qemu_pkgs qemu-user-static-arm"
+elif [ ! -f  /usr/bin/qemu-aarch64-static ]; then
+  qemu_pkgs="$qemu_pkgs qemu-user-static-x86"
+fi
+
+if [[ $qemu_pkgs ]]; then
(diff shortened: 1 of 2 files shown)


pkg/config/lima_config_applier.go
```diff
@@ -154,7 +154,18 @@ func toggleUserModeEmulationInstallationScript(limaCfg *limayaml.LimaYAML, enabl
...
+if [ ! -f /usr/bin/qemu-aarch64-static ]; then
+  qemu_pkgs="$qemu_pkgs qemu-user-static-aarch64"
+elif [ ! -f /usr/bin/qemu-aarch64-static ]; then
+  qemu_pkgs="$qemu_pkgs qemu-user-static-arm"
+elif [ ! -f  /usr/bin/qemu-aarch64-static ]; then
+  qemu_pkgs="$qemu_pkgs qemu-user-static-x86"
+fi
+
+if [[ $qemu_pkgs ]]; then
+  dnf install -y --setopt=install_weak_deps=False ${qemu_pkgs}
```
sam-berning, line 167: nit: I think the current solution should be 100% fine for this current case, but I think it'd be a little safer to have `qemu_pkgs` be an array, e.g.

```
qemu_pkgs=()
qemu_pkgs=("${qemu_pkgs[@]}" "qemu-user-static-aarch64")
dnf install -y --setopt=install_weak_deps=False "${qemu_pkgs[@]}"
```

just because it allows us to quote the variable expansion.

sam-berning (approved):

pkg/config/lima_config_applier.go
```diff
@@ -154,7 +154,18 @@ func toggleUserModeEmulationInstallationScript(limaCfg *limayaml.LimaYAML, enabl
 			Mode: "system",
 			Script: fmt.Sprintf(`%s
 #!/bin/bash
-dnf install -y --setopt=install_weak_deps=False qemu-user-static-aarch64 qemu-user-static-arm qemu-user-static-x86
+qemu_pkgs=""
```
monirul, line 157: we have repeat this section multiple times in other files. Is it possible to create as bash function. something like
https://unix.stackexchange.com/questions/599604/create-reusable-functions-bash
Btw, i am not sure whether this is even doable in your case.

monirul (approved):

pendo324:
Thanks for the comments. Will see if I can incorporate these recommendations in a future PR.

PR #395, opened 2023-05-09
ci: explicitly set shell when running commands

pendo324:
Issue #, if available:

*Description of changes:*
 - I noticed that the CI failures that we've been seeing all happen on M1 macOS and are related to brew not being in the path. This might be happening because brew is installed/configured in a non-default path, and the PATH setup that does happen as part of brew's install all happens in ZSH specific files (`~/.zprofile`/`~/.zshrc`), and the scripts are all being executed with bash if you check the failed CI run logs

*Testing done:*
 - local testing
   - this is a little hard to test locally, so we're just going to have to see if the runs work on this PR

#### License Acceptance

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -111,11 +111,14 @@ jobs:
       - name: Install Rosetta 2
         run: echo "A" | softwareupdate --install-rosetta || true 
       - run: brew install go lz4 automake autoconf libtool 
+        shell: zsh
       - name: Build project
         run: |
           export PATH="/opt/homebrew/opt/libtool/libexec/gnubin:$PATH"
           make
+        shell: zsh
       - run: make test-e2e
+        shell: zsh
   mdlint:
     runs-on: ubuntu-latest
     steps:


weikequ:
Seems like there's still some errors. Could be related to the fact that when I do `sudo su - ec2-user -c "which brew"`, it doesn't return the correct path. 

```shell
ip-172-31-62-65:~ root# su - ec2-user -c "which brew"
brew not found
ip-172-31-62-65:~ root# su - ec2-user -c 'echo $PATH'
/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin
```

.github/workflows/ci.yaml
```diff
@@ -111,11 +111,14 @@ jobs:
       - name: Install Rosetta 2
         run: echo "A" | softwareupdate --install-rosetta || true 
       - run: brew install go lz4 automake autoconf libtool 
+        shell: zsh {0}
```
monirul, line 114: What is the significance of adding this line?

pendo324:
> Seems like there's still some errors on m1 macs. Could be related to the fact that when I do `sudo su - ec2-user -c "which brew"`, it doesn't return the correct path.
> 
> ```shell
> ip-172-31-62-65:~ root# su - ec2-user -c "which brew"
> brew not found
> ip-172-31-62-65:~ root# su - ec2-user -c 'echo $PATH'
> /usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin
> ```

There are errors because I didn't manually fix all of the macOS runners. I can do that now, but I wanted to fix it in the infrastructure package

.github/workflows/ci.yaml
pendo324: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepsshell

Since macOS 10.15.x, the default shell on macOS has been `zsh`. The default shell for GitHub workflows is `bash`. As I noted in the PR description, Homebrew install configures ZSH specific environment because ZSH is the default shell. Since ARM macOS runners do not have `brew` installed to a "default" path (e.g. `/usr/bin`) and instead have it installed to `/opt/homebrew`, if running bash instead of a configured ZSH, `brew` will not be found.

weikequ (approved):
LGTM

PR #403, opened 2023-05-16
ci: Fix brew/aws command and Rosetta tests

KevinLiAWS:
Issue #, if available:

*Description of changes:*
Fixed the missing configuration step for brew/aws command and Rosetta tests in release related github action.

*Testing done:*
Have run the action on this branch

#### License Acceptance

### DIFF
--- a/.github/workflows/release-homebrew.yaml
+++ b/.github/workflows/release-homebrew.yaml
@@ -74,6 +74,7 @@ jobs:
           fi
           cd $(brew --repo homebrew/cask) 
           brew uninstall --cask ./Casks/finch.rb || true
+        shell: zsh {0}
       - name: Set up Homebrew
         id: set-up-homebrew
         uses: Homebrew/actions/setup-homebrew@master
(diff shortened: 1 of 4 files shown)


ningziwen:
There is a previous PR about this. https://github.com/runfinch/finch/pull/395

It seems cumbersome to add `zsh {0}` everywhere. Is it possible to load .zshrc during runner setup? @weikequ @pendo324

pendo324 (approved):

pendo324:
> There is a previous PR about this. #395
> 
> It seems cumbersome to add `zsh {0}` everywhere. Is it possible to load .zshrc during runner setup? @weikequ @pendo324

`~/.zshenv` _is_ loaded and sets the brew path already, the problem is that we aren't executing ZSH by default, we're using bash.

I'm pretty sure this previously worked because `~/.bashrc` / `bash_profile` / some-other-file _also_ set the homebrew path, but now it's only set for ZSH.

Btw, `~/.zshrc` won't be loaded by default by non-interactive shells, that's why in my previous commit I added the configuration to `~/.zshenv`.

PR #424, opened 2023-05-27
ci: Add release automation action

KevinLiAWS:
*Description of changes:*
This PR added a new passive action subscribe to main branch new tag creation.
New tag creation means Finch has a new release, the new workflow will trigger build pkg, upload pkg and homebrew release on the latest tag.

*Testing done:*
This is a new action and has cross branch/tag involved, needs some experiment  on main branch after check in.

#### License Acceptance

### DIFF
--- a/.github/workflows/release-automation.yaml
+++ b/.github/workflows/release-automation.yaml
@@ -24,10 +24,10 @@ jobs:
         with:
           GITHUB_REF_NAME: ${{ steps.latest-tag.outputs.tag }}
 
-      - name: Upload .pkg installer to the latest release
-        uses: ./.github/workflows/upload-installer-to-release.yaml
-        with:
-          GITHUB_REF_NAME: ${{ steps.latest-tag.outputs.tag }}
+      #- name: Upload .pkg installer to the latest release
+      #  uses: ./.github/workflows/upload-installer-to-release.yaml
+      #  with:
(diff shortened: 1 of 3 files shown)


.github/workflows/release-automation.yaml
```diff
@@ -0,0 +1,33 @@
+name: Release Finch latest version
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      - main
```
vsiravar, line 6: Why do we need a release for every push on the main branch?

KevinLiAWS: I think this is for every new tag on main branch

.github/workflows/release-automation.yaml
```diff
@@ -0,0 +1,31 @@
+name: Release Finch latest version
+on:
+  workflow_dispatch:
+  create:
+    tags:
+      - 'v*'
```
vsiravar, line 6: This might run on all branches when a new tag is created, it's okay for now. Maybe adding a TODO to indicate this would help.

vsiravar (approved):

.github/workflows/release-automation.yaml
KevinLiAWS: Will test and update comments

PR #444, opened 2023-06-03
fix: Add cleanup script to Makefile

KevinLiAWS:
*Description of changes:*
Remove files under other directories besides _output, and kill running processes for network, lima and qemu.

*Testing done:*
Tested locally

#### License Acceptance

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -317,6 +317,12 @@ mdlint-ctr:
 
 .PHONY: clean
 clean:
+	-sudo pkill '^socket_vmnet'
+	-sudo pkill '^qemu-system-'
+	-sudo pkill '^limactl'
+	-sudo rm -rf "/opt/finch/"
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -317,6 +317,12 @@ mdlint-ctr:
 
 .PHONY: clean
 clean:
+	-sudo pkill '^socket_vmnet'
```
vsiravar, line 320: This might abruptly stop finch installed via installer and create a bad state for the version installed by brew/installer. It might be safer to kill the `qemu `and `limactl` pinned in this repo. Something like ` pkill <Install directory>/finch/_output/lima/bin/qemu-system-x86_64`

KevinLiAWS: It will be anyway impacted since the shared folder is removed i.e., /opt/finch/.
The user anyway need to remove the vm and re-init it in this case.

vsiravar (approved):

Issue #209, opened 2023-02-06
Finch cannot resolve special IP `host-gateway`


ningziwen:
**Describe the bug**
Finch cannot resolve special IP `host-gateway` in `finch run --add-host name:host-gateway IMAGE`

**Steps to reproduce**
```
finch run --add-host name:host-gateway -it public.ecr.aws/amazonlinux/amazonlinux:2 bash
# cat /etc/hosts
...
host-gateway    name
...
```

**Expected behavior**
```
finch run --add-host name:host-gateway -it public.ecr.aws/amazonlinux/amazonlinux:2 bash
# cat /etc/hosts
...
<resolved-ip>    name
...
```

**Screenshots or logs**

**Additional context**
In Docker,
```
docker run --add-host name:host-gateway -it public.ecr.aws/amazonlinux/amazonlinux:2 bash
# cat /etc/hosts
...
192.168.65.2	name
...
```

Can reproduce in nerdctl in Finch VM:
```
nerdctl run --add-host name:host-gateway -it public.ecr.aws/amazonlinux/amazonlinux:2 bash
# cat/etc/hosts
...
host-gateway    name
...
```

ningziwen:
In Finch, to make the feature parity with Docker, the ideal user stories will be:

As a Finch user, if I use `host-gateway` in my `finch run` with configuring `host-gateway-ip` in Finch config, when I try to access host with the alias in container, it will use the ip I configured in Finch config.

```
cat ~/.finch/finch.yaml
...
host-gateway-ip: "192.168.65.2"
...
```

```
finch run --add-host name:host-gateway -it public.ecr.aws/amazonlinux/amazonlinux:2 bash
# cat /etc/hosts
...
192.168.65.2    name
...

## curl name:<port> with a valid port will succeed if 192.168.65.2 is a valid host gateway ip
```

As a Finch user, if I use `host-gateway` in my `finch run` without configuring `host-gateway-ip` in Finch config, my container can still use the IP alias to access host.

```
finch run --add-host name:host-gateway -it public.ecr.aws/amazonlinux/amazonlinux:2 bash
# cat /etc/hosts
...
<default-ip>    name
...
## curl name:<port> with a valid port will succeed
```

ningziwen:
The solution could be split to multiple stages:

At the first stage, we can resolve `host-gateway` to a static IP without being configurable.

[...]

PR #453, opened 2023-06-14
feat: resolve special IP host-gateway

TrungBui59:
Issue #209 

*Description of changes:*

I add the host-gateway-ip to the finch config file and use that to make the host-gateway-ip configurable using nerdctl

*Testing done:*
I have done some unit test, it passed most of the e2e tests, but now it show some weird error that I don't think is a problem because of my code change

Screenshot 2023-06-14 at 1 46 36 PM

#### License Acceptance


ningziwen:
@TrungBui59 Thanks. Looks like your screenshot only covers the final summary of the failing e2e tests. Could you scroll up, find the verbose logs for each failing scenario, and paste here (Please paste instead of screenshotting for convenient search)?

cmd/finch/nerdctl.go
```diff
@@ -121,11 +124,13 @@ func (nc *nerdctlCommand) run(cmdName string, args []string) error {
...
 			switch arg {
 			case "--add-host":
-				args[i+1] = resolveIP(args[i+1], nc.logger)
+				args[i+1] = resolveIP(nc.fc, args[i+1], nc.logger)
 			default:
-				resolvedIP := resolveIP(arg[11:], nc.logger)
+				resolvedIP := resolveIP(nc.fc, arg[11:], nc.logger)
 				arg = fmt.Sprintf("%s%s", arg[0:11], resolvedIP)
 			}
+			nerdctlArgs = append(nerdctlArgs, "--host-gateway-ip")
```
ningziwen, line 132: Why do we need to append this?

TrungBui59: This is for adding the host-gateway-ip into the nerdctl, other option might be adding it during the init

ningziwen: The logic around here is to do special processing for specific args. If --host-gateway-ip can be directly passed to nerdctl and doesn't require special processing, I guess you don't need to change anything.

TrungBui59:
> @TrungBui59 Thanks. Looks like your screenshot only covers the final summary of the failing e2e tests. Could you scroll up, find the verbose logs for each failing scenario, and paste here (Please paste instead of screenshotting for convenient search)?

Here are the log, for the errors that is causing by my changes, it is mostly timeout issue
Screenshot 2023-06-14 at 2 34 29 PM
Screenshot 2023-06-14 at 2 34 47 PM
Screenshot 2023-06-14 at 2 35 56 PM
Screenshot 2023-06-14 at 2 36 46 PM

cmd/finch/nerdctl.go
TrungBui59: @ningziwen Initially I plan to change the resolveIP function in the nerdctl.go file:
`
func resolveIP(fc *config.Finch, host string, logger flog.Logger) string {
	parts := strings.SplitN(host, ":", 2)
	// If the IP Address is a string called "host-gateway", replace this value with the IP address that can be used to
	// access host from the containers.
	// TODO: make the host gateway ip configurable.
	if parts[1] == dockerops.HostGatewayName {
		resolveIP := *fc.HostGatewayIP
		logger.Debugf(`Resolving special IP "host-gateway" to %q for host %q`, resolveIP, parts[0])
		return fmt.Sprintf("%s%s", parts[0], resolveIP)
	}
	return host
}
`

Since I already change the config file to add that option in the Finch struct

ningziwen:
There are 3 ways of filling host gateway IP in nerdctl: config, command arg and env var.

Finch should also support all of them for feature parity.

By default users can already use command arg to fill IP in Finch, as Finch inherits commands from Nerdctl. Env var should also work as Finch passes the envs to nerdctl commands.

Only Finch config needs some additional work. If we fill the value from Finch config to nerdctl command arg (as the current revision of the PR is doing) or env var, they may have conflict with the user-input command arg and env var. Finch can set some rules to let one override another, but that would cause unnecessary business logic in Finch. We could have directly passed the value from Finch config to nerdctl config, so nerdctl will decide whatever rules to let one override another.

TrungBui59:
> There are 3 ways of filling host gateway IP in nerdctl: config, command arg and env var.
> 
> 
> 
> Finch should also support all of them for feature parity.
> 
> 
> 
> By default users can already use command arg to fill IP in Finch, as Finch inherits commands from Nerdctl. Env var should also work as Finch passes the envs to nerdctl commands.
> 
> 
> 
> Only Finch config needs some additional work. If we fill the value from Finch config to nerdctl command arg (as the current revision of the PR is doing) or env var, they may have conflict with the user-input command arg and env var. Finch can set some rules to let one override another, but that would cause unnecessary business logic in Finch. We could have directly passed the value from Finch config to nerdctl config, so nerdctl will decide whatever rules to let one override another.
> 
> 

Oh I see, I will working on that approach right away

pkg/config/defaults.go
```diff
@@ -46,5 +47,9 @@ func applyDefaults(cfg *Finch, deps LoadSystemDeps, mem fmemory.Memory) *Finch {
 		cfg.Rosetta = pointer.Bool(false)
 	}
 
+	if cfg.HostGatewayIP == nil {
```
ningziwen, line 50: Maybe we don't have to fill host gateway IP for users by default. We can keep it nil and process it based on nil check. That behaviour is docker compliance.

Finch also has an existing field "additional_directories" that isn't filled by default.

cmd/finch/nerdctl_test.go
```diff
@@ -369,11 +369,10 @@ func TestNerdctlCommand_run(t *testing.T) {
 				getVMStatusC.EXPECT().Output().Return([]byte("Running"), nil)
 				logger.EXPECT().Debugf("Status of virtual machine: %s", "Running")
 				ncsd.EXPECT().LookupEnv("FINCH_DOCKER_COMPAT").Return("", false)
-				logger.EXPECT().Debugf(`Resolving special IP "host-gateway" to %q for host %q`, "192.168.5.2", "name")
 				ncsd.EXPECT().LookupEnv("COSIGN_PASSWORD").Return("", false)
 				c := mocks.NewCommand(ctrl)
 				lcc.EXPECT().Create("shell", limaInstanceName, "sudo", "-E", nerdctlCmdName, "run",
-					"--rm", "--add-host", "name:192.168.5.2", "alpine:latest").Return(c)
+					"--rm", "--add-host", "name:host-gateway", "alpine:latest").Return(c)
```
ningziwen, line 375: Do we still need these test cases as unit tests if the tested logic is not in the file anymore?

pkg/config/nerdctl_config_applier_test.go
```diff
@@ -231,10 +238,15 @@ func TestNerdctlConfigApplier_Apply(t *testing.T) {
...
+				Memory:  pointer.String("2GiB"),
+				CPUs:    pointer.Int(4),
+				VMType:  pointer.String("qemu"),
+				Rosetta: pointer.Bool(false),
+			}
 			d := mocks.NewDialer(ctrl)
-
 			tc.mockSvc(t, fs, d)
-			got := NewNerdctlApplier(d, fs, tc.path, tc.hostUser).Apply(tc.remoteAddr)
+			got := NewNerdctlApplier(fc, d, fs, tc.path, tc.hostUser).Apply(tc.remoteAddr)
```
ningziwen, line 250: Need a test case that includes host-gateway-ip in finch config. You can run `make coverage` to self-check if any lines are not covered.

ningziwen:
Could you add e2e tests? 
The host-gateway-ip in finch config is under e2e/vm? Similar to https://github.com/runfinch/finch/blob/main/e2e/vm/additional_disk_test.go

The host-gateway-ip arg and env var can be added in common-tests repo around here. (can be a separated PR as it is separated repo) https://github.com/runfinch/common-tests/blob/6c7275007bf34fb6ddecc4013c16f1d79ff6d1d0/tests/run.go#L335

pkg/config/nerdctl_config_applier_test.go
TrungBui59: Oh ok I see

cmd/finch/nerdctl_test.go
TrungBui59: will change it to a more meaningful test

ningziwen:
Please rename the PR name to conventional commit message style. https://www.conventionalcommits.org/en/v1.0.0/

TrungBui59:
@ningziwen Sorry for not being responsive, I was busy with my intern work. I still dont know how to make it change the host-gateway-ip when the VM starts/restarts. I can make it change when we init the VM. Any suggestion on how to make it work with start and restart? so far, I have tried reading the finch config during the postVMStartAction, but it still does not work as I planned

TrungBui59:
@ningziwen Sorry for not being really responsive, I was busy with my intern work. I still dont know how to make it change the host-gateway-ip when the VM starts/restarts. I can make it change when we init the VM. Any suggestion on how to make it work with start and restart? so far, I have tried reading the finch config during the postVMStartAction, but it still does not work as I planned

ningziwen:
@TrungBui59 Reading the code, it seems passing it [here](https://github.com/runfinch/finch/blob/fef6e772c28429ef14068702c1dfd8e9fec2dea1/cmd/finch/virtual_machine.go#L93) should work. Have you tried it? What is the error?

TrungBui59:
> @TrungBui59 Reading the code, it seems passing it [here](https://github.com/runfinch/finch/blob/fef6e772c28429ef14068702c1dfd8e9fec2dea1/cmd/finch/virtual_machine.go#L93) should work. Have you tried it? What is the error?

@ningziwen The current approach is that I will pass the already existing finch config to the nerdctl, but when we restart it, the config change. And I still haven't found a way to pass the new value to it. I have tried to read the new config and create a new NerdctlConfigApplier with a new config, but it doesn't work

pkg/config/config.go
```diff
@@ -67,6 +68,7 @@ type Nerdctl struct {
 	CgroupManager    string   `toml:"cgroup_manager,omitempty"`
 	InsecureRegistry bool     `toml:"insecure_registry,omitempty"`
 	HostsDir         []string `toml:"hosts_dir,omitempty"`
+	HostGatewayIP *string `toml:"host_gateway_ip,omitempty"`
```
pendo324, line 71: nit: formatting

pkg/config/config.go
```diff
@@ -50,6 +50,7 @@ type Finch struct {
 	// Has no effect on systems where Rosetta 2 is not available (Intel/AMD64 macs, or macOS < 13.0).
 	// This setting will only be applied on vm init.
 	Rosetta *bool `yaml:"rosetta,omitempty"`
+	HostGatewayIP *string `yaml:"host_gateway_ip,omitempty"`
```
pendo324, line 53: nit: formatting

pkg/config/nerdctl_config_applier.go
```diff
@@ -9,10 +9,12 @@ import (
 	"path"
 	"strings"
 
+	"github.com/lima-vm/lima/pkg/networks"
```
pendo324, line 12: nit: import order

pkg/config/nerdctl_config_applier.go
```diff
@@ -23,6 +25,7 @@ const (
 )
 
 type nerdctlConfigApplier struct {
+	fc			   *Finch
```
pendo324, line 28: nit: formatting

pkg/config/nerdctl_config_applier.go
```diff
@@ -33,8 +36,9 @@ var _ NerdctlConfigApplier = (*nerdctlConfigApplier)(nil)
 
 // NewNerdctlApplier creates a new NerdctlConfigApplier that
 // applies nerdctl configuration changes by SSHing to the lima VM to update the nerdctl configuration file in it.
-func NewNerdctlApplier(dialer fssh.Dialer, fs afero.Fs, privateKeyPath, hostUser string) NerdctlConfigApplier {
+func NewNerdctlApplier(fc *Finch, dialer fssh.Dialer, fs afero.Fs, privateKeyPath, hostUser string) NerdctlConfigApplier {
 	return &nerdctlConfigApplier{
+		fc: 			fc,
```
pendo324, line 41: nit: formatting

pendo324 (commented):
concept / implementation LGTM, just a few more minor things in addition to what @ningziwen already commented on

pkg/config/nerdctl_config_applier.go
```diff
@@ -96,6 +100,12 @@ func updateNerdctlConfig(fs afero.Fs, user string, rootless bool) error {
 	}
 
 	var cfg Nerdctl
+	if (fc.HostGatewayIP == nil) {
```
ningziwen, line 103: @TrungBui59 I just made it work by moving the logic after line 116.

Basically line 109-116 reads the current existing nerdctl config and fill it to `cfg`. So in vm start scenario, the existing host gateway ip will be filled to `cfg` and overwrites the value you filled in line 106.

Moving it after 116 could let you overwrite the value of existing nerdctl config.

pkg/config/nerdctl_config_applier.go
TrungBui59: How can I know prevent these nit issue? I ran golang-lint and it didnt give me anything

TrungBui59:
> Could you add e2e tests? The host-gateway-ip in finch config is under e2e/vm? Similar to https://github.com/runfinch/finch/blob/main/e2e/vm/additional_disk_test.go
> 
> The host-gateway-ip arg and env var can be added in common-tests repo around here. (can be a separated PR as it is separated repo) https://github.com/runfinch/common-tests/blob/6c7275007bf34fb6ddecc4013c16f1d79ff6d1d0/tests/run.go#L335

Sure, I will also need to change some unit test

pkg/config/nerdctl_config_applier.go
ginglis13: Hi @TrungBui59, golang-lint will lint code, but not format it. You can use `gofmt -w .` from the top level directory of a Go project to format all Go files, writing the correctly formatted files. Go plugins for modern IDEs will typically do this every time a Go file is saved.

See https://go.dev/blog/gofmt

cmd/finch/nerdctl_test.go
TrungBui59: @ningziwen I already delete those unrelated tests. I add 2 more to the nerdctl_config_applier_test on my latest commit

ningziwen: Did you push it? It is not reflected in the PR.

TrungBui59: Yes, you can check my last commit to the PR

TrungBui59: @ningziwen can you check again?

cmd/finch/nerdctl_test.go
```diff
@@ -351,32 +351,6 @@ func TestNerdctlCommand_run(t *testing.T) {
...
-				logger.EXPECT().Debugf(`Resolving special IP "host-gateway" to %q for host %q`, "192.168.5.2", "name")
-				ncsd.EXPECT().LookupEnv("COSIGN_PASSWORD").Return("", false)
-				c := mocks.NewCommand(ctrl)
-				lcc.EXPECT().Create("shell", limaInstanceName, "sudo", "-E", nerdctlCmdName, "run",
-					"--rm", "--add-host", "name:192.168.5.2", "alpine:latest").Return(c)
-				c.EXPECT().Run()
-			},
-		},
 		{
 			name:    "with --add-host flag but without using special IP by space",
```
ningziwen, line 355: Do we need these two?

cmd/finch/support_bundle.go
```diff
@@ -60,7 +60,7 @@ func newGenerateSupportBundleAction(
 	}
 }
 
-func (gsa *generateSupportBundleAction) runAdapter(cmd *cobra.Command, args []string) error {
+func (gsa *generateSupportBundleAction) runAdapter(cmd *cobra.Command, cargs []string) error {
```
ningziwen, line 63: No need this change?

pkg/config/nerdctl_config_applier_test.go
```diff
@@ -174,9 +176,76 @@ func Test_updateNerdctlConfig(t *testing.T) {
 			t.Parallel()
 
 			fs := afero.NewMemMapFs()
+			fc := &Finch{
```
ningziwen, line 179: How about making fc one field in testCases struct so you can have all the test cases in one function instead of two?

cmd/finch/support_bundle.go
TrungBui59: this is not my change, I will take a look why it is here

cmd/finch/nerdctl_test.go
TrungBui59: I delete those two

ningziwen: "with --add-host flag but without using special IP by space" is still in the PR.

cmd/finch/nerdctl_test.go
```diff
@@ -351,57 +351,6 @@ func TestNerdctlCommand_run(t *testing.T) {
...
-				ncsd.EXPECT().LookupEnv("FINCH_DOCKER_COMPAT").Return("", false)
-				ncsd.EXPECT().LookupEnv("COSIGN_PASSWORD").Return("", false)
-				c := mocks.NewCommand(ctrl)
-				lcc.EXPECT().Create("shell", limaInstanceName, "sudo", "-E", nerdctlCmdName, "run",
-					"--rm", "--add-host", "name:0.0.0.0", "alpine:latest").Return(c)
-				c.EXPECT().Run()
-			},
-		},
 		{
 			name:    "with --add-host flag but without subsequent arg",
```
ningziwen, line 355: Do we still need this case?

TrungBui59: I thought it gonna check the error for add-host flag?

ningziwen: I feel it is not necessary to do unit tests for any args that are not specially handled in the code. We can use the e2e test to cover these cases.

TrungBui59: ok, I already push another one to fix delete this one

TrungBui59:
@ningziwen sorry for the late response, I have finalized by adding unit tests and e2e tests

cmd/finch/nerdctl.go
```diff
@@ -9,8 +9,8 @@ import (
 	"path/filepath"
 	"strings"
 
-	dockerops "github.com/docker/docker/opts"
-	"github.com/lima-vm/lima/pkg/networks"
+	// dockerops "github.com/docker/docker/opts"
```
ningziwen, line 12: Debug leftover

pkg/config/config.go
```diff
@@ -54,7 +54,8 @@ type Finch struct {
 	// Requires macOS 13.0 or later and an Apple Silicon (ARM64) mac.
 	// Has no effect on systems where Rosetta 2 is not available (Intel/AMD64 macs, or macOS < 13.0).
 	// This setting will only be applied on vm init.
-	Rosetta *bool `yaml:"rosetta,omitempty"`
+	Rosetta     *bool   `yaml:"rosetta,omitempty"`
+	HostGateway *string `yaml:"host_gateway_ip,omitempty"`
```
ningziwen, line 58: Let's just name it "HostGatewayIP" to prevent confusion.

e2e/vm/host_gateway_ip_test.go
```diff
@@ -0,0 +1,57 @@
...
+	"github.com/onsi/gomega/gexec"
+	"github.com/runfinch/common-tests/command"
+	"github.com/runfinch/common-tests/option"
+
+	"github.com/runfinch/finch/e2e"
+)
+
+const defaultImage    = "public.ecr.aws/docker/library/alpine:latest"
+
+var testHostGatewayIp = func(o *option.Option, installed bool) {
```
ningziwen, line 18: Could you do some Finch config file content check for each case? Similar to here https://github.com/runfinch/finch/blob/baf645f250344a42bc4ee3ea53e97b2c82dbb55a/e2e/vm/config_test.go#L116.

Ensure there is no host-gateway-ip if it is not specified and ensure it exists if specfied.

e2e/vm/host_gateway_ip_test.go
```diff
@@ -0,0 +1,57 @@
...
+			gomega.Expect(mapping).Should(gomega.ContainSubstring("test-host"))
+		})
+		ginkgo.It("should add a custom host-to-IP mapping with --add-host flag: custom value", func() {
+			startCmdSession := updateAndApplyConfig(o, []byte("memory: 4GiB\ncpus: 6\nhost_gateway_ip: 192.168.31.1"))
+			gomega.Expect(startCmdSession).Should(gexec.Exit(0))
+			mapping := command.StdoutStr(o, "run", "--add-host", "test-host:host-gateway", defaultImage, "cat", "/etc/hosts")
+			gomega.Expect(mapping).Should(gomega.ContainSubstring("192.168.31.1"))
+			gomega.Expect(mapping).Should(gomega.ContainSubstring("test-host"))
+		})
+	})
```
ningziwen, line 56: Could you also add two test cases for host-gateway-ip as cli arg and env var?

e2e/vm/host_gateway_ip_test.go
TrungBui59: @ningziwen  how do I access the toml file for nerdctl?

ningziwen: I think it is  similar to [here](https://github.com/runfinch/finch/blob/baf645f250344a42bc4ee3ea53e97b2c82dbb55a/e2e/vm/config_test.go#L95C19-L95C46).
```
finchCfgBuf = os.ReadFile(filepath.Clean(finchconfigFilepPath))
err = yaml.Unmarshal(finchCfgBuf, &finchCfg)
```

e2e/vm/host_gateway_ip_test.go
```diff
@@ -0,0 +1,58 @@
...
+	"github.com/onsi/gomega"
+	"github.com/onsi/gomega/gexec"
+	"github.com/runfinch/common-tests/command"
+	"github.com/runfinch/common-tests/option"
+
+	"github.com/runfinch/finch/e2e"
+)
+
+const defaultImage    = "public.ecr.aws/docker/library/alpine:latest"
+const nerdctlRootlessCfgPath = "/home/root.linux/.config/nerdctl/nerdctl.toml"
```
ningziwen, line 17: Why need this?

e2e/vm/host_gateway_ip_test.go
TrungBui59: @ningziwen I thought this is for the lima yaml file, the host_gateway_ip will be update to the nerdctl toml file, so how can I get that file to check it?

ningziwen: Yeah but you can replace it to finch config file instead of Lima. You already have `finchConfigFilePath` as a variable and already use it in this file (at line 24).

I think it is not necessary to check nerdctl toml file in Finch e2e tests because finch e2e tests test Finch user visible behaviours. (although it doesn't hurt to have)

TrungBui59: oh I see what you mean, thank you very much

e2e/vm/host_gateway_ip_test.go
TrungBui59: oh I thought I need to access the toml file for the host_gateway_ip value since it is update there

pkg/config/config.go
```diff
@@ -54,7 +54,8 @@ type Finch struct {
 	// Requires macOS 13.0 or later and an Apple Silicon (ARM64) mac.
 	// Has no effect on systems where Rosetta 2 is not available (Intel/AMD64 macs, or macOS < 13.0).
 	// This setting will only be applied on vm init.
-	Rosetta *bool `yaml:"rosetta,omitempty"`
+	Rosetta     *bool   `yaml:"rosetta,omitempty"`
+	HostGatewayIp *string `yaml:"host_gateway_ip,omitempty"`
```
ningziwen, line 58: Please make the naming consistent with line 76. Probably follow the pattern in nerdctl. https://github.com/containerd/nerdctl/blob/1f8f8af4d19d651a6300d8d48f3f21c91048a716/pkg/config/config.go#L41

`HostGatewayIP`

ningziwen (approved):
One nit pick. Also need to fix the errors in CI. 

Others LGTM. 

Need a follow-up PR in https://github.com/runfinch/common-tests to test CLI flag and env var.

pkg/config/config.go
TrungBui59: Just fix it

ginglis13: Please add godoc comment i.e.
```suggestion
        // HostGatewayIp sets the default host gateway ip....
	HostGatewayIp *string `yaml:"host_gateway_ip,omitempty"`
```

TrungBui59:
> One nit pick. Also need to fix the errors in CI.
> 
> Others LGTM.
> 
> Need a follow-up PR in https://github.com/runfinch/common-tests to test CLI flag and env var.

I just finish with it, I will do the follow up work on the common tests file. Can you assign me another issue to do?

pkg/config/nerdctl_config_applier.go
```diff
@@ -124,6 +126,10 @@ func updateNerdctlConfig(fs afero.Fs, user string, rootless bool) error {
 	}
 
 	cfg.Namespace = nerdctlNamespace
+	cfg.HostGatewayIP = "192.168.5.2"
```
ginglis13, line 129: instead of having this magic string down here, could we define as a constant `defaultNerdctlHostGatewayIP` here https://github.com/runfinch/finch/pull/453/files#diff-531afb9f97a7384e0de895046f9e14befbebc49c896b080fce4c41c75f14f1b4L20-L21

then you can also reference it in unit tests where you expect the `HostGatewayIP` to be the default value.

ningziwen:
> > One nit pick. Also need to fix the errors in CI.
> > Others LGTM.
> > Need a follow-up PR in https://github.com/runfinch/common-tests to test CLI flag and env var.
> 
> @ningziwen I just finish with it, I will do the follow up work on the common tests file. Can you assign me another issue to do?

Same issue is fine. One issue can map to multiple PRs.

ningziwen:
Could you make the PR title more accurate? Such as "feat: make host gateway ip configurable". The PR title will be the final commit message.

ningziwen:
Please follow the guideline to fix DCO. https://github.com/runfinch/finch/pull/453/checks?check_run_id=16527210509

go linter and go-mod-tidy-check are still failing.

TrungBui59:
> > > One nit pick. Also need to fix the errors in CI.
> 
> > > Others LGTM.
> 
> > > Need a follow-up PR in https://github.com/runfinch/common-tests to test CLI flag and env var.
> 
> > 
> 
> > @ningziwen I just finish with it, I will do the follow up work on the common tests file. Can you assign me another issue to do?
> 
> 
> 
> Same issue is fine. One issue can map to multiple PRs.

Sure, I will do it. I mean like after the follow up work, can you assign me another one? I am still a newbie with Finch so I will need sometime to understand and find a way to address it

cmd/finch/nerdctl_test.go
```diff
@@ -365,144 +365,6 @@ func TestNerdctlCommand_run(t *testing.T) {
 				logger.EXPECT().Debugf("Status of virtual machine: %s", "Running")
 			},
 		},
-		{
```
ginglis13, line 368: should we still keep these tests or is upstream testing in nerdctl sufficient ? cc: @ningziwen

ningziwen: These are unit tests so I think they are not related to upstream testing. The test parity of nerdctl upstream tests here should be Finch e2e tests instead of unit tests. We already have them covered.

From what I understand, the unit tests should strictly test the unit. Now the special handling logic is removed from nerdctl.go so the unit tests are also being removed from nerdctl_tests.go. We already mocked nerdctl in unit tests so any logic inside nerdctl should not be tested here.

ningziwen:
> > > > One nit pick. Also need to fix the errors in CI.
> > 
> > 
> > > > Others LGTM.
> > 
> > 
> > > > Need a follow-up PR in https://github.com/runfinch/common-tests to test CLI flag and env var.
> > 
> > 
> > > 
> > 
> > 
> > > @ningziwen I just finish with it, I will do the follow up work on the common tests file. Can you assign me another issue to do?
> > 
> > 
> > Same issue is fine. One issue can map to multiple PRs.
> 
> Sure, I will do it. I mean like after the follow up work, can you assign me another one? I am still a newbie with Finch so I will need sometime to understand and find a way to address it

@TrungBui59 We don't do assignments proactively normally. Feel free to try out any issues that you are interested. If you are concerned about the difficulty, you can filter "good first issue". If you are not clear about the requirements of specific issue, you can also reply to clarify. There are also other simple but interesting issues under other subprojects of runfinch, such as https://github.com/runfinch/common-tests/issues/72. Once you find the issue you want to take, you can reply it and we will assign to you.

ningziwen:
@TrungBui59 e2e tests are failing in CI. Could you fix it?

PR #462, opened 2023-06-28
feat: ECR credential integration into Finch 

kiryl1:
Issue #, if available: 116

*Description of changes:*
Integrating ECR credential helper into Finch. 

**To Install the Credential Helper**
- The ECR credential helper can be setup with minimal configuration by setting the `creds_helpers` option in `finch.yaml` with the value `- ecr-login`.

- Once the option has been set the credential helper will be installed on either `finch vm init `  or `finch vm start`. The binary will be downloaded on the host machine and a `config.json` will be created and populated in the `~/.finch/` folder if it doesn't already exist. If it already exists, the value of `credsStore` will be overwritten to `ecr-login` .

**How to stop using the credential helper with Finch**
- `config.json` needs to be either deleted, or the `credsStore` parameter needs to have the value `ecr-login ` removed.  Additionally the creds_helper option in the finch.yaml needs to be removed. 

- To fully remove the credential helper from the host machine, this can be done by deleting it from the 
`~/.finch/cred-helpers` folder.

The credential helper will support credentials from both the aws credentials file and those that are stored as environment variables. 

*Testing done:*
Added Unit Tests
Added Integration Test 

- [ x] I've reviewed the guidance in CONTRIBUTING.md

#### License Acceptance


pkg/dependency/credhelper/cred_helper_binary.go
```diff
@@ -0,0 +1,214 @@
...
+
+	"github.com/runfinch/finch/pkg/command"
+	"github.com/runfinch/finch/pkg/config"
+	"github.com/runfinch/finch/pkg/dependency"
+	"github.com/runfinch/finch/pkg/flog"
+	"github.com/runfinch/finch/pkg/path"
+)
+
+// ConfigFile is taken from https://github.com/docker/cli/blob/master/cli/config/configfile/file.go
+type ConfigFile struct {
```
pendo324, line 23: Can we just import this? Something like (at the top):

```go
"https://github.com/docker/cli/blob/master/cli/config/configfile"
```

And then when you use the struct:

```go
configfile.ConfigFile
```

pkg/dependency/credhelper/cred_helper_binary_test.go
```diff
@@ -0,0 +1,278 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+// Ensures that the binaries required for networking are installed in a privileged location.
```
pendo324, line 4: nit: comment. Probably just remove it

pendo324 (changes requested):
Left some comments. You'll also need to sign your commit with the DCO by using the `-s` flag when you run `git commit` ([more info](https://github.com/apps/dco))

pkg/config/config.go
```diff
@@ -33,8 +33,9 @@ type AdditionalDirectory struct {
 
 // Finch represents the configuration file for Finch CLI.
 type Finch struct {
-	CPUs   *int    `yaml:"cpus"`
-	Memory *string `yaml:"memory"`
+	CPUs        *int    `yaml:"cpus"`
+	Memory      *string `yaml:"memory"`
+	CredsHelper *string `yaml:"credsHelper,omitempty"`
```
vsiravar, line 38: Should probably make this a slice here since a user can install multiple credential helpers (although we only support ecr credentials helper with this change) and modify the Install [method](https://github.com/runfinch/finch/pull/462/files#diff-7ac37c68742ad8a4a56915628dcc400e073c8953e175af08b4ba4a63370a54cb).

pkg/config/nerdctl_config_applier.go
```diff
@@ -56,18 +67,26 @@ func NewNerdctlApplier(dialer fssh.Dialer, fs afero.Fs, privateKeyPath, hostUser
...
 //
 // [registry nerdctl docs]: https://github.com/containerd/nerdctl/blob/master/docs/registry.md
+
 func updateEnvironment(fs afero.Fs, user string) error {
+	cmdArr := [4]string{
+		fmt.Sprintf("export DOCKER_CONFIG=\"/Users/%s/.finch\"", user),
+		fmt.Sprintf("[ -L /usr/local/bin/docker-credential-ecr-login ] "+
+			"|| sudo ln -s /Users/%s/.finch/cred-helpers/docker-credential-ecr-login /usr/local/bin/", user),
+		fmt.Sprintf("[ -L $HOME/.aws ] || ln -s  /Users/%s/.aws $HOME/.aws", user),
+		fmt.Sprintf("[ -L /root/.aws ] || sudo ln -fs  /Users/%s/.aws /root/.aws", user),
```
vsiravar, line 77: Why do we need this line?

pkg/dependency/credhelper/cred_helper_binary.go
```diff
@@ -0,0 +1,188 @@
...
+
+func (bin *binaries) Install() error {
+	credsHelper := bin.fc.CredsHelper
+	if credsHelper == nil {
+		return nil
+	}
+	if strings.Compare(*credsHelper, bin.credHelperConfigName()) != 0 {
+		return nil
+	}
+	bin.l.Info("Installing credential helper")
```
vsiravar, line 161: Adding the name of the cred helper to the info log will help.

pkg/dependency/credhelper/cred_helper_binary.go
```diff
@@ -0,0 +1,188 @@
...
+	if err != nil {
+		return fmt.Errorf("error creating installation directory %s, err: %w", bin.hcfg.installFolder, err)
+	}
+
+	curlCmd := bin.cmdCreator.Create("curl", "--retry", "5", "--retry-max-time", "30", "--url",
+		bin.hcfg.credHelperURL, "--output", bin.fullInstallPath())
+
+	_, err = curlCmd.Output()
+	if err != nil {
+		return fmt.Errorf("error installation binary %s, err: %w", bin.hcfg.binaryName, err)
```
vsiravar, line 173: nit: typo , "installation" -> "installing"

vsiravar (changes requested):
Left a few minor comments. Overall looks good. Suggest adding a note in the [README](https://github.com/runfinch/finch/blob/main/README.md#configuration) similar to `additional_directories` explaining what the user experience is going to be like.

e2e/vm/cred_helper_test.go
```diff
@@ -0,0 +1,37 @@
...
+	ginkgo.Describe("Credential Helper", func() {
+		ginkgo.Describe("Credential Helper", func() {
+			ginkgo.It("should pull from container registry", func() {
+				resetVM(o, installed)
+				resetDisks(o, installed)
+				if registry == "" {
+					ginkgo.Skip("No Provided Container Registry Url")
+				}
+				writeFile(finchConfigFilePath, []byte("memory: 4GiB\ncredsHelper: ecr-login\ncpus: 6\nvmType: vz\nrosetta: false"))
+				cfgJSONPath := strings.ReplaceAll(finchConfigFilePath, "finch.yaml", "config.json")
```
pendo324, line 26: Shouldn't we be checking if Finch updates the config.json path itself rather than just setting it here?

pkg/dependency/credhelper/cred_helper_binary.go
```diff
@@ -0,0 +1,189 @@
...
+	"github.com/runfinch/finch/pkg/command"
+	"github.com/runfinch/finch/pkg/config"
+	"github.com/runfinch/finch/pkg/dependency"
+	"github.com/runfinch/finch/pkg/flog"
+	"github.com/runfinch/finch/pkg/path"
+
+	"github.com/docker/cli/cli/config/configfile"
+)
+
+type binaries struct {
```
pendo324, line 24: nit: rename to `credhelperbin` or similar

pendo324 (commented):
Looks good, just a few things. Also think we have to update the GitHub workflow files and Makefile to actually add the Registry option, right?

pendo324 (approved):
Changes look good to me. Lets take it out of draft and work on configuring the integration tests to run with the correct registry by adding the secret value and then using it in [ci.yaml](https://github.com/runfinch/finch/blob/19392e1bd40d52bde469c73ceee9fbd771bb06f8/.github/workflows/ci.yaml#L123)?

e2e/vm/cred_helper_test.go
```diff
@@ -0,0 +1,31 @@
...
+	"github.com/onsi/ginkgo/v2"
+	"github.com/onsi/gomega"
+	"github.com/onsi/gomega/gexec"
+	"github.com/runfinch/common-tests/command"
+	"github.com/runfinch/common-tests/option"
+)
+
+var testCredHelper = func(o *option.Option, installed bool, registry string) {
+	ginkgo.Describe("Credential Helper", func() {
+		ginkgo.Describe("Credential Helper", func() {
```
vsiravar, line 16: nit: Don't require this nested `Describe` node.

e2e/vm/cred_helper_test.go
```diff
@@ -0,0 +1,31 @@
...
+			ginkgo.It("should pull from container registry", func() {
+				resetVM(o, installed)
+				resetDisks(o, installed)
+				if registry == "" {
+					ginkgo.Skip("No Provided Container Registry Url")
+				}
+				writeFile(finchConfigFilePath, []byte("memory: 4GiB\ncredsHelper: ecr-login\ncpus: 6\nvmType: vz\nrosetta: false"))
+				initCmdSession := command.New(o, virtualMachineRootCmd, "init").WithTimeoutInSeconds(600).Run()
+				gomega.Expect(initCmdSession).Should(gexec.Exit(0))
+				command.New(o, "pull", registry).WithTimeoutInSeconds(600).Run()
```
vsiravar, line 26: Should assert that this command did indeed pull from the intended registry and succeeded.

e2e/vm/cred_helper_test.go
```diff
@@ -0,0 +1,31 @@
...
+				gomega.Expect(command.Stdout(o, "images", "-q", registry)).NotTo(gomega.BeEmpty())
+			})
+		})
+	})
+}
```
vsiravar, line 31: Is there going to be a test for `push` in a subsequent PR?

vsiravar:
We use [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/) for creating [finch release](https://github.com/runfinch/finch/blob/main/.github/workflows/release-please.yaml) based on semantic versioning. We should probably prefix the PR title as `feat:` .

e2e/vm/cred_helper_test.go
pendo324: Push or pull should both invoke the same credential helper behavior. Not sure we really need to test both separately.

pendo324 (commented):
Still need the changes to actually use the remote registry unless I'm missing something, otherwise LGTM

pkg/dependency/credhelper/cred_helper_binary.go
```diff
@@ -0,0 +1,189 @@
...
+	fileExists, err := afero.Exists(bin.fs, cfgPath)
+	if err != nil {
+		return err
+	}
+	if !fileExists {
+		file, err := bin.fs.Create(cfgPath)
+		if err != nil {
+			return err
+		}
+		JSONstr := fmt.Sprintf("{\"credsStore\":\"%s\"}", binCfgName)
```
ginglis13, line 63: I would assume it's out of scope for this PR, but might be worth opening an issue.

See https://github.com/awslabs/amazon-ecr-credential-helper/issues/474#issuecomment-1538822507

Users can also configure credential helper as

```
{
	"credHelpers": {
		"public.ecr.aws": "ecr-login",
		"<aws_account_id>.dkr.ecr.<region>.amazonaws.com": "ecr-login"
	}
}
```

pkg/dependency/credhelper/cred_helper_binary.go
```diff
@@ -0,0 +1,189 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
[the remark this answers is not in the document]
ginglis13: also some godoc comments above at least the public functions in this file would be helpful

pkg/dependency/credhelper/cred_helper_binary.go
```diff
@@ -0,0 +1,189 @@
...
+	if err != nil {
+		bin.l.Error(err)
+		return false
+	}
+	hash, err := digest.FromReader(file)
+	if err != nil {
+		bin.l.Error(err)
+		return false
+	}
+	defer file.Close() //nolint:errcheck // closing the file
```
ginglis13, line 141: nit: move this up to before the call to `digest.FromReader`. Otherwise, if reading that hash results in error, the successfully opened file would not be closed.

pkg/dependency/credhelper/cred_helper_binary.go
```diff
@@ -0,0 +1,189 @@
...
+		err := bin.fs.Remove(bin.fullInstallPath())
+		if err != nil {
+			bin.l.Error(err)
+		}
+		return false
+	}
+	return true
+}
+
+func (bin *credhelperbin) Install() error {
```
ginglis13, line 153: should you be setting `bin.hcfg.hash` in this function?

vsiravar:
> Still need the changes to actually use the remote registry unless I'm missing something, otherwise LGTM

Yeah, I think it may be worth exporting the aws credentials like how we export [COSIGN_PASSWORD](https://github.com/runfinch/finch/blob/main/cmd/finch/nerdctl.go#L173) if users wish to use env vs file. Might also simply the e2e test which will rely on https://github.com/aws-actions/configure-aws-credentials which exports the credentials as env variables.

pkg/dependency/credhelper/cred_helper_binary.go
kiryl1: The hash is mainly for checking that the downloaded binary hash matches the hardcoded one to ensure that the correct file is installed. The hash in bin.hcfg.hash is hard coded so I don't really see a point in setting it there.

pendo324 (approved):

vsiravar (dismissed):
LGTM, we can merge once the linter check passes. https://github.com/runfinch/finch/actions/runs/5526370034/jobs/10097740941?pr=462#step:4:18795

ginglis13 (approved):
LGTM

.github/workflows/ci.yaml
```diff
@@ -82,21 +85,27 @@ jobs:
...
-    runs-on: ${{ matrix.os }} 
+    runs-on: ${{ matrix.os }}
     steps:
       - uses: actions/checkout@v3
         with:
           # We need to get all the git tags to make version injection work. See VERSION in Makefile for more detail.
           fetch-depth: 0
           persist-credentials: false
           submodules: true
+      - name: configure aws credentials
```
pendo324, line 103: I think we have to scope this to only run on push/merge, see https://github.com/aws/s2n-quic/blob/0ebd9560194364b2419aabbacf1c73634801b74d/.github/workflows/book.yml#L48-L53 for an example

.github/workflows/ci.yaml
```diff
@@ -162,7 +179,7 @@ jobs:
       - run: |
           git status
           git clean -f -d
-          FINCH_DOCKER_COMPAT=1 make test-e2e
+          FINCH_DOCKER_COMPAT=1 REGISTRY=${{ env.REGISTRY }} make test-e2e
```
pendo324, line 182: I think we also need to add a guard here. Might want to make the variable reusable earlier on, in its own step, as well.

Something like this should work. Otherwise, the test is going to try to pull from the registry but it won't have creds to make the pull actually work

```
      - name: Set reusable variables
        id: vars
        run: |
          has_creds=${{ github.event_name == 'push' || github.repository == github.event.pull_request.head.repo.full_name }}
          echo "has_creds=$has_creds" >> $GITHUB_OUTPUT
      - name: configure aws credentials
        uses: aws-actions/configure-aws-credentials@v2.2.0
        if: steps.vars.outputs.has_creds == true
        with:
...
     - run: |
          git status
          git clean -f -d
          FINCH_DOCKER_COMPAT=1 REGISTRY=${{ steps.vars.outputs.has_creds == true && env.REGISTRY || '' }} make test-e2e
```

pendo324 (dismissed):

vsiravar (approved):

ginglis13 (dismissed):

pkg/config/config.go
ningziwen: What is the resolution of this conversatino?

ningziwen, line 38: It is sad that the yaml field naming is not consistent now. There is not an official YAML naming pattern though. https://www.w3schools.io/file/yaml-naming-conventions/#:~:text=There%20are%20no%20official%20naming,C%23%20use%20kebab%20style%20naming.

We can either make credHelper consistent to VMType or additional_directories. Should we follow additional_directories and VMType is experiment and temporary?

cmd/finch/nerdctl.go
```diff
@@ -170,7 +170,11 @@ func (nc *nerdctlCommand) run(cmdName string, args []string) error {
 		envVars[evar] = eval
 	}
 
-	passedEnvs := []string{"COSIGN_PASSWORD"}
+	passedEnvs := []string{
```
ningziwen, line 173: Could you add some design context in the PR description? What are the user scenarios are we trying to support here? 
Is the env var the only way to pass AWS credential? Any other possible ways (like credential file)? Do we need to support them? https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-files.html 

Once we clarify the scope, I think there can be some docs to guide users using the feature. "ECR credential helper integration" is more like the implementation details but the user facing feature should be something like "simplify ECR interfaction in FInch". I think the documentation should focus on its usage.

ginglis13:
Wanted to callout an unexpected behavior which we have options to resolve. Consider this non-blocking for the PR but follow up work. I follow these steps after successfully configured the credential helper and verifying that push and pull work with my ECR repo:

1. `finch vm stop`
2. `finch vm remove`
3. remove `credsStore: ecr-login` from `~/.finch/finch.yaml`
4. finch vm init,
5. pull  -> Success!

I no longer have finch configured with a `credsStore` in `~/.finch/finch.yaml`, but I am still able to auth, push and pull with finch. This is because `~/.finch/config.json` still contains the `"credsStore":"ecr-login"` entry. What's expected? Should this be removed automatically? Or should users be expected to edit this files themselves? (I personally vote the former)

I can turn this comment into an issue if you folks agree @kiryl1 @pendo324 @vsiravar

pkg/config/config.go
kiryl1: Out of scope for current PR, this can be added when multiple cred helpers are supported.

pendo324:
> What's expected? Should this be removed automatically? Or should users be expected to edit this files themselves? (I personally vote the former)
> 
> I can turn this comment into an issue if you folks agree @kiryl1 @pendo324 @vsiravar

+1 for turning this into an issue. I think the expected behavior to remove the feature is probably to just edit the same config file again, not also have to edit `config.json`. This is a pretty big PR so following it up in another more focussed PR makes sense to me. We should just make sure to document the current behavior properly @kiryl1

pkg/config/config.go
ningziwen: Wouldn't it be a breaking change to add multiple cred helpers late for renaming and changing data types?

cmd/finch/nerdctl.go
kiryl1: Both the env variable and the credential file are supported. I will add some documentation to the README.

pkg/config/config.go
kiryl1: I think for consistency we can do the same as additional_directories since it isn't a required field either.

vsiravar:
> Should this be removed automatically? 

+1 for this, it'll end up confusing the user if the config is missing in `~/.finch/finch.yaml` but the finch still tries to use the credentials helper.

pkg/config/config.go
kiryl1: This is true, if preferred I can add this change. However I believe that adding this change can give the impression to the user that multiple cred helpers are supported even though this isn't the case.

ningziwen: I think it is pretty common to keep the door open for future potential changes.

For example:
In Docker API, `/containers/json` always returns one name in `Names` field. https://docs.docker.com/engine/api/v1.43/#tag/Container

In ECR Public, `DescribeRegistries` always returns one registry. https://docs.aws.amazon.com/AmazonECRPublic/latest/APIReference/API_DescribeRegistries.html

Now seems we are believing there is a large chance to support multiple cred helpers later (unless we are not). We can't just keep the door close and leave the breaking change happen in the future.

README.md
```diff
@@ -99,6 +99,17 @@ An example `finch.yaml` looks like this:
 cpus: 4
 # Memory: the amount of memory to dedicate to the virtual machine. (required)
 memory: 4GiB
+# CredsHelper: a list of credential helpers that will be installed and configured automatically.
```
ningziwen, line 102: ?CredsHelpers"

README.md
```diff
@@ -99,6 +99,17 @@ An example `finch.yaml` looks like this:
...
 memory: 4GiB
+# CredsHelper: a list of credential helpers that will be installed and configured automatically. 
+# Currently,  ecr-login is the only supported option.
+# Once the option has been set the credential helper will be installed on either finch vm init or finch vm start. 
+# The binary will be downloaded on the host machine and a config.json will be created and populated inside the ~/.finch/ folder 
+# if it doesn't already exist. If it already exists, the value of credsStore will be overwritten. 
+# To opt out of using the credential helper, remove the value from the credsStore parameter of config.json 
+# and remove the creds_helper value from finch.yaml. 
+# To completely remove the credential helper, either remove the binary from ~/.finch/creds-helpers or remove the creds-helpers
+# folder entirely. (optional)
```
ningziwen, line 110: This still didn't clarify how to use the ecr credential helper. Maybe add something like "supported credential helper list: ecr-login" and put the ecr credential helper link here?

pendo324 (approved):

ginglis13:
Are e2e tests using the credential helper? For example [in build 1018 `CI / e2e-tests (self-hosted, macos, amd64, 13, test)`](https://github.com/runfinch/finch/actions/runs/5578802929/jobs/10195214219?pr=462):

```
# Set output variables
has_creds=false
```

And 

```
Finch Container Development E2E Tests log in a container registry when the private registry is running and an image is built should push an image after successfully logging in the registry with a correct credential

...

  WARNING: Your password will be stored unencrypted in /Users/ec2-user/.finch/config.json.
  Configure a credential helper to remove this warning. See
  https://docs.docker.com/engine/reference/commandline/login/#credentials-store

```

Also, looks like each e2e step is running twice? e.g.. there are two occurrences of `CI / e2e-tests (self-hosted, macos, amd64, 13, test)` .

pkg/dependency/credhelper/cred_helper_binary.go
ginglis13: since this PR has been updated, the direction appears to be supporting the syntax for `"credHelpers"` rather than `"credsStore"`. Should we be updating how `config.json` sets credential helpers, then? As of right now in this PR, a user could provide multiple credential helpers in `finch.yaml`, but only one of those would be added to the config

kiryl1:
> Are e2e tests using the credential helper? For example [in build 1018 `CI / e2e-tests (self-hosted, macos, amd64, 13, test)`](https://github.com/runfinch/finch/actions/runs/5578802929/jobs/10195214219?pr=462):
> 
> ```
> # Set output variables
> has_creds=false
> ```
> 
> And
> 
> ```
> Finch Container Development E2E Tests log in a container registry when the private registry is running and an image is built should push an image after successfully logging in the registry with a correct credential
> 
> ...
> 
>   WARNING: Your password will be stored unencrypted in /Users/ec2-user/.finch/config.json.
>   Configure a credential helper to remove this warning. See
>   https://docs.docker.com/engine/reference/commandline/login/#credentials-store
> ```
> 
> Also, looks like each e2e step is running twice? e.g.. there are two occurrences of `CI / e2e-tests (self-hosted, macos, amd64, 13, test)` . (e: seeing `Run echo "Skipping CI for docs & contrib files"` in one of each of the duplicates, likely from [949a184](https://github.com/runfinch/finch/commit/949a184238007bae74275d0ba06e16947932809e))

No the existing e2e tests are not using the Credential Helper, Since the existing e2e tests are using finch login the tests would fail if the credential helper was enabled.

Im also not exactly sure why it appears to run in duplicates.

pkg/dependency/credhelper/cred_helper_binary.go
kiryl1: Since this will require additional user input in the form of `<aws_account_id>` and `<region>` I think it might be better to leave this responsibility to the user. Other wise additional fields need to be added to finch.yaml so the user can configure it.

ningziwen (approved):

ginglis13 (approved):

Issue #488, opened 2023-07-20
Finch benchmarks show incorrect label for VM init charts.


austinvazquez:
**Describe the bug**
Inspecting the data on VM init benchmarks show incorrect label on the data. The chart label is `s/op` but the data is still showing `ns/op` even after https://github.com/runfinch/finch/pull/470

**Steps to reproduce**
Hover a VM init benchmark datapoint e.g. https://runfinch.github.io/finch/dev/bench/macOS/12/arm64/index.html

**Expected behavior**
Expected label to be `s/op`

**Screenshots or logs**
If applicable, add screenshots or logs to help explain your problem.

**Additional context**
The data seems converted correctly, so probably just a missed label.

_To help debug the issue as quickly as possible, we recommend generating a support bundle with `finch support-bundle generate` and attaching it to this issue. This packages all Finch-related configs and logs into one file._

ningziwen:
Closed as completed.

PR #470, opened 2023-07-10
chore: convert from ns/op to s/op

ningziwen:
Convert from ns/op to s/op
Drag index.html to explorer to test.

- [ X ] I've reviewed the guidance in CONTRIBUTING.md

#### License Acceptance

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #493, opened 2023-07-22
fix: convert from ns/op to s/op in tooltips

ningziwen:
Issue #, if available:

*Description of changes:*

*Testing done:*

https://github.com/runfinch/finch/issues/488

Follow up fix of previous change. https://github.com/runfinch/finch/pull/470

Changing to s/op for better readability.

#### License Acceptance


monirul (approved):
Why do we need to change the value 's/op'?
Can you pls add the reason to the description of the PR?

LGTM

ningziwen:
> Why do we need to change the value 's/op'? Can you pls add the reason to the description of the PR?
> 
> LGTM

@monirul Done.

PR #506, opened 2023-07-28
feat: adding config option for SOCI installation on VM

CodeChanning:
*Description of changes:*
Adding snapshotter option in config file and integrating SOCI with Finch

**Setting Snapshotters**
- Users can set the snapshotters they'd like to use by listing them in the config file as the ```snapshotters``` value.
- All listed snapshotters will be installed if they are currently supported with Finch. As of this PR the two that are supported are OverlayFS (the default snapshotter) and SOCI.
- The first snapshotter listed will be made the default snapshotter used. Other snapshotters can be used by specifying when running commands (i.e. ```finch --snapshotter={exampleSnapshotter} pull ...```)

*Example ```snapshotters``` setting in ```finch.yaml```:*
```yaml
snapshotters:
    -overlayfs
    -soci
```

This would install SOCI on the user's VM and allow for it to be used when specified in commands, but would keep OverlayFS as the default for commands

**To Install SOCI**

- SOCI can be setup with minimal configuration by adding ```"- soci"``` to the ```snapshotters``` list in ```finch.yaml``` .

- Once the option has been set SOCI will be installed on either ```finch vm init```  or ```finch vm start```. The binary will be downloaded onto the user's VM and the needed settings for SOCI containerd configuration will be appended to ```etc/containerd/config.toml``` in the VM. If SOCI is the first snapshotter listed it will also be set as the default nerdctl snapshotter in the user's VM which would allow the user to pull images with SOCI simply by doing ```finch pull ...``` .

**To Stop Using SOCI by default**

- ```"- soci"``` should be removed or not be the first snapshotter of the ```snapshotters``` list in ```finch.yaml``` . The user can also make the first snapshotter listed ```"- overlayfs"``` to revert back to the original default used by nerdctl/containerd.

**Note:** removing a snapshotter from ```snapshotters``` list will not uninstall the snapshotter from the user's VM. In order to fully uninstall the snapshotter the user must shell into the VM and remove the binaries from ```/usr/local/bin```. 

Testing done:

[ x] I've reviewed the guidance in CONTRIBUTING.md
License Acceptance


ningziwen:
The commit message should mention config. (We are not packaing SOCI by default)

pkg/config/lima_config_applier.go
```diff
@@ -188,3 +198,46 @@ func hasUserModeEmulationInstallationScript(limaCfg *limayaml.LimaYAML) (int, bo
 
 	return scriptIdx, hasCrossArchToolInstallationScript
 }
+
+func toggleSoci(limaCfg *limayaml.LimaYAML, enabled bool) {
+	idx, hasScript := hasSociInstallationScript(limaCfg)
```
ningziwen, line 203: We may not need this. I remember it doesn't change the config persistently but from the default script every time.

ningziwen:
The unit tests can be added here:
https://github.com/runfinch/finch/blob/main/pkg/config/config_test.go
https://github.com/runfinch/finch/blob/main/pkg/config/lima_config_applier_test.go

pkg/config/lima_config_applier.go
```diff
@@ -188,3 +198,46 @@ func hasUserModeEmulationInstallationScript(limaCfg *limayaml.LimaYAML) (int, bo
...
+
+func hasSociInstallationScript(limaCfg *limayaml.LimaYAML) (int, bool) {
+	hasSociInstallationScript := false
+	var scriptIdx int
+	for idx, prov := range limaCfg.Provision {
+		trimmed := strings.Trim(prov.Script, " ")
+		if !hasSociInstallationScript && strings.HasPrefix(trimmed, sociInstallationProvisioningScriptHeader) {
+			hasSociInstallationScript = true
+			scriptIdx = idx
+		}
```
ginglis13, line 239: how about an early break of the loop?

```suggestion
		if strings.HasPrefix(trimmed, sociInstallationProvisioningScriptHeader) {
			hasSociInstallationScript = true
			scriptIdx = idx
			break
		}
```

pkg/config/lima_config_applier.go
```diff
@@ -188,3 +198,46 @@ func hasUserModeEmulationInstallationScript(limaCfg *limayaml.LimaYAML) (int, bo
...
+			Mode: "system",
+			Script: fmt.Sprintf(`%s
+if [ ! -f /usr/local/bin/soci ]; then
+    #download soci
+    curl -OL "https://github.com/awslabs/soci-snapshotter/releases/download/v0.1.0/soci-snapshotter-0.1.0-linux-arm64.tar.gz"
+    #move to usr/local/bin
+    tar -C /usr/local/bin -xvf soci-snapshotter-0.1.0-linux-arm64.tar.gz ./soci ./soci-snapshotter-grpc
+    #changing containerd config
+    export config=etc/containerd/config.toml
+    //copy config to soci-config
```
ginglis13, line 215: `# copy config to soci-config`

pkg/config/lima_config_applier.go
CodeChanning: I think when I was testing this I saw the soci installation being appended to the override.yaml multiple times when I stop and started my VM. So, I thought the toggle was necessary. I'll try to add tests to double check if this is needed though.

pkg/config/lima_config_applier.go
```diff
@@ -17,6 +17,8 @@ import (
 )
 
 const userModeEmulationProvisioningScriptHeader = "# cross-arch tools"
+const sociInstallationProvisioningScriptHeader = "# soci installation and configuring"
+const sociVersion = "0.1.0"
```
ginglis13, line 21:
```suggestion
const sociVersion = "0.3.0"
```

https://github.com/awslabs/soci-snapshotter/releases/tag/v0.3.0

CodeChanning: I'm not sure why but when I change ```sociVersion``` to anything above ```0.1.0``` locally it causes issues with the installation. I'll keep looking into this though and figure it out.

pkg/config/lima_config_applier.go
```diff
@@ -188,3 +199,50 @@ func hasUserModeEmulationInstallationScript(limaCfg *limayaml.LimaYAML) (int, bo
...
+func toggleSoci(limaCfg *limayaml.LimaYAML, enabled bool, sociVersion string, arch string) {
+	idx, hasScript := hasSociInstallationScript(limaCfg)
+	fname := fmt.Sprintf("soci-snapshotter-%s-linux-%s.tar.gz", sociVersion, arch)
+	sociDownloadUrl := fmt.Sprintf("https://github.com/awslabs/soci-snapshotter/releases/download/v%s/%s", sociVersion, fname)
+	if !hasScript && enabled {
+		limaCfg.Provision = append(limaCfg.Provision, limayaml.Provision{
+			Mode: "system",
+			Script: fmt.Sprintf(`%s
+if [ ! -f /usr/local/bin/soci ]; then
+    #download soci
```
ningziwen, line 212: Might need a `sleep 5` here. https://github.com/lima-vm/lima/issues/287

CodeChanning: I don't think the script needs to wait for any of the mounts to run correctly, so ```sleep 5``` shouldn't be necessary. It doesn't hurt to add a ```sleep``` just in case though, so I'll add it back.

pkg/config/lima_config_applier.go
```diff
@@ -188,3 +199,50 @@ func hasUserModeEmulationInstallationScript(limaCfg *limayaml.LimaYAML) (int, bo
...
+	export config=etc/containerd/config.toml
+    curl -OL "%s" >> $config
+    #move to usr/local/bin
+    tar -C /usr/local/bin -xvf %s ./soci ./soci-snapshotter-grpc
```
ningziwen, line 216: `./` doesn't work. Need to remove it. Follow what you had in nerdctl. https://github.com/containerd/nerdctl/blob/1a7e89dcc9edbee903adbb49756f2575b8450c53/Dockerfile#L290

pkg/config/lima_config_applier.go
ningziwen: `0.3.0` has to be used here. `0.1.0` doesn't work with your nerdctl code.

pkg/config/lima_config_applier.go
ningziwen: If you are sure your provisioning scripts don't deal with mounts, don't need to add it.

pkg/config/lima_config_applier.go
CodeChanning: Update on this. The toggle isn't needed to stop the script from being appended multiple times, but I found that it is needed to remove the script and env var from the lima config when ```soci_snapshotter``` is set from true to false.

pkg/config/lima_config_applier_test.go
```diff
@@ -85,8 +86,133 @@ fi
...
+  tar -C /usr/local/bin -xvf %s soci soci-snapshotter-grpc
+  #changing containerd config
+  export config=etc/containerd/config.toml
+  echo "    [proxy_plugins.soci]
+  type = \"snapshot\"
+  address = \"/run/soci-snapshotter-grpc/soci-snapshotter-grpc.sock\" " >> $config
+
+  sudo systemctl restart containerd.service
+  sudo soci-snapshotter-grpc &> ~/soci-snapshotter-logs &
+fi
```
ginglis13, line 142: this string is the same as that defined in `toggleSoci`, correct? Can you make it a constant and refer to it here and in `toggleSoci` so it can be maintained easier if it ever needs changing?

same comment for the `# cross-arch tools` string

e2e/vm/soci_test.go
```diff
@@ -0,0 +1,60 @@
...
+			finchPullMounts := countMounts(limactlO)
+			command.New(o, "rmi", "-f", FfmpegSociImage).WithTimeoutInSeconds(30).Run()
+			command.New(limactlO, "shell", "finch", "sudo", "nerdctl", "pull", FfmpegSociImage).WithTimeoutInSeconds(30).Run().Out.Contents()
+			nerdctlPullMounts := countMounts(limactlO)
+			command.New(o, "rmi", "-f", FfmpegSociImage).WithTimeoutInSeconds(30).Run()
+			gomega.Expect(finchPullMounts).Should(gomega.Equal(nerdctlPullMounts))
+		})
+	})
+
+	ginkgo.Describe("Soci", func() {
```
ginglis13, line 36: nit: duplicate Describe

e2e/vm/vm_test.go
```diff
@@ -34,6 +36,11 @@ func TestVM(t *testing.T) {
 		t.Fatal(err)
 	}
 
+	limactlO, limaHomePath, err := e2e.CreateLimaOption()
+	if err != nil {
+		t.Fatal(err)
+	}
```
ginglis13, line 42: let's move this into `soci_test.go` as a BeforeEach (https://onsi.github.io/ginkgo/#extracting-common-setup-beforeeach). then you can simplify the `testSoci` function signature, too

pkg/config/config.go
```diff
@@ -35,6 +35,8 @@ type AdditionalDirectory struct {
 type Finch struct {
 	CPUs   *int    `yaml:"cpus"`
 	Memory *string `yaml:"memory"`
+	//Soci: boolean value of whether user wants SOCI installed
+	Soci *bool `yaml:"soci_snapshotter,omitempty"`
```
ginglis13, line 39: what do you think about making this `Snapshotter` or `ContainerdSnapshotter` ? This could leave the door open for adding support for more snapshotters in the future

pkg/config/lima_config_applier.go
```diff
@@ -96,6 +121,8 @@ func (lca *limaConfigApplier) Apply(isInit bool) error {
 		limaCfg = *cfgAfterInit
 	}
 
+	toggleSoci(&limaCfg, *lca.cfg.Snapshotter == "soci", sociVersion, system.NewStdLib().Arch())
```
ginglis13, line 124: could `*lca.cfg.Snapshotter` be `nil`?

e2e/vm/soci_test.go
```diff
@@ -0,0 +1,76 @@
...
+	ginkgo.Describe("SOCI", func() {
+		var limactlO *option.Option
+		var limaHomePathEnv string
+		var wd string
+
+		ginkgo.BeforeEach(func() {
+			wd, _ = os.Getwd()
+			limaHomePathEnv = "LIMA_HOME=" + filepath.Join(wd, "../../_output/lima/data")
+			limactlO, _ = option.New([]string{filepath.Join(wd, "../../_output/lima/bin/limactl")},
+				option.Env([]string{limaHomePathEnv}))
```
ginglis13, line 34: let's not discard these errors just in case. You can assert they are `nil` eg
```go
wd, err = os.Getwd()
gomega.Expect(err).Should(BeNil())
```

pkg/config/lima_config_applier_test.go
```diff
@@ -85,8 +86,118 @@ fi
...
-		},
-		{
+		}, {
+			name: "adds soci script when soci is set to true in config",
+			config: &Finch{
+				Memory:  pointer.String("2GiB"),
+				CPUs:    pointer.Int(4),
+				VMType:  pointer.String("qemu"),
+				Rosetta: pointer.Bool(false),
+				Soci:    pointer.Bool(true),
```
ginglis13, line 96:
```suggestion
				Snapshotter:    pointer.String("soci"),
```

pkg/config/config.go
```diff
@@ -35,6 +35,8 @@ type AdditionalDirectory struct {
 type Finch struct {
 	CPUs   *int    `yaml:"cpus"`
 	Memory *string `yaml:"memory"`
+	// Soci: the snapshotter that will be installed and configured automatically on vm init or on vm start
```
ginglis13, line 38: nit: update godoc

pkg/config/lima_config_applier.go
```diff
@@ -16,7 +16,32 @@ import (
 	"github.com/runfinch/finch/pkg/system"
 )
 
-const userModeEmulationProvisioningScriptHeader = "# cross-arch tools"
+const (
+	sociVersion                              = "0.3.0"
```
ginglis13, line 20: maybe not in this PR but we should think about / track automatically updating this version in CI somewhere

CodeChanning: created issue #526 for this comment

e2e/vm/soci_test.go
```diff
@@ -0,0 +1,79 @@
...
+			nerdctlPullMounts := countMounts(limactlO)
+			command.New(o, "rmi", "-f", FfmpegSociImage).WithTimeoutInSeconds(30).Run()
+			gomega.Expect(finchPullMounts).Should(gomega.Equal(nerdctlPullMounts))
+		})
+	})
+}
+
+// counts the mounts present in the VM after pulling an image.
+func countMounts(limactlO *option.Option) int {
+	mountOutput := string(command.New(limactlO, "shell", "finch", "mount").WithTimeoutInSeconds(30).Run().Out.Contents())
```
ginglis13, line 77:
```suggestion
	mountOutput := command.StdoutStr(limactlO, "shell", "finch", "mount").WithTimeoutInSeconds(30))
```

https://github.com/runfinch/common-tests/blob/58d5e20765a09a00a86049b48f909115626e49c0/command/command.go#L147-L148

e2e/vm/soci_test.go
```diff
@@ -0,0 +1,79 @@
...
+		ginkgo.BeforeEach(func() {
+			wd, err = os.Getwd()
+			gomega.Expect(err).Should(gomega.BeNil())
+			limaHomePathEnv = "LIMA_HOME=" + filepath.Join(wd, "../../_output/lima/data")
+			limactlO, err = option.New([]string{filepath.Join(wd, "../../_output/lima/bin/limactl")},
+				option.Env([]string{limaHomePathEnv}))
+			gomega.Expect(err).Should(gomega.BeNil())
+		})
+
+		ginkgo.It("finch pull should have same mounts as nerdctl pull with SOCI", func() {
```
ginglis13, line 39: nit for both tests: use `command.Run()` instead of `command.New().Run()` where applicable

ginglis13 (approved):
LGTM, a few nits and CI failure tracked in #527

e2e/vm/soci_test.go
```diff
@@ -0,0 +1,79 @@
...
+
+	"github.com/onsi/ginkgo/v2"
+	"github.com/onsi/gomega"
+	"github.com/onsi/gomega/gexec"
+	"github.com/runfinch/common-tests/command"
+	"github.com/runfinch/common-tests/option"
+)
+
+const (
+	FfmpegSociImage = "public.ecr.aws/soci-workshop-examples/ffmpeg:latest"
```
ginglis13, line 19:
```suggestion
	ffmpegSociImage = "public.ecr.aws/soci-workshop-examples/ffmpeg:latest"
```

It doesn't look like this is being used in other packages so we don't need to export it.

e2e/vm/soci_test.go
```diff
@@ -0,0 +1,79 @@
...
+		})
+
+		ginkgo.It("finch pull should have same mounts as nerdctl pull with SOCI", func() {
+			resetVM(o, installed)
+			resetDisks(o, installed)
+			writeFile(finchConfigFilePath, []byte("cpus: 6\nmemory: 4GiB\nsnapshotter: soci\n"+
+				"vmType: qemu\nrosetta: false"))
+			initCmdSession := command.New(o, virtualMachineRootCmd, "init").WithTimeoutInSeconds(600).Run()
+			gomega.Expect(initCmdSession).Should(gexec.Exit(0))
+			command.New(o, "pull", FfmpegSociImage).WithTimeoutInSeconds(30).Run()
```
ginglis13, line 46:
```suggestion
			command.Run(o, virtualMachineRootCmd, "init").WithTimeoutInSeconds(600)
			command.Run(o, "pull", FfmpegSociImage).WithTimeoutInSeconds(30)
```

Run will ensure exit code is 0: https://pkg.go.dev/github.com/runfinch/common-tests@v0.7.1/command#Run

e2e/vm/soci_test.go
```diff
@@ -0,0 +1,79 @@
...
+		})
+
+		ginkgo.It("finch run should have same mounts as nerdctl run with SOCI", func() {
+			resetVM(o, installed)
+			resetDisks(o, installed)
+			writeFile(finchConfigFilePath, []byte("cpus: 6\nmemory: 4GiB\nsnapshotter: soci\n"+
+				"vmType: qemu\nrosetta: false"))
+			initCmdSession := command.New(o, virtualMachineRootCmd, "init").WithTimeoutInSeconds(600).Run()
+			gomega.Expect(initCmdSession).Should(gexec.Exit(0))
+			command.New(o, "run", FfmpegSociImage).WithTimeoutInSeconds(30).Run()
```
ginglis13, line 63:
```suggestion
			command.Run(o, virtualMachineRootCmd, "init").WithTimeoutInSeconds(600)
			command.Run(o, "pull", FfmpegSociImage).WithTimeoutInSeconds(30)
```

same as above

e2e/vm/soci_test.go
```diff
@@ -0,0 +1,79 @@
...
+			finchPullMounts := countMounts(limactlO)
+			command.Run(o, "rmi", "-f", FfmpegSociImage)
+			command.New(limactlO, "shell", "finch",
+				"sudo", "nerdctl", "--snapshotter=soci", "run", FfmpegSociImage).WithTimeoutInSeconds(30).Run()
```
ginglis13, line 67:
```suggestion
			command.Run(limactlO, "shell", "finch",
				"sudo", "nerdctl", "--snapshotter=soci", "run", FfmpegSociImage).WithTimeoutInSeconds(30)
```

e2e/vm/soci_test.go
```diff
@@ -0,0 +1,79 @@
...
+			resetDisks(o, installed)
+			writeFile(finchConfigFilePath, []byte("cpus: 6\nmemory: 4GiB\nsnapshotter: soci\n"+
+				"vmType: qemu\nrosetta: false"))
+			initCmdSession := command.New(o, virtualMachineRootCmd, "init").WithTimeoutInSeconds(600).Run()
+			gomega.Expect(initCmdSession).Should(gexec.Exit(0))
+			command.New(o, "pull", FfmpegSociImage).WithTimeoutInSeconds(30).Run()
+			finchPullMounts := countMounts(limactlO)
+			command.Run(o, "rmi", "-f", FfmpegSociImage)
+			command.New(limactlO, "shell", "finch",
+				"sudo", "nerdctl", "--snapshotter=soci", "pull", FfmpegSociImage).WithTimeoutInSeconds(30).Run()
```
ginglis13, line 50:
```suggestion
			command.Run(limactlO, "shell", "finch",
				"sudo", "nerdctl", "--snapshotter=soci", "pull", FfmpegSociImage).WithTimeoutInSeconds(30)
```

ginglis13 (changes requested):

e2e/vm/soci_test.go
CodeChanning: I don't think I'm able to set a timeout for the command without using .New() instead of just .Run()

ginglis13 (dismissed):
LGTM 🚀

pkg/config/lima_config_applier.go
```diff
@@ -16,7 +16,32 @@ import (
...
+	sociInstallationProvisioningScriptHeader = "# soci installation and configuring"
+	sociFileNameFormat                       = "soci-snapshotter-%s-linux-%s.tar.gz"
+	sociDownloadURLFormat                    = "https://github.com/awslabs/soci-snapshotter/releases/download/v%s/%s"
+	sociInstallationScriptFormat             = `%s
+if [ ! -f /usr/local/bin/soci ]; then
+	#download soci
+	curl -OL "%s"
```
pendo324, line 27: Should we retry the cURL download? Should the script exit early if this step fails instead of trying to reconfigure containerd?

README.md
```diff
@@ -99,6 +99,15 @@ An example `finch.yaml` looks like this:
...
 memory: 4GiB
+# snapshotter: the snapshotter a user want to use as there default snapshotter
+# Supported Snapshotters List:
+# - soci https://github.com/awslabs/soci-snapshotter/tree/main
+# Once the option has been set the snapshotter will be installed on either finch vm init or finch vm start.
+# The snapshotter binary will be downloaded on the virtual machine and will be configured and ready for use.
+# To change your default snpahotter back to overlayfs, simply remove the snapshotter value from finch.yaml
+# To completely remove the snapshotter binary, shell into your VM and remove /usr/local/bin/{snapshotter binary}
+# and remove the snapshotter configuration in the containerd config file found at /etc/containerd/config.toml
+snapshotter: soci
```
pendo324, line 110: If contianerd supports multiple snapshotters being enabled at once, this should probably be a slice

pendo324 (commented):
Looks good overall, left a few comments.

README.md
CodeChanning: That makes sense. My only question with that is how would the user specify what they want their default snapshotter to be? Should that be a separate config option to set the default snapshotter?

pkg/config/lima_config_applier.go
```diff
@@ -96,6 +121,15 @@ func (lca *limaConfigApplier) Apply(isInit bool) error {
 		limaCfg = *cfgAfterInit
 	}
 
+	var sociEnabled bool
+	if lca.cfg.Snapshotter == nil {
+		sociEnabled = false
+	} else {
+		sociEnabled = (*lca.cfg.Snapshotter == "soci")
+	}
+
```
sam-berning, line 130: we should add an error here if `snapshotter` doesn't have a valid value so users don't wonder why their config change isn't working

pkg/config/lima_config_applier.go
```diff
@@ -188,3 +222,37 @@ func hasUserModeEmulationInstallationScript(limaCfg *limayaml.LimaYAML) (int, bo
...
+		})
+	} else if hasScript && !enabled {
+		limaCfg.Env = map[string]string{"CONTAINERD_SNAPSHOTTER": ""}
+		if len(limaCfg.Provision) > 0 {
+			limaCfg.Provision = append(limaCfg.Provision[:idx], limaCfg.Provision[idx+1:]...)
+		}
+	}
+}
+
+func hasSociInstallationScript(limaCfg *limayaml.LimaYAML) (int, bool) {
```
sam-berning, line 245: nit: I think given the returns, a name like `findSociInstallationScript` works better here

README.md
ginglis13: We could have the default be the first snapshotter listed in this configuration. And then clearly document that. Additionally, users can specify snapshotter at the CLI on a per-command basis e.g. `finch run --snapshotter=<x> ...` - which again could be documented.

sam-berning (commented):
LGTM overall, just a few minor comments

e2e/vm/soci_test.go
```diff
@@ -0,0 +1,76 @@
...
+			nerdctlPullMounts := countMounts(limactlO)
+			command.Run(o, "rmi", "-f", ffmpegSociImage)
+			gomega.Expect(finchPullMounts).Should(gomega.Equal(nerdctlPullMounts))
+		})
+	})
+}
+
+// counts the mounts present in the VM after pulling an image.
+func countMounts(limactlO *option.Option) int {
+	mountOutput := command.StdoutStr(limactlO, "shell", "finch", "mount")
```
ningziwen, line 74: Great job! Love the e2e tests!

ningziwen (dismissed):

pendo324 (dismissed):
Thanks, LGTM 👍

pkg/config/config.go
```diff
@@ -35,6 +35,9 @@ type AdditionalDirectory struct {
 type Finch struct {
 	CPUs   *int    `yaml:"cpus"`
 	Memory *string `yaml:"memory"`
+	// Snapshotter: the snapshotter that will be installed and configured automatically on vm init or on vm start
```
ginglis13, line 38:
```suggestion
	// Snapshotters: the snapshotters that will be installed and configured automatically on vm init or on vm start.
```

pkg/config/lima_config_applier.go
```diff
@@ -188,3 +235,74 @@ func hasUserModeEmulationInstallationScript(limaCfg *limayaml.LimaYAML) (int, bo
 
 	return scriptIdx, hasCrossArchToolInstallationScript
 }
+
+// initializes the bool slice for what snapshotter the user has enabled to all false
+// this will be changed later depending on the user's snapshotters config values.
+func initializeEnabledSnapshotterSlice(numSupportedSnapshotters int) [2][2]bool {
+	var enabledSnapshotters [2][2]bool
```
ginglis13, line 242: I don't think we need this function. Go will set uninitialized values to the type's "zero" value, which for a `bool` is `false`. e.g.

```
var b [2][2]bool
fmt.Printf("%+v\n", b)

[[false false] [false false]]
```

ginglis13 (approved):

sam-berning (approved):
LGTM, great work!

Issue #492, opened 2023-07-21
Support Finch on Windows


pendo324:
**What is the problem you're trying to solve?.**
Windows users should be able to use Finch. I provided some thoughts on how this could be implemented below, but I'm happy to take comments on this issue directly to influence the design before the implementation is finalized.

**Describe the feature you'd like**
A user experience consistent with Finch on macOS, but for container developers that prefer Windows.

**Additional context**
Similar to #377, but for Windows instead of Linux.

Regarding implementation details, I think the ideal way to proceed is to add support for Windows to our dependencies, namely, Lima. This way, most of the work could be reused by other projects that depend on Lima, and the Finch changes would be minimal, and mostly around packaging and distribution.

Based on my research so far, it seems like WSL2 is the best way to implement this in Lima, via a new VM driver (similar to how there's a QEMU driver and a VZ driver in Lima already). WSL2 seems to be a long-term, Microsoft supported feature of Windows, so we don't have to worry about maintaining the Virtualization stack ourselves. However, this also means that we can't really contribute to bug fixes and adding new features to WSL2. Technically, there are two main points against WSL2: disk I/O speed and hardware support.

Disk I/O performance is limited because of the way files are shared between the main Windows filesystem and the Linux filesystem (plan 9 fs server). It's likely that the existing performance is good enough for most user's use cases though. It's possible that Microsoft addresses this concern over time.

Hardware support for WSL2 is limited to CPUs/Motherboards that have VT-x/VT-d / AMD-V / SVM support. Other options, like HyperV also have these hardware limitations. QEMU technically does not have this limitation as it can fall back to software virtualization, but this is not really practical for any sort of "good" user experience. Besides the hardware support, certain Windows features need to be enabled, but this can be automated by Finch or Lima directly and is not as much of an issue (will just require more careful error handling / messaging).

WSL2 VMs initialize from root filesystem archives instead of disk images (qcow2/img/iso), so we will also need to adjust our OS building process to support WSL2, and will need to adjust Lima to support these new root filesystem archives when using the WSL2 driver.

estesp:
Can we close this issue?

pendo324:
Closed by #778, and installer available in [v1.1.1](https://github.com/runfinch/finch/releases/tag/v1.1.1)

PR #507, opened 2023-07-28
ci: add script to update rootfs URL

ginglis13:
Issue #, if available: #492

*Description of changes:*

for Finch on Windows #492, we need to provide a rootfs to WSL2. This rootfs lives in our dependencies bucket, and will be updated from time-to-time for security patches, bug fixes, etc. This commit will automatically pull the most recent rootfs from the dependencies bucket as part of the update deps action.

TODO: similar change to finch-core.

*Testing done:*

locally run script with `FINCH_ROOTFS_URL ?= https://deps.runfinch.com/finch-rootfs-production-amd64-000000000.tar.zst` set in Makefile, observe update.

#### License Acceptance

### DIFF
--- a/.github/workflows/update-deps.yaml
+++ b/.github/workflows/update-deps.yaml
@@ -29,6 +29,7 @@ jobs:
       - name: update dependencies url
         run: |
           ./.github/bin/update-lima-dep.sh -d ${{ secrets.DEPENDENCY_BUCKET_NAME }} -a {{ secrets.ARTIFACT_BUCKET_NAME }}
+          ./.github/bin/update-rootfs.sh -d ${{ secrets.DEPENDENCY_BUCKET_NAME }}
 
       - name: create PR
         uses: peter-evans/create-pull-request@v5
--- a/.github/bin/update-rootfs.sh
+++ b/.github/bin/update-rootfs.sh
@@ -1,0 +1,26 @@
+#!/bin/sh
+set -euxo pipefail
+
+DEPENDENCY_CLOUDFRONT_URL="https://deps.runfinch.com/"
+AARCH64_FILENAME_PATTERN="finch-rootfs-production-arm64-[0-9].*\.tar.zst$"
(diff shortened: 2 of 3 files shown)


pendo324 (approved):
LGTM

.github/workflows/update-deps.yaml
```diff
@@ -29,6 +29,7 @@ jobs:
       - name: update dependencies url
         run: |
           ./.github/bin/update-lima-dep.sh -d ${{ secrets.DEPENDENCY_BUCKET_NAME }} -a {{ secrets.ARTIFACT_BUCKET_NAME }}
+          ./.github/bin/update-rootfs.sh -d ${{ secrets.DEPENDENCY_BUCKET_NAME }}
```
KevinLiAWS, line 32: Can we have a different bucket/sub-folder for the new artifact?

ginglis13: sure, or how about under respective ${ARCH} folder as lima/qemu is today? 
https://github.com/runfinch/finch/blob/81d8b20d0253016bcfd49f72d763ec3aaa93b7af/Makefile#L38

KevinLiAWS: I think a folder at the same level with aarch64 and amd64 maybe better so that each type of the artifact won't be stored in the same folder which will be cleaner and reduce the risk for artifact remove case for a dir.

KevinLiAWS: By the way, I think we may only need this change in the Core repo

ginglis13: > By the way, I think we may only need this change in the Core repo

Yea thanks for pointing that out... There are a couple places where we can deduplicate code / actions between finch and finch-core. Opened https://github.com/runfinch/finch/issues/508 to track

ginglis13:
Closing in favor of https://github.com/runfinch/finch-core/pull/131 and #511

PR #510, opened 2023-07-31
fix: ecr creds helper and credential_process

pendo324:
Issue #, if available: #656

*Description of changes:*
  - invoke `aws configure export-credentials --format process` before running any commands that could need credentials (build, pull, push)
  - Pass the results into the VM in the as environment variables if the command returns output
    - Not every user may have the AWS CLI on the correct version to have the `export-credentials --format process command`, falls back to previous behavior in this case

TODO: add tests for new function

*Testing done:*
  - local testing

#### License Acceptance


go.mod
```diff
@@ -50,6 +50,7 @@ require (
 )
 
 require (
+	github.com/aws/aws-sdk-go v1.44.312
```
austinvazquez, line 53: Should we go ahead and use `aws-sdk-go-v2`?

pendo324: Probably doesn't matter much either way, since the struct is only used to hold values. But why not, if its also available in v2: https://github.com/aws/aws-sdk-go-v2/blob/main/aws/credentials.go#L74-L93

I'll put out a revision shortly

austinvazquez (approved):
LGTM

PR #524, opened 2023-08-08
chore: remove FINCH_DOCKER_COMPAT

### LINKED ISSUE #418: Nested volume mounting behaviour is not compatible with Docker
**Describe the bug**
There appears to be a behavior mismatch between Finch and Docker in the handling of nested volume mounts. In Docker, a nested volume within another volume is correctly accessible, while in finch, the same nested volume fails to be accessible and returns an error "No such file or directory".

**Steps to reproduce**
```
finch run --rm -it -v /tmp/test-1/test-2:/test-1/test-2 -v /tmp/test-empty:/test-1 public.ecr.aws/amazonlinux/amazonlinux:2 bash
```

In bash,
```
ls /test-1/test-2
```
The 'ls' command results in the following error:
```
ls: cannot access /test-1/test-2: No such file or directory
```

**Expected behavior**

In Docker, it can return the correct content under test-2. The compatibility with docker is expected.

**Additional context**
Nerdctl issue: https://github.com/containerd/nerdctl/issues/2254

_To help debug the issue as quickly as possible, we recommend generating a support bundle with `finch support-bundle-generate` and attaching it to this issue. This packages all Finch-related configs and logs into one file._

ginglis13:
Issue #, if available:

Closes #418 

*Description of changes:*

with release of nerdctl v1.5 we can remove the FINCH_DOCKER_COMPAT environment variable used as a workaround for nested volume mount behavior.

*Testing done:*

#### License Acceptance

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -139,60 +139,6 @@ jobs:
           git clean -f -d
           REGISTRY=${{ steps.vars.outputs.has_creds == true && env.REGISTRY || '' }} make test-e2e
         shell: zsh {0}
-  e2e-tests-for-docker-compat:
-    strategy:
-      fail-fast: false
-      matrix:
-        os:
-          [
-            [self-hosted, macos, amd64, 13, test],
-            [self-hosted, macos, arm64, 13, test],
-          ]
-    runs-on: ${{ matrix.os }}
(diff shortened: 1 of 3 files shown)


pendo324 (approved):
Thanks for cleaning this up

Issue #297, opened 2023-03-16
Automatically (or semi-automatically) map TMPDIR into VM


rix0rrr:
**What is the problem you're trying to solve?.**

CDK will sometimes put directories with Dockerfiles into a user's temporary directory, and ask finch to build them. Right now, since the user's temporary directory isn't mounted into the finch VM, they get a confusing error like:

```
time="2023-03-16T09:43:07Z" level=fatal msg="lstat /private/var/folders/zx/d5wln9n10sn0tcj1v9798f1c0000gr/T/jsii-kernel-9VYgrO/node_modules/@aws-cdk/aws-lambda-python-alpha/lib/Containerfile: no such file or directory"
```

It's confusing on many levels:

- There is indeed no `Containerfile`, but there is a `Dockerfile`
- It's not obvious to a user that this path is inside the finch VM, instead of their own file system

**Describe the feature you'd like**

Right now we're telling users to put this into their `finch.yaml`:

```yaml
additional_directories:
  - path: /var/folders
  - path: /private/var/folders
```

It would be nice if finch could do this automatically.

If not, I wonder if there are performance implications of mapping everything under `/var/folders`. If so, could finch maybe support a symbolic name for the temporary directory instead?

```yaml
additional_directories:
  - path: TMPDIR
```

And either map that automatically always, or give the user the ability to add the above.

**Additional context**

Our issue: https://github.com/aws/aws-cdk/issues/24458

rix0rrr:
Slight complication in the fact that:

```
$TMPDIR contains /var/folders/zx/d5wln9n10sn0tcj1v9798f1c0000gr/T/
```

But actually

```
/var -> /private/var
```

And the actual directory (that should probably be mapped) is:

```
/private/var/folders/zx/d5wln9n10sn0tcj1v9798f1c0000gr/T/
```

But now, unless the VM also has the `/var -> /private/var` symlink, it matters whether we use the unresolved `$TMPDIR` or the resolved `$TMPDIR`, because they will contain different path names and only one of them will be mapped...

vsiravar:
Released in[ v0.8.0](https://github.com/runfinch/finch/releases/tag/v0.8.0)

Issue #387, opened 2023-04-28
Finch could not be used as a docker replacement for Dev Containers in VSCode


normalerweise:
**Describe the bug**
Briefly describe the problem you are having.

**Steps to reproduce**
Follow: https://code.visualstudio.com/docs/devcontainers/tutorial 
However after installing Dev Containers extension, go to the extensions settings and replace Docker and Docker Compose Path settings with the corresponding finch path (see screenshot)

**Expected behavior**
Visual Studio Code Dev Containers successfully starts a dev container using finch

**Screenshots or logs**
![image](https://user-images.githubusercontent.com/1102742/235253516-f38c9d36-5561-4e18-976b-27e08cf71afd.png)

![image](https://user-images.githubusercontent.com/1102742/235253260-ab22b2b5-038d-4ab9-9434-613e275ab0e5.png)

**Additional context**
Finch seems to not support the --format parameter for the version command

_To help debug the issue as quickly as possible, we recommend generating a support bundle with `finch support-bundle-generate` and attaching it to this issue. This packages all Finch-related configs and logs into one file._

ningziwen:
The issue to track `--format`. https://github.com/runfinch/finch/issues/199

weikequ:
Was able to do a patch to get past the version --format check, but getting some subsequent errors. Anyone have any idea?

```sh
[31 ms] Dev Containers 0.295.0 in VS Code 1.79.2 (695af097c7bd098fbf017ce3ac85e09bbc5dda06).
[30 ms] Start: Resolving Remote
[38 ms] Start: Check Docker is running
[39 ms] Start: Run: /Users/weikequ/weikequ/finch/_output/bin/finch version --format {{.Server.APIVersion}}
[202 ms] Stop (163 ms): Run: /Users/weikequ/weikequ/finch/_output/bin/finch version --format {{.Server.APIVersion}}
[202 ms] Server API version: Client:
 Version:       fef6e77.modified
 OS/Arch:       linux/arm64
 GitCommit:     fef6e772c28429ef14068702c1dfd8e9fec2dea1
 nerdctl:
  Version:      v1.4.0
  GitCommit:    7e8114a82da342cdbec9a518c5c6a1cce58105e9
 buildctl:
  Version:      v0.11.6
  GitCommit:    2951a28cd7085eb18979b1f710678623d94ed578

Server:
 containerd:
  Version:      v1.7.1
  GitCommit:    1677a17964311325ed1c31e2c0a3589ce6d5c30d
[...]

PR #525, opened 2023-08-09
fix: Mount /var/folders to finch vm

vsiravar:
Issue #, if available:
https://github.com/runfinch/finch/issues/297

*Description of changes:*
Mount /var/folders into finch vm automatically to improve compatibility with docker. 

Docker mount directories
image

*Testing done:*
Yes

#### License Acceptance

### DIFF
--- a/finch.yaml
+++ b/finch.yaml
@@ -76,6 +76,11 @@ mounts:
   # 🟢 Builtin default: false
   # 🔵 This file: true (only for "/tmp/lima")
   writable: true
+- location: "/var/folders"
+  writable: true
+# Because readlink /var => private/var
+- location: "/private/var/folders"
+  writable: true
 
 # Lima disks to attach to the instance. The disks will be accessible from inside the
 # instance, labeled by name. (e.g. if the disk is named "data", it will be labeled


weikequ (approved):
Thanks for this! I think this is a great idea as it might help with getting eventual vscode integration as well. See issue: https://github.com/runfinch/finch/issues/387. Is there any security concerns with respect to the vm/containers having access to these temp folders? 

Edit: Seems like there are some e2e tests, mainly: 

`Run a container image [BeforeEach] when running a container with volume flags`

vsiravar:
> Thanks for this! I think this is a great idea as it might help with getting eventual vscode integration as well. See issue: https://github.com/runfinch/finch/issues/387. Is there any security concerns with respect to the vm/containers having access to these temp folders?

Thanks for the additional context about vscode integration, it would very cool to have finch integrate with vscode. 

Regarding security concerns, there is not a lot of Apple Documentation about `/var/folders` but I did find some useful posts
- /var/folders stores temporary files and caches and it's **ok** to delete everything in it except the folder itself which is not possible to do from finch since the folder belongs to root. https://apple.stackexchange.com/questions/176371/can-i-delete-files-or-folders-from-private-var-folders
- lima enforces that finch [can't be run as root](https://github.com/lima-vm/lima/blob/c5594366a3800f7e5405c73ad03919f70b07bb5f/cmd/limactl/main.go#L87) so a container can't modify files that don't belong to the user that launched the container from the host

ginglis13 (approved):

vsiravar:
> Thanks for this change. I know `/var/folders` was chosen specifically to address #297, but are there any concerns with adding other directories listed in the screenshot as well?

Apart from `/tmp` which would hide the system /tmp in the vm all others can be mounted. https://github.com/lima-vm/lima/issues/571. To bypass mounting `/tmp` from macOS we can mount `/private/tmp` since  `readlink /tmp` =>  
`private/tmp`. Maybe can add this to a subsequent PR.

Issue #538, opened 2023-08-18
Finch does not apply `iptables` rules to containers running in a bridge network


sam-berning:
**Describe the bug**
If I start a container in a bridge network, `iptables` rules that I have applied to that network will not be applied as expected.

**Steps to reproduce**
```
$ finch network create test-network
$ finch run -d --network test-network --name nginx-test nginx:latest
$ finch inspect nginx-test --format '{{json .NetworkSettings.IPAddress}}'
$ LIMA_HOME=/Applications/Finch/lima/data /Applications/Finch/lima/bin/limactl shell finch
[sam@lima-finch finch]$ sudo cat /etc/cni/net.d/nerdctl-test-network.conflist
# fetch bridge name (should be br-1841d116ee60)
[sam@lima-finch finch]$ sudo iptables -t nat -I PREROUTING -p tcp -d <some-ip> --dport 80 -j DNAT --to-destination <nginx-ip>:80 -i br-1841d116ee60
[sam@lima-finch finch]$ exit
$ finch run -it --network test-network amazonlinux:2
# this works!
bash-4.2# curl <nginx-ip>                                                                                        
<!DOCTYPE html>
<html>
<head>
<title>Welcome to nginx!</title>
<style>
html { color-scheme: light dark; }
body { width: 35em; margin: 0 auto;
font-family: Tahoma, Verdana, Arial, sans-serif; }
</style>
</head>
<body>
<h1>Welcome to nginx!</h1>
<p>If you see this page, the nginx web server is successfully installed and
working. Further configuration is required.</p>

<p>For online documentation and support please refer to
<a href="http://nginx.org/">nginx.org</a>.<br/>
Commercial support is available at
<a href="http://nginx.com/">nginx.com</a>.</p>

<p><em>Thank you for using nginx.</em></p>
</body>
</html>

# this will hang indefinitely
bash-4.2# curl <forwarded-ip>
```

**Expected behavior**
I should be able to `curl`  the `<forwarded-ip>` and see the nginx output.

**Screenshots or logs**
N/A

**Additional context**
N/A

_To help debug the issue as quickly as possible, we recommend generating a support bundle with `finch support-bundle generate` and attaching it to this issue. This packages all Finch-related configs and logs into one file._

mharwani:
Issue is fixed. Closing

PR #539, opened 2023-08-18
fix: enables bridge-nf-call-iptables by default

sam-berning:
Issue #, if available: https://github.com/runfinch/finch/issues/538

*Description of changes:*

Enables `net.bridge.bridge-nf-call-iptables` by default in `sysctl`. This will [send packets in a bridge network to iptables for processing](https://wiki.libvirt.org/Net.bridge.bridge-nf-call_and_sysctl.conf.html).

This will also fix the warning on `finch info`:
```
WARNING: bridge-nf-call-iptables is disabled
WARNING: bridge-nf-call-ip6tables is disabled
```

*Testing done:*

1. `finch info`
1. Inspected `/etc/sysctl.d/99-lima.conf` to ensure all of the expected fields are there
1. Reproduced https://github.com/runfinch/finch/issues/538 against my local Finch, and it succeeded.

**To discuss:** is it worth creating a new sysctl config file instead of appending to `/etc/sysctl.d/99-lima.conf`?

#### License Acceptance

### DIFF
--- a/finch.yaml
+++ b/finch.yaml
@@ -143,6 +143,15 @@ provision:
 - mode: boot
   script: |
     modprobe virtiofs
+- mode: boot
+  script: |
+    modprobe br_netfilter
+    cat <<EOF > /etc/sysctl.d/99-finch.conf
+    net.bridge.bridge-nf-call-iptables = 1
+    net.bridge.bridge-nf-call-ip6tables = 1
+    net.ipv4.ip_forward = 1
+    EOF
+    sysctl --system
 # # `user` is executed without the root privilege
 - mode: user
   script: |


pendo324 (approved):
Looks good, I do wonder if this type of thing should be added directly to Lima though. We don't really need a detection that `br_netfilter` is loaded (with modprobe), but Lima would.

finch.yaml
```diff
@@ -143,6 +143,15 @@ provision:
 - mode: boot
   script: |
     modprobe virtiofs
+- mode: system
+  script: |
+    cat <<EOF >> /etc/sysctl.d/99-lima.conf
```
mharwani, line 148: Does this run only on `vm init` or `vm start`? Wondering if changes will be appended to the file. I also noticed that `modprobe br_netfilter` would automatically set `bridge-nf-call-iptables` to 1.

weikequ:
> Looks good, I do wonder if this type of thing should be added directly to Lima though. We don't really need a detection that br_netfilter is loaded (with modprobe), but Lima would.

Any reason why it's disabled by default? There's no issues on lima nor is there any references to this in the codebase except for a [k8s example](https://github.com/lima-vm/lima/blob/e788c08785b860c3428e18d1ce4536c1ec95c123/examples/k8s.yaml#L49) that is almost the same as what we've written here. Good job on figuring this out!

weikequ (approved):

finch.yaml
sam-berning: good catch, it does append to the file on every `vm start`. it doesn't seem to affect the functionality but the file would end up getting pretty long 😅 

> I also noticed that modprobe br_netfilter would automatically set bridge-nf-call-iptables to 1.

yeah, confirmed this works for me too. good catch! will follow up with a new revision

sam-berning:
> Looks good, I do wonder if this type of thing should be added directly to Lima though. We don't really need a detection that `br_netfilter` is loaded (with modprobe), but Lima would.

Yeah, I think that's a good call. Will open a PR to Lima too

pendo324:
> > Looks good, I do wonder if this type of thing should be added directly to Lima though. We don't really need a detection that `br_netfilter` is loaded (with modprobe), but Lima would.
> 
> Yeah, I think that's a good call. Will open a PR to Lima too

This is probably where I'd put it in Lima
https://github.com/lima-vm/lima/blob/master/pkg/cidata/cidata.TEMPLATE.d/boot/00-modprobe.sh

sam-berning:
> Any reason why it's disabled by default? There's no issues on lima nor is there any references to this

According to [this page](https://wiki.libvirt.org/Net.bridge.bridge-nf-call_and_sysctl.conf.html): it's disabled in Fedora by default because it can cause traffic to virtual machines to be blocked if the host's iptables rules do not account for the guest. It also notes that it's enabled in the Linux kernel settings by default, and only some distributions have it disabled -- I think this is likely why Lima doesn't have any logic to enable it.

pendo324:
> > Looks good, I do wonder if this type of thing should be added directly to Lima though. We don't really need a detection that `br_netfilter` is loaded (with modprobe), but Lima would.
> 
> Yeah, I think that's a good call. Will open a PR to Lima too

You might want to ask on an issue or on Slack why its not in the list already though. It seems like a weird thing to not include, unless they left it out on purpose.

mharwani (dismissed):

pendo324 (approved):

sam-berning:
@mharwani I realized that though the `br_netfilter` kernel module enables these configs by default, it's probably safer long term to explicitly configure them vs. relying on the defaults, so I added `/etc/sysctl.d/99-finch.conf`. I think it probably won't matter but this approach is ever so slightly safer

mharwani (dismissed):

mharwani (approved):

monirul (approved):
LGTM.

PR #549, opened 2023-08-29
feat: supports adding files inside the VM to support bundles

sam-berning:
Issue #, if available:

*Description of changes:*

Allows users to add files from inside their finch VM to a support bundle. The file can be specified in a config or in a command line flag with `vm:/path/to/file/in/vm`

*Testing done:*

```
make test-unit
```

manual testing

#### License Acceptance

### DIFF
--- a/pkg/support/support.go
+++ b/pkg/support/support.go
@@ -6,6 +6,7 @@ package support
 
 import (
 	"archive/zip"
+	"bufio"
 	"bytes"
 	"errors"
 	"fmt"
@@ -51,6 +52,7 @@ type bundleBuilder struct {
 	config BundleConfig
 	finch  fpath.Finch
 	ecc    command.Creator
+	lcc    command.LimaCmdCreator
 	lima   wrapper.LimaWrapper
 }
 
@@ -61,6 +63,7 @@ func NewBundleBuilder(
 	config BundleConfig,
 	finch fpath.Finch,
 	ecc command.Creator,
+	lcc command.LimaCmdCreator,
 	lima wrapper.LimaWrapper,
 ) BundleBuilder {
 	return &bundleBuilder{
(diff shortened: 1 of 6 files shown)


pkg/support/support.go
```diff
@@ -108,7 +111,13 @@ func (bb *bundleBuilder) GenerateSupportBundle(additionalFiles []string, exclude
...
 			continue
 		}
-		err := bb.copyInFile(writer, file, path.Join(zipPrefix, logPrefix))
+		bb.logger.Debugf("Copying %s...", file)
+		var err error
+		if isFileFromVM(file) {
+			err = bb.streamFileFromVM(writer, file, path.Join(zipPrefix, logPrefix))
+		} else {
+			err = bb.copyInFile(writer, file, path.Join(zipPrefix, logPrefix))
+		}
```
ginglis13, line 120: nitpicky but could this be wrapped in a small util function, since this logic is repeated 2 other times?

ginglis13 (commented):
LGTM, should we also add something to the helptext for `finch support-bundle generate` to document the use case of `vm:<file>`  ?

ginglis13 (dismissed):
LGTM. Nonblocking IMO, but do you think this change warrants e2e tests as well?

sam-berning:
> LGTM. Nonblocking IMO, but do you think this change warrants e2e tests as well?

Yeah, I'll make a follow up PR

pkg/support/support.go
```diff
@@ -196,12 +214,50 @@ func (bb *bundleBuilder) copyInFile(writer *zip.Writer, fileName string, prefix
...
+	err := cmd.Start()
+	if err != nil {
+		return err
+	}
+
+	waitStatus := make(chan error)
+	go func() {
+		err := cmd.Wait()
+		_ = pipeWriter.Close()
+		waitStatus <- err
```
mharwani, line 244: Does the error returned from `cmd.Wait()` display the correct error message, like "file not found"? Maybe we should read `errBuf` to generate the error message.

mharwani (approved):

ginglis13 (approved):

PR #564, opened 2023-09-07
docs: add Commits section to CONTRIBUTING.md

ginglis13:
Issue #, if available:

*Description of changes:*

Add a "Commits" section to contributing guidelines with information about conventional commits. Additionally, add guidance around providing subtypes to conventional commit messages, especially when writing commits that are specific to one platform or another.

*Testing done:*

#### License Acceptance

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -14,8 +14,9 @@
     - [Build](#build)
     - [Unit Testing](#unit-testing)
     - [E2E Testing](#e2e-testing)
-  - [Sign Your Commits](#sign-your-commits)
-    - [DCO](#dco)
+  - [Commits](#commits)
+    - [Conventional Commit Messages](#conventional-commit-messages)
+    - [Sign Your Commits with DCO](#sign-your-commits-with-dco)
   - [Pull Request Checklist](#pull-request-checklist)
     - [Testing](#testing)
       - [Unit Testing - Parallel by Default](#unit-testing---parallel-by-default)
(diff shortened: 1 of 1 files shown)


weikequ (approved):
Looks good. I know we have a PR title check for conventional commits + a DCO check as well. Do you think we need to mention those?

vsiravar (dismissed):
Thanks! Looks good.

ginglis13:
> Looks good. I know we have a PR title check for conventional commits + a DCO check as well. Do you think we need to mention those?

I think this should be sufficient - DCO is already mentioned in contributing guidelines. As for PR title check, if the PR author has followed conventional commits in their commits, then the pre-populated PR title will be the commit message and pass those checks.

pendo324 (dismissed):
LGTM. Do you think we should follow up by adding some custom rules to [release-please.yaml](../../tree/main/.github/workflows/release-please.yaml#L25) to make different platforms have their own sections in the release notes?

ginglis13:
> Do you think we should follow up by adding some custom rules to [release-please.yaml](../../tree/main/.github/workflows/release-please.yaml#L25) to make different platforms have their own sections in the release notes?

I considered that, but `scope`s are not included in the parsing spec https://github.com/commitizen/conventional-commit-types/blob/master/index.spec.json . Ideally we could define our rules as something like the below but it is not supported :/

```yaml
changelog-types: >
            [
              {
                "type":"feat",
                "scope":"", 
                "section":"Features",
                "hidden":false
              },
              {
                "type":"feat",
                "scope":"macOS",
                "section":"Features",
                "hidden":false
              },
              {
                "type":"feat",
                "scope":"Windows",
                "section":"Features",
                "hidden":false
              },
              ...
             ]
```

(also I updated to change `<subtype>` -> `<scope>` in our docs in latest rev)

As is without new custom rules, a commit `feat(windows): the feature` would appear in release notes like:

```text
Features

* windows: the feature
* some other feature w/o a scope tag
```

pendo324:
> > Do you think we should follow up by adding some custom rules to [release-please.yaml](../../tree/main/.github/workflows/release-please.yaml#L25) to make different platforms have their own sections in the release notes?
> 
> I considered that, but `scope`s are not included in the parsing spec https://github.com/commitizen/conventional-commit-types/blob/master/index.spec.json . Ideally we could define our rules as something like the below but it is not supported :/
> 
> ```yaml
> changelog-types: >
>             [
>               {
>                 "type":"feat",
>                 "scope":"", 
>                 "section":"Features",
>                 "hidden":false
>               },
>               {
>                 "type":"feat",
>                 "scope":"macOS",
>                 "section":"Features",
>                 "hidden":false
>               },
>               {
>                 "type":"feat",
>                 "scope":"Windows",
>                 "section":"Features",
>                 "hidden":false
>               },
>               ...
>              ]
> ```
> 
> (also I updated to change `<subtype>` -> `<scope>` in our docs in latest rev)
> 
> As is without new custom rules, a commit `feat(windows): the feature` would appear in release notes like:
> 
> ```
> Features
> 
> * windows: the feature
> * some other feature w/o a scope tag
> ```

Cool, thanks for the info!

pendo324 (approved):

vsiravar (approved):

PR #581, opened 2023-09-25
feat: add handling for nerdctl commands and options

vsiravar:
Issue #, if available:

*Description of changes:*
- introduces aliases for nerdctl commands. Eg `nerdctl run` is equivalent to `nerdctl container run`. cli commands are restructured based on https://www.docker.com/blog/whats-new-in-docker-1-13/ , `nerdctl container run` is the recommended way of using the cli. This also helps in handling options for commands and avoids duplication for having to handle the same options for equivalent commands such as `run` and `container run`. 
- introduces command handlers. command handlers are functions that handle arguments of a specific command. For example `cpHandler`, handles `nerdctl cp` command. It has a one to one mapping with nerdctl commands that need command handling. 
- introduces argHandlers, which handle particular options in nerdctl commands. These are not tied to a specific command. For instance `handleFilePath` function can handle `-f/--file` option in `nerdctl build`. It can also handle `--label-file` in `nerdctl run` command. The mapping is defined by `argHandlerMap`. 

*Testing done:*
Yes, unit tests added for cmdhandler and argHandlers. all container e2e tests and passed.

----- 
#### Notes 
There were several options considered for translating windows filepaths to wsl filepaths for nerdctl commands. 

* add logic [handleEnv](https://github.com/runfinch/finch/blob/main/cmd/finch/nerdctl.go#L229) function to handle options. This approach had several disadvantages
    *  options require different handling depending on the command. For instance -f/--file option in build command has different semantics to -f in rm command. 
    *  Not extensible if new options are added to nerdctl command.
* search and replace. match every argument in a nerdctl to a windows filepath. If a match is found then replace it with it's equivalent wsl filepath. This approach also has a couple of problems. 

    * There is no perfect regex to match windows paths in nerdctl commands. Windows paths can have spaces and can also be relative paths and when used in conjunction with nerdctl commands there is ambiguity. Eg nerdctl cp C:\\Test USER conname:<containerpath>. Do we match C:\\Test USER or C:\\Test USER conname since both are valid windows paths. 
    * Matching every argument adds to the  computation cost of parsing nerdctl commands.
    
Therefore the choice was made to have a mapping between commands and options with the path translation logic which can guarantee correctness in handling nerdctl arguments(if implemented correctly :) ) and is extensible.

For macOS, the handers are no-op. We can refactor [handling logic ](https://github.com/runfinch/finch/blob/main/cmd/finch/nerdctl.go#L229) in a later PR. 

-------
  
#### License Acceptance

### DIFF
--- a/cmd/finch/nerdctl_windows.go
+++ b/cmd/finch/nerdctl_windows.go
@@ -26,24 +26,24 @@ func convertToWSLPath(systemDeps NerdctlCommandSystemDeps, winPath string) (stri
 }
 
 // substitues wsl path for the provided option in place for nerdctl args
-func handleFilePath(systemDeps NerdctlCommandSystemDeps, args []string, index int) error {
-	var prefix = args[index]
+func handleFilePath(systemDeps NerdctlCommandSystemDeps, nerdctlCmdArgs []string, index int) error {
+	var prefix = nerdctlCmdArgs[index]
 
 	// If --filename="<filepath> then we need to cut <filepath> and convert that to wsl path
-	if strings.Contains(args[index], "=") {
+	if strings.Contains(nerdctlCmdArgs[index], "=") {
 		before, after, _ := strings.Cut(prefix, "=")
 		wslPath, err := convertToWSLPath(systemDeps, after)
 		if err != nil {
 			return err
 		}
-		args[index] = fmt.Sprintf("%s=%s", before, wslPath)
+		nerdctlCmdArgs[index] = fmt.Sprintf("%s=%s", before, wslPath)
 	} else {
-		if (index + 1) < len(args) {
-			wslPath, err := convertToWSLPath(systemDeps, args[index+1])
+		if (index + 1) < len(nerdctlCmdArgs) {
+			wslPath, err := convertToWSLPath(systemDeps, nerdctlCmdArgs[index+1])
 			if err != nil {
 				return err
 			}
-			args[index+1] = wslPath
+			nerdctlCmdArgs[index+1] = wslPath
 		} else {
 			return fmt.Errorf("invalid positional parameter for %s", prefix)
 		}
@@ -52,16 +52,16 @@ func handleFilePath(systemDeps NerdctlCommandSystemDeps, args []string, index in
 }
 
 // hanldes -v/--volumes option. For anonymous volumes and named volumes this is no-op. For bind mounts path is converted to wsl path.
-func handleVolume(systemDeps NerdctlCommandSystemDeps, args []string, index int) error {
-	var prefix = args[index]
+func handleVolume(systemDeps NerdctlCommandSystemDeps, nerdctlCmdArgs []string, index int) error {
+	var prefix = nerdctlCmdArgs[index]
 	var v = ""
 	var found = false
 	var before = ""
-	if strings.Contains(args[index], "=") {
+	if strings.Contains(nerdctlCmdArgs[index], "=") {
 		before, v, found = strings.Cut(prefix, "=")
 	} else {
-		if (index + 1) < len(args) {
-			v = args[index+1]
+		if (index + 1) < len(nerdctlCmdArgs) {
+			v = nerdctlCmdArgs[index+1]
 		} else {
 			return fmt.Errorf("invalid positional parameter for %s", prefix)
 		}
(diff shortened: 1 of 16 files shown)


cmd/finch/nerdctl_windows.go
```diff
@@ -107,21 +106,250 @@ func handleVolume(systemDeps NerdctlCommandSystemDeps, args []string, index int)
...
+		if (index + 1) < len(args) {
+			v = args[index+1]
+		} else {
+			return fmt.Errorf("invalid positional parameter for %s", prefix)
+		}
+	}
+
+	// https://docs.docker.com/storage/bind-mounts/#choose-the--v-or---mount-flag  order does not matter, so convert to a map
+	entries := strings.Split(v, ",")
+	m := make(map[string]string)
```
pendo324, line 127: Any thoughts on using a struct for this instead of a map? nerdctl / buildkit / docker-cli code might have something we can just import? It just might make it clearer if we can enumerate all of the options on a mount and check if they're empty, instead of just checking if a map has a key

cmd/finch/nerdctl_windows.go
```diff
@@ -107,21 +106,250 @@ func handleVolume(systemDeps NerdctlCommandSystemDeps, args []string, index int)
...
+	}
+
+	return nil
+}
+
+// cp command handler, takes command arguments and converts hostpath to wsl path in place. It ignores all other arguments
+func cpHandler(systemDeps NerdctlCommandSystemDeps, cmdArgs []string) error {
+	for i, arg := range cmdArgs {
+		// -L and --follow-symlink don't have to be processed
+		if arg == "-" || strings.HasPrefix(arg, "-") {
```
pendo324, line 268:
```suggestion
		if strings.HasPrefix(arg, "-") {
```
This should be equivalent

cmd/finch/nerdctl_windows.go
```diff
@@ -107,21 +106,250 @@ func handleVolume(systemDeps NerdctlCommandSystemDeps, args []string, index int)
...
+			cmdArgs[i] = wslPath
+		}
+	}
+	return nil
+}
+
+// this is the handler for image build command. It translates build context to wsl path.
+func imageBuildHandler(systemDeps NerdctlCommandSystemDeps, cmdArgs []string) error {
+	var err error
+	if cmdArgs[len(cmdArgs)-1] != "--debug" {
```
pendo324, line 291: Can only the first option be `--debug`? Do we have to search arguments instead of assuming order, or is this a safe assumption?

cmd/finch/nerdctl_windows.go
```diff
@@ -107,21 +106,250 @@ func handleVolume(systemDeps NerdctlCommandSystemDeps, args []string, index int)
...
 		"-v":           handleVolume,
 		"--volume":     handleVolume,
+		"--mount":      handleBindMounts,
 	},
+	"compose": {
+		"--file": handleFilePath,
+	},
+}
+
+var commandHanderMap = map[string]commandHandler{
```
pendo324, line 343: Are there any 3 word mappings?

pendo324 (changes requested):
Overall LGTM. I left some minor comments.

Can you also either rebase and sign all commits or squash and sign the new commit and any further commits going forward? Need DCO to pass for the rest of the CI to run.

cmd/finch/nerdctl_windows.go
vsiravar: It does not matter, if the last argument is `--debug`, then build context is the one before that, if not the last argument is buildcontext. `--debug` will be skipped from processing either way. It's mainly to check if --debug is the last argument while translating buildcontext.

cmd/finch/nerdctl_windows.go
vsiravar: There are no commands which have 2 subcommands that require filepath handling. The only nerdctl command that has 2 subcommands is [nerdctl-ipfs-registry-serve](https://github.com/containerd/nerdctl/blob/main/docs/command-reference.md#nerd_face-nerdctl-ipfs-registry-serve) but no filepath handling in it. In the future if there were 3 subcommands that required handling, we'll have to add them as commands to https://github.com/runfinch/finch/blob/0e2dbf9d936795cc373802357200a52a001a82a4/cmd/finch/nerdctl.go#L374.

pendo324 (approved):

pendo324:
Closing/re-opening to get the CI to run

PR #594, opened 2023-10-03
feat: windows persistent disk and other misc fixes/refactors

pendo324:
Issue #, if available:

*Description of changes:*
- Adds persistent disk support to Windows
  - This took more work than anticipated, in order to deal with non-Admin users
  - `pkg/disk/dpgo/` is brand new code, and should be a focus of the review
  - `pkg/winutil/run_windows.go` is also new and requires careful review. This is what allows Finch to run as the regular user, except for when it needs Admin access to call `diskpart` (to create the persistent disk)
- Fix paths in `nerdctl_config_applier` to make the post-boot/init shelling work
- Added winres to allow the finch.exe to have metadata attached to it. This is WIP, need final icons and descriptions etc.
- Large (in terms of lines changed) refactor of `pkg/path/finch.go`, but it should have no impact on functionality (needs careful review)
- Fixed the Makefile's `clean` target for Windows
- Most of the other changes are just noise from refactoring (like, literally renaming things). Nothing major, but take a look if possible. Sorry the diff is so large 

*Testing done:*

#### License Acceptance


pkg/disk/dpgo/disk_windows.go
```diff
@@ -0,0 +1,128 @@
+//go:build windows
```
vsiravar, line 1: nit: missing license.

cmd/finch/virtual_machine_init.go
```diff
@@ -86,6 +86,9 @@ func (iva *initVMAction) run() error {
 		return err
 	}
 
+	// ignore error, this is to ensure that the disk is only mounted once
```
ginglis13, line 89: might need `//nolint:errcheck` here (and line 90)

e: and a few other places in this PR.

pkg/config/lima_config_applier.go
```diff
@@ -201,3 +213,36 @@ func findSociInstallationScript(limaCfg *limayaml.LimaYAML) (int, bool) {
...
+	for idx, prov := range limaCfg.Provision {
+		trimmed := strings.Trim(prov.Script, " ")
+		if !hasWslDiskFormatScript && strings.HasPrefix(trimmed, wslDiskFormatScriptHeader) {
+			hasWslDiskFormatScript = true
+			scriptIdx = idx
+			break
+		}
+	}
+
+	return scriptIdx, hasWslDiskFormatScript
```
ginglis13, line 247: nit: scriptIdx not used

ginglis13 (commented):
just a few nits on first pass

pkg/disk/dpgo/disk_windows.go
```diff
@@ -0,0 +1,128 @@
...
+		fmt.Fprintf(dpStdin, "%s\r\n", tmpl.String())
+		dpStdin.Close()
+	}()
+
+	cd.logger.Debugln("waiting for diskpart to exit")
+	if err := cmd.Wait(); err != nil {
+		return err
+	}
+
+	cd.logger.Infof("stdout: %s", dpStdout.String())
```
vsiravar, line 125: nit: 
```suggestion
	cd.logger.Infof("create disk cmd stdout: %s", dpStdout.String())
```

pkg/disk/disk_windows.go
```diff
@@ -5,8 +5,169 @@
...
+		if err := m.fs.MkdirAll(disksDir, 0o755); err != nil {
+			return false, fmt.Errorf("could not create persistent disk directory: %w", err)
+		}
+	}
+
+	return true, nil
+}
+
+//go:embed createDiskAdmin.TEMPLATE.ps1
+var createDiskTmpl string
```
vsiravar, line 81: Is this used?

vsiravar:
Really cool PR, I left a few minor comments from the first pass, especially some files missing licenses.

pkg/disk/dpgo/disk_windows.go
pendo324: Good catch, I'm just going to make this a debug log and do some parsing to make sure it's created correctly.

pkg/disk/disk_windows.go
pendo324: Not anymore, since the logic is moved to another package. I'll remove it 👍

cmd/finch/virtual_machine_init.go
pendo324: I think assigning it to `_` works; I don't see it in the linter output. Is there any reason why `//nolint:errcheck` would be better?

pkg/disk/dpgo/disk_windows.go
pendo324: Added log parsing / verification in latest revision

pkg/disk/disk_windows.go
pendo324: Removed in latest revision

pkg/config/lima_config_applier.go
pendo324: Fixed in latest revision

pkg/disk/dpgo/disk_windows.go
pendo324: Fixed in latest revision

pkg/disk/disk_windows.go
```diff
@@ -5,8 +5,209 @@
...
+	if errors.Is(err, fs.ErrNotExist) {
+		if err := m.fs.MkdirAll(disksDir, 0o755); err != nil {
+			return false, fmt.Errorf("could not create persistent disk directory: %w", err)
+		}
+	}
+
+	return true, nil
+}
+
+type createDiskOpts struct {
```
vsiravar, line 82: Is this leftover?

pendo324: Yeah, just removed it, thanks

finch.yaml
```diff
@@ -90,8 +90,6 @@ mounts:
 # `/mnt/lima-${VOLUME}`.
 # 🟢 Builtin default: null
 # For Finch, this value should always be the same as the diskName in pkg/disk/disk.go
-additionalDisks:
-- "finch"
```
vsiravar, line 94: Is this not required anymore?

ginglis13: +1, maybe a comment here directing readers to additionalDisks being set in lima_config_applier?

pkg/disk/disk_windows.go
```diff
@@ -5,8 +5,203 @@
...
+	"strings"
+
+	"github.com/spf13/afero"
+
+	"github.com/runfinch/finch/pkg/flog"
+	"github.com/runfinch/finch/pkg/winutil"
+)
+
 // EnsureUserDataDisk checks the current disk configuration and fixes it if needed.
 // on Windows, this is a no-op.
```
ginglis13, line 25: nit: update comment, no longer no-op

pkg/disk/disk_windows.go
```diff
@@ -5,8 +5,203 @@
...
+
+// DetachUserDataDisk unmounts the disk in wsl.
+func (m *userDataDiskManager) DetachUserDataDisk() error {
+	cmd := m.ecc.Create(
+		"wsl.exe",
+		"--unmount",
+		`\\?\`+m.finch.UserDataDiskPath(m.rootDir),
+	)
+
+	m.logger.Debugf("running attach cmd: %v", cmd)
```
ginglis13, line 60:
```suggestion
	m.logger.Debugf("running detach cmd: %v", cmd)
```

pkg/disk/disk_windows.go
```diff
@@ -5,8 +5,203 @@
...
+
+	out, err := cmd.CombinedOutput()
+	if err != nil {
+		return fmt.Errorf("failed to attach disk: %w, command output: %s", err, out)
```
ginglis13, line 64:
```suggestion
		return fmt.Errorf("failed to detach disk: %w, command output: %s", err, out)
```

finch.yaml
pendo324: Yeah, this is now handled in `lima_config_applier`. The idea is to move any platform/config specific options to the `config_applier`s, so that we only have one "base" YAML file.

The problem with a comment is that it will be persisted to the output finch.yaml, and that might be confusing. I can add it if you really think it'll help though.

pkg/disk/disk_windows.go
pendo324: Fixed in latest revision

pkg/disk/disk_windows.go
pendo324: Fixed in latest revision

pkg/disk/disk_windows.go
pendo324: Fixed in latest revision

pkg/winutil/run_windows.go
```diff
@@ -0,0 +1,197 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to convert %q to UTF16Ptr: %w", verb, err)
+	}
+	exePtr, err := syscall.UTF16PtrFromString(cmdPath)
+	if err != nil {
+		return fmt.Errorf("failed to convert %q to UTF16Ptr: %w", cmdPath, err)
+	}
+	wdPtr, err := syscall.UTF16PtrFromString(wd)
+	if err != nil {
+		return fmt.Errorf("failed to convert %q to UTF16Ptr: %w", verb, err)
```
vsiravar, line 89:
```suggestion
		return fmt.Errorf("failed to convert %q to UTF16Ptr: %w", wd, err)
```

pkg/winutil/run_windows.go
```diff
@@ -0,0 +1,197 @@
...
+	ret, _, _ := procShellExecuteEx.Call(uintptr(unsafe.Pointer(pExecInfo)))
+	if ret == 1 && pExecInfo.fMask&SEE_MASK_NOCLOSEPROCESS != 0 {
+		s, e := syscall.WaitForSingleObject(syscall.Handle(pExecInfo.hProcess), syscall.INFINITE)
+		switch s {
+		case syscall.WAIT_OBJECT_0:
+			break
+		case syscall.WAIT_FAILED:
+			return os.NewSyscallError("WaitForSingleObject", e)
+		default:
+			return fmt.Errorf("Unexpected result from WaitForSingleObject")
```
vsiravar, line 160:
```suggestion
			return errors.New("Unexpected result from WaitForSingleObject")
```

pkg/disk/disk_unix.go
```diff
@@ -13,13 +13,13 @@ import (
 	"io/fs"
 	"path"
 
+	"github.com/docker/go-units"
 	limaStore "github.com/lima-vm/lima/pkg/store"
 )
 
 const (
 	// diskName must always be consistent with the value under additionalDisks in finch.yaml.
```
vsiravar, line 21: nit: maybe we can modify this to keep the name consistent with value in lima_config_applier.go?

ginglis13 (approved):

pkg/winutil/io.go
```diff
@@ -0,0 +1,26 @@
+package winutil
```
vsiravar, line 1: do we need to compile this for non-windows?

vsiravar (approved):
looks great. Left a couple of nits.

pkg/winutil/io.go
pendo324: Not really, but it doesn't hurt I guess? I can add a build directive if you want

pkg/disk/disk_unix.go
pendo324: Not sure what you mean, can you put a code suggestion?

pkg/winutil/run_windows.go
pendo324: Fixed in latest revision

pkg/winutil/run_windows.go
pendo324: I just used `fmt.Errorf` to not need to import `errors`. The behavior should be the same but I can change it if you want

pkg/disk/disk_unix.go
vsiravar: I meant the comment, since finch.yaml does not define additionalDisks and it's changed to https://github.com/runfinch/finch/pull/594/files#diff-0cc05068a0ab38b109e62d43a953ab4a2d8563df608ab52f3a12d0933e1e3a9fR146. 
```suggestion
	// diskName must always be consistent with the value under lima_config_applier.go 
```

pkg/winutil/io.go
vsiravar: I guess it's fine this way since this would not throw any compilation errors on darwin as well.

pkg/disk/disk_unix.go
pendo324: Oh, got it. Fixed in latest revision

vsiravar (approved):

Issue #221, opened 2023-02-14
`finch run -h` does not produce expected result


stephenmsachs:
`finch help run` tells me to run `finch run -h` or `finch run --help` for help. The former is not "help", but 'hostname' and will fail without an argumnet.

weikequ:
Can recreate. Seems that `finch help command` returns the cobra generated help instead of passing it down to nerdctl.

```bash
$ finch help run
Run a command in a new container

Usage:
  finch run [flags]

Flags:
  -h, --help   help for run

Global Flags:
      --debug   running under debug mode

$ finch run --help
Run a command in a new container. Optionally specify "ipfs://" or "ipns://" scheme to pull image from IPFS.

Usage: finch run [flags] IMAGE [COMMAND] [ARG...]

Flags:
      --add-host strings            Add a custom host-to-IP mapping (host:ip)
.....
See also 'finch --help' for the global flags such as '--namespace', '--snapshotter', and '--cgroup-manager'.
```

Running `finch command --help` will always return the correct help call.

pradeepnnv:
I would like to work on this. Can you please assign this to me?

PR #651, opened 2023-10-18
docs: auto-generate docs

pendo324:
Issue #, if available:

*Description of changes:*
- Add `finch gen-docs generate` command, which can be used to generate documentation
  - This actually runs all of the commands, so a finch vm needs to be running
    - This is because many of the commands shell out to nerdctl inside of a Linux VM, so simply leveraging cobra's built-in documentation generation won't work
    - Actually running the commands and then parsing the output into a markdown file is... kinda hacky, so this might break from time to time
- This only works for top-level nerdctl commands right now, because those are the only commands that finch "knows" about. Meaning, while `finch volume create` is a command, with it's own usage information, it won't have a page here. The doc for `finch volume` does reference its subcommands, however.

*Testing done:*
- TODO
  -  [x] unit tests

#### License Acceptance


ningziwen:
Good to have a gen-doc-no-diff ci.
Similar to https://github.com/runfinch/finch/blob/e21ae836d4f73ea07852e3cd1fc6cc16d56e4f32/.github/workflows/ci.yaml#L37

pendo324:
The nature of this function makes it very hard to add more tests than what I've added in the latest commit. It both modifies os.Stdout, AND is recursive. I tried mocking out the scenario where there's a parent command and it does not seem worth the effort since this is not really a critical command anyway.

KevinLiAWS (dismissed):

cmd/finch/gen_docs.go
```diff
@@ -0,0 +1,231 @@
...
+	gd.deps.SetStdout(rescueStdout)
+
+	_ = w.Close()
+	helpText, err := io.ReadAll(r)
+	if err != nil {
+		return fmt.Errorf("error while reading stdout from pipe: %w", err)
+	}
+
+	seeAlso := ""
+	if cmd.HasParent() && cmd.Parent().Runnable() {
```
vsiravar, line 154: nit: can we add an example for this branch in code, i.e when the parent command is runnable.

cmd/finch/gen_docs.go
```diff
@@ -0,0 +1,231 @@
...
+}
+
+func (gd *genDocsAction) run(outDir string) error {
+	return gd.captureHelpOutput(gd.rootCmd, outDir)
+}
+
+func (gd *genDocsAction) captureHelpOutput(cmd *cobra.Command, outDir string) error {
+	for _, c := range cmd.Commands() {
+		if err := gd.captureHelpOutput(c, outDir); err != nil {
+			return err
```
vsiravar, line 106: nit: adding command that resulted in the err would help with debug.
```suggestion
			return fmt.Errorf("error while generating docs for %s: %w", c.CommandPath(), e)
```

pendo324: > nit: adding command that resulted in the err would help with debug.

The reason I didn't wrap that error is because all errors that occur inside the `captureHelpOutput` function are already wrapped, so it would lead to double wrapping. Do you think it's still necessary?

cmd/finch/gen_docs.go
pendo324: Sure, a contrived example is that if you could run `finch images` as well as `finch images delete`, finch images is the parent command and finch images delete is the child command, so both commands would get their own page.

The technical explanation is the a `cobra.Command` is runnable if it has a `Run` or `RunE` method defined.

cmd/finch/gen_docs.go
pendo324: Decided that wrapping the error doesn't really hurt anything, updated in latest commit

cmd/finch/gen_docs.go
pendo324: Added comment in latest commit

vsiravar: Cool thanks for the example.

docs/cmd/finch_help.md
```diff
@@ -0,0 +1,14 @@
+# finch help
```
weikequ, line 1: Does not work properly as per https://github.com/runfinch/finch/issues/221

vsiravar (approved):
Nice!

weikequ (dismissed):
Left some small comments. Thanks for the good work!

docs/cmd/finch_help.md
pendo324: Yeah, `run` is special cased to disable `-h`, not sure the best way to handle this. Might need to change something upstream to note that -h doesn't work on `run`.

https://github.com/containerd/nerdctl/blob/main/cmd/nerdctl/container_run.go#L77-L78

cmd/finch/main_test.go
```diff
@@ -139,7 +139,7 @@ func TestNewApp(t *testing.T) {
 	assert.Equal(t, cmd.SilenceUsage, true)
 	assert.Equal(t, cmd.SilenceErrors, true)
 	// confirm the number of command, comprised of nerdctl commands + finch commands (version, vm, support-bundle)
```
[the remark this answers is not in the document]
pendo324: Good point, I just updated it in the latest revision.

docs/cmd/finch_help.md
weikequ: Oh, got it. Cool, will keep tracking that issue then. Thanks!

weikequ (approved):
🚀

pendo324:
Skipping CI e2e tests take a long time and it's passed previously (these changes are unrelated to e2e tests)

PR #673, opened 2023-10-31
fix: improve creds helper UX

pendo324:
Issue #, if available:

*Description of changes:*
- Make it so aws configure is also run on the run command, since that may invoke a pull
- Clean up code and consider `config.json`'s state when checking if the cred helper is "installed" or not
- TODO: remove need for var overriding to facilitate testing

*Testing done:*
- local testing, e2e tests

#### License Acceptance

### DIFF
--- a/cmd/finch/nerdctl.go
+++ b/cmd/finch/nerdctl.go
@@ -173,7 +173,7 @@ func (nc *nerdctlCommand) run(cmdName string, args []string) error {
 
 	var additionalEnv []string
 	switch cmdName {
-	case "build", "pull", "push":
+	case "build", "pull", "push", "run":
 		ensureRemoteCredentials(nc.fc, nc.ecc, &additionalEnv, nc.logger)
 	}
 
--- a/cmd/finch/nerdctl.go
+++ b/cmd/finch/nerdctl.go
@@ -173,6 +173,14 @@ func (nc *nerdctlCommand) run(cmdName string, args []string) error {
 
 	var additionalEnv []string
 	switch cmdName {
+	case "image":
+		if slices.Contains(args, "build") || slices.Contains(args, "pull") || slices.Contains(args, "push") {
+			ensureRemoteCredentials(nc.fc, nc.ecc, &additionalEnv, nc.logger)
+		}
+	case "container":
+		if slices.Contains(args, "run") {
+			ensureRemoteCredentials(nc.fc, nc.ecc, &additionalEnv, nc.logger)
+		}
 	case "build", "pull", "push", "run":
 		ensureRemoteCredentials(nc.fc, nc.ecc, &additionalEnv, nc.logger)
 	}
--- a/pkg/dependency/credhelper/cred_helper_binary.go
+++ b/pkg/dependency/credhelper/cred_helper_binary.go
@@ -82,7 +82,7 @@ func updateConfigFile(bin *credhelperbin) error {
 		}
 		credsStore := cfg.CredentialsStore
 		defer fileRead.Close() //nolint:errcheck // closing the file
-		if strings.Compare(credsStore, binCfgName) != 0 {
+		if credsStore != binCfgName {
 			file, err := bin.fs.OpenFile(cfgPath, os.O_RDWR|os.O_CREATE|os.O_TRUNC, 0o755)
 			if err != nil {
 				return err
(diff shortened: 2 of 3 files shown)


pkg/dependency/credhelper/cred_helper_binary.go
```diff
@@ -103,6 +103,38 @@ func updateConfigFile(bin *credhelperbin) error {
 	return nil
 }
 
+func (bin *credhelperbin) binaryInstalled() (bool, error) {
+	return binaryInstalled(bin)
+}
+
+func (bin *credhelperbin) configFileInstalled() (bool, error) {
+	cfgPath := fmt.Sprintf("%s%s", bin.hcfg.finchPath, "config.json")
```
austinvazquez, line 111: Does finch path include the OS separator already?

pkg/dependency/credhelper/cred_helper_binary.go
```diff
@@ -103,6 +103,38 @@ func updateConfigFile(bin *credhelperbin) error {
...
+	if err != nil {
+		return false, err
+	}
+	var cfg configfile.ConfigFile
+	err = json.Unmarshal(bytes, &cfg)
+	if err != nil {
+		return false, err
+	}
+	credsStore := cfg.CredentialsStore
+	defer fileRead.Close() //nolint:errcheck // closing the file
```
austinvazquez, line 134: There are two exit points before the defer close. Should this be moved up?

pkg/dependency/credhelper/cred_helper_binary.go
pendo324: Yes

pkg/dependency/credhelper/cred_helper_binary.go
pendo324: Good catch, fixed in latest revision

cmd/finch/nerdctl.go
```diff
@@ -173,7 +173,7 @@ func (nc *nerdctlCommand) run(cmdName string, args []string) error {
 
 	var additionalEnv []string
 	switch cmdName {
-	case "build", "pull", "push":
+	case "build", "pull", "push", "run":
```
vsiravar, line 176: We should also handle `image build`, `image pull`, `image push` and `container run` as they are the new command syntax for the above case

pkg/dependency/credhelper/cred_helper_binary.go
vsiravar: maybe it's simpler to use `filepath.Join` to keep it consistent with how we construct filepaths

pkg/dependency/credhelper/cred_helper_binary.go
```diff
@@ -117,79 +149,121 @@ func (bin *credhelperbin) fullInstallPath() string {
...
+			"--url",
+			bin.hcfg.credHelperURL,
+			"--output",
+			bin.fullInstallPath(),
+		)
+
+		if _, err = curlCmd.Output(); err != nil {
+			return fmt.Errorf("error installing binary %s, err: %w", bin.hcfg.binaryName, err)
+		}
+		if err := bin.fs.Chmod(bin.fullInstallPath(), 0o755); err != nil {
```
vsiravar, line 205: nit: `0o755` -> `0o700`

pkg/dependency/credhelper/cred_helper_binary.go
```diff
@@ -117,79 +149,121 @@ func (bin *credhelperbin) fullInstallPath() string {
...
+	if !binInstalled {
+		if bin.helper == "" {
+			return nil
+		}
+		if bin.helper != bin.credHelperConfigName() {
+			return nil
+		}
+		credHelperName := strings.ReplaceAll(bin.credHelperConfigName(), "-login", "")
+		bin.l.Infof("Installing %s credential helper", credHelperName)
+		mkdirCmd := bin.cmdCreator.Create("mkdir", "-p", bin.hcfg.installFolder)
```
vsiravar, line 185: nit: Maybe its easier to `bin.fs.MkdirAll(bin.hcfg.installFolder, 0o700)` since bin already has the fs dependency.

pkg/dependency/credhelper/cred_helper_binary.go
pendo324: thanks, fixed in latest revision

pkg/dependency/credhelper/cred_helper_binary.go
pendo324: yep, fixed in latest revision

cmd/finch/nerdctl.go
pendo324: good point, added the subcommands in latest revision. I won't create a system to more eloquently match these subcommands as I know some of that work has been done in the windev branch already

vsiravar (approved):

Issue #537, opened 2023-08-17
`make clean` retains downloads in `deps/finch-core`


ginglis13:
**Describe the bug**
`make clean` does not clean submodule downloads in `deps/finch-core/downloads`

**Steps to reproduce**
```sh
make

ls deps/finch-core/downloads/os

make clean

ls deps/finch-core/downloads/os
```

**Expected behavior**
I would expect `make clean` to remove the downloaded OSs as each are > 500mb, or there to at least be a rule such as `make deps-clean` to remove these. Alternatively, if the idea behind *not* cleaning these is so that they don't need to keep being redownloaded, the rule could have some regex and/or could be modified by the [`Sync Submodules and Dependencies`](https://github.com/runfinch/finch/actions/workflows/sync-submodules-and-deps.yaml) action to only delete OS/rootfs that don't match the latest available.

haytok:
Hi, @ginglis13 
I would like to work on this issue, so could you please assign me ?

PR #685, opened 2023-11-05
fix: Fix to remove the virtual machine storage in deps/finch-core/dow…

haytok:
…nloads/os by make clean

Issue #, if available: #537

*Description of changes:*

In finch, when booting a virtual machine using Lima, it is necessary to download the virtual machine storage in QCOW2 format.

The QCOW2 virtual machine storage thus downloaded and located in deps/finch-core/downloads/os is not deleted when make clean is run in the home directory as currently implemented.

This file may be larger than 500 MB in size and is reported in issue/#537 such that it is expected to be deleted when make clean is run.

Therefore, this fix will remove the QCOW2 format downloaded files in deps/finch-core/downloads/os when make clean is run in the home directory.

*Testing done:* yes

---------

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -327,6 +327,7 @@ clean:
 	-sudo rm -rf "/private/etc/sudoers.d/finch-lima"
 	-@rm -rf $(OUTDIR) 2>/dev/null || true
 	-@rm -rf ./deps/finch-core/_output || true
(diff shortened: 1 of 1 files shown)


pendo324 (approved):
LGTM, thanks!

Issue #687, opened 2023-11-07
After uninstalling from script and `brew install finch` the output says something about "upgrade"


stockholmux:
**Describe the bug**
Briefly describe the problem you are having.

I had a weird experience installing finch. After uninstalling, then re-installing with `brew`, I got a message that it was an upgrade making me think the uninstall isn't fully clean.

**Steps to reproduce**
A clear, step-by-step set of instructions to reproduce the bug.

1. I had an older version of Finch (v0.6.2) installed on a machine.
2. I couldn't remember if I installed it via a script or homebrew so I ran `brew uninstall finch` which returned `Error: Cask 'finch' is not installed.`
3. I ran `sudo bash /Applications/Finch/uninstall.sh` which yielded `Application uninstall process finished`
4. I ran `brew install finch` which yielded:
```
installer: Package name is Finch
installer: Upgrading at base path /
installer: The upgrade was successful.
 finch was successfully installed!
 ```
"Upgrading" and "upgrade was successful" seem out-of-place after an unstall.

**Expected behavior**
Description of what you expected to happen.

`brew install finch` should not be able to detect that finch was ever installed on a machine if I run the uninstall script.

**Screenshots or logs**
If applicable, add screenshots or logs to help explain your problem.

**Additional context**

_To help debug the issue as quickly as possible, we recommend generating a support bundle with `finch support-bundle generate` and attaching it to this issue. This packages all Finch-related configs and logs into one file._

PR #688, opened 2023-11-07
fix: Clean up all previous finch version installation registries in postinstall and uninstall

KevinLiAWS:
Issue #, if available:
https://github.com/runfinch/finch/issues/687
Old finch installation registry remains if uninstall script not explicitly called, the mac will treat a new installation with upgrade info message.

*Description of changes:*
Added a full clean up step in the postinstall script to remove all old registries except the current installed version. Also make the uninstall script not only remove the current but also remove all historical registry record.

*Testing done:*
Tested with cloud build: https://github.com/runfinch/finch/actions/runs/6791189430
Also validated locally with pkg installation:
```
// Before install the fixed pkg, it print all finch registries
% pkgutil --pkgs | grep '^org\.Finch'
org.Finch.v0.8.0
org.Finch.pkg-tool
org.Finch.v0.7.0
org.Finch.v0.6.2
org.Finch.main
org.Finch.0.6.0
org.Finch.0.4.1
org.Finch.0.4.0

// After install the fixed pkg, it only has the current registry
% pkgutil --pkgs | grep '^org\.Finch'
org.Finch.fix-unregistry

// After execute uninstall, all the registries are removed
% pkgutil --pkgs | grep '^org\.Finch'
% // empty
```

#### License Acceptance

### DIFF
--- a/installer-builder/darwin/scripts/postinstall
+++ b/installer-builder/darwin/scripts/postinstall
@@ -25,4 +25,11 @@ sudo rm -rf "/opt/finch/"
 sudo rm -rf "/private/var/run/finch-lima"
 sudo rm -rf "/private/etc/sudoers.d/finch-lima"
 
+# Forget previous pkgutil packages starting with org.Finch except for the current version
+echo "Remove historical pkgutil packages..."
+pkgutil --pkgs | grep '^org\.Finch\.' | grep -v '^org\.Finch\.__VERSION__' | while read -r pkg; do
+    echo "Forgetting package $pkg"
+    sudo pkgutil --forget "$pkg"
+done
+
 echo "Post installation process finished."
--- a/installer-builder/darwin/Resources/uninstall.sh
+++ b/installer-builder/darwin/Resources/uninstall.sh
@@ -27,7 +27,7 @@ echo "[1/3] [DONE] Successfully deleted shortcut links"
 echo "Remove historical pkgutil packages..."
 pkgutil --pkgs | grep '^org\.Finch\.' | while read -r pkg; do
     echo "Forgetting package $pkg"
-    sudo pkgutil --forget "$pkg"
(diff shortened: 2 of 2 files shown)


installer-builder/darwin/scripts/postinstall
```diff
@@ -25,4 +25,11 @@ sudo rm -rf "/opt/finch/"
 sudo rm -rf "/private/var/run/finch-lima"
 sudo rm -rf "/private/etc/sudoers.d/finch-lima"
 
+# Forget previous pkgutil packages starting with org.Finch except for the current version
+echo "Remove historical pkgutil packages..."
```
pendo324, line 29: Since this is in the postinstall script, will it remove the version of Finch that was just installed from the registry too? Or does that only get added to the registry after the postinstall script?

pendo324 (commented):
This will work as long as we only allow one version of Finch to be installed per machine (which we do now, so no concerns). Just something to think about if we ever allow changing the install location, or per-user installs.

installer-builder/darwin/scripts/postinstall
KevinLiAWS: grep -v '^org\.Finch\.__VERSION__' excluded the current version, so it will only remove historical versions

pendo324 (approved):

KevinLiAWS:
> This will work as long as we only allow one version of Finch to be installed per machine (which we do now, so no concerns). Just something to think about if we ever allow changing the install location, or per-user installs.

For this concern:
1) pkg forget always need admin permission
2) mac os standard pkg installation is system level, if we support user level installation, pkgutil will be not in the scope, it only maintains system level registry. In other words, user level doesn't installation need to touch it. When we support user installation case, we can add addtional check based on some condition, such as the location software is installed.

PR #703, opened 2023-11-26
fix: Fix to delete ~/.finch when uninstalling finch

haytok:
In the current implementation, uninstalling finch does not delete the ~/.finch directory.

This directory may hold several GB of container data disks, which can consume a lot of space and be left unremoved.

This issue has also been reported in issue/#457 and needs to be fixed.

Therefore, this fix will ensure that when finch is uninstalled, ~/.finch is also deleted to free up disk space for finch.

Issue #, if available: issue/#457

*Description of changes:* Details are described in commit messages.

*Testing done:* Yes

#### License Acceptance

### DIFF
--- a/installer-builder/darwin/Resources/uninstall.sh
+++ b/installer-builder/darwin/Resources/uninstall.sh
@@ -21,7 +21,7 @@ sudo pkill '^limactl'
 
 if [ "$(readlink '/usr/local/bin/finch')" = "/Applications/Finch/bin/finch" ]; then sudo rm /usr/local/bin/finch; fi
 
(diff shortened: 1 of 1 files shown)


weikequ:
Thanks for the contribution! I think since this deals with users' persistent data, it would be better to have it opt in instead when uninstalling, or at least give a prompt for if the user would like to keep their data.

haytok (commented):
Hi, @weikequ

Thanks for the review !!!
I have changed the code in response to your review, so could you please check back when you have time?

KevinLiAWS (approved):

Issue #399, opened 2023-05-14
I get an error when starting the VM.


north-hoshi:
**Describe the bug**
When I run `finch vm start` I get the following error:
```
FATA[0044] failed to update the user's .profile file: failed to read config file: file does not exist
```

**Steps to reproduce**
`finch vm start`

**Expected behavior**
The vm boot completes without any errors.

**Screenshots or logs**
```
INFO[0000] Starting existing Finch virtual machine...
INFO[0044] Finch virtual machine started successfully
FATA[0044] failed to update the user's .profile file: failed to read config file: file does not exist
```

**Additional context**

- My mac

macOS Monterey
ver 12.3.1
Intel Core i7

- How i installed finch

brew install finch --cask

_To help debug the issue as quickly as possible, we recommend generating a support bundle with `finch support-bundle-generate` and attaching it to this issue. This packages all Finch-related configs and logs into one file._

KevinLiAWS:
Hi @north-hoshi, I checked the .profile permission and it is the same with mine. Could you try finch vm stop, finch vm remove, and finch vm init to see if it helps unblock?

vsiravar:
@north-hoshi What is the `USER` env variable set to on your machine? Finch uses this variable to write to `.bashrc` inside the VM and if the user running finch does not have this variable set on his machine or it's different to the user running the finch process, it's possible to see this error. Can you try `export USER=<user running finch>` and they try `finch vm start`. Thanks!

strongishllama:
Hey @vsiravar having the same issue. I can confirm that the `USER` environment variable is set and I also installed finch via Homebrew. Thanks!

![Screenshot 2023-08-16 at 10 32 57 am](https://github.com/runfinch/finch/assets/30204322/cf324fb3-5dab-41e6-8a60-7ae387f0ae99)

vsiravar:
@strongishllama Thanks for trying finch! Is this(68825) the same user as the one running the finch process? Finch [updates the finch user's .bashrc file in the finch vm](https://github.com/runfinch/finch/blob/main/pkg/config/nerdctl_config_applier.go#L82) post vm initialization. If this is the case can you try running finch as 68825 or run 
```
export USER=`whoami`
```

strongishllama:
@vsiravar Looks like the issue might be on my end. I tried it on another laptop and it worked fine. I'll keep digging for a solution on my end, cheers! 😄

neil-greenwood:
[...]

PR #712, opened 2023-11-30
fix: use LimaUser method instead of host user name

w-haibara:
Issue #, if available: #399

*Description of changes:*
In `nerdctl_config_applier.go`, editing `.bashrc` uses a file path based on the host's username. However, if the host's username contains `.` or `@`, lima uses `lima` as the username instead of the host's username. Therefore, if the host's user name contains `.` or `@`, editing `.bashrc` fails.
This PR fixes this so that you can get the actual username from `LimaWrapper.LimaUser()`.

*Testing done:* yes

#### License Acceptance

### DIFF
--- a/cmd/finch/main.go
+++ b/cmd/finch/main.go
@@ -90,6 +90,7 @@ var newApp = func(logger flog.Logger, fp path.Finch, fs afero.Fs, fc *config.Fin
 		fp.QEMUBinDir(),
 		system.NewStdLib(),
 	)
+	lima := wrapper.NewLimaWrapper()
 	supportBundleBuilder := support.NewBundleBuilder(
 		logger,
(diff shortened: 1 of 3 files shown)


w-haibara:
@vsiravar Thanks for review!
I added a test in [nerdctl_config_applier_test.go](https://github.com/runfinch/finch/blob/05cdd112b50d472da59eb8370980a298c09f6e23/pkg/config/nerdctl_config_applier_test.go#L42) when local user is not a valid linux username.

I have also made additional changes. I would appreciate it if you would review again.

vsiravar (approved):
Thanks, the changes look good.

w-haibara:
Thank you for merge it!

PR #759, opened 2024-01-12
ci: make tests more reliable and debuggable

pendo324:
Issue #, if available: The most common case for macOS tests failing seems to be vm init or vm start failing (or, partially failing), see: https://github.com/runfinch/finch/actions/runs/7493200344/job/20398390952, https://github.com/runfinch/finch/actions/runs/7493150724/job/20398225275, https://github.com/runfinch/finch/actions/runs/7493910137/job/20400678411

*Description of changes:*
- Adds a delay between vm state changes. I'm not sure if this will fix the issue, but its worth a try

*Testing done:*

#### License Acceptance


e2e/vm/additional_disk_test.go
```diff
@@ -42,7 +42,9 @@ var testAdditionalDisk = func(o *option.Option, installed bool) {
 			time.Sleep(20 * time.Second)
 
 			command.New(o, virtualMachineRootCmd, "stop").WithTimeoutInSeconds(90).Run()
+			time.Sleep(10 * time.Second)
```
mharwani, line 45: Wouldn't 0.5 to 1 second be enough? Adding a delay of 10 seconds would keep us from catching any issues that come up if user tries to change VM status without long delays.

pendo324: Yeah, that's a good point. I'll bring it down to 1 second to simulate "real user" inputs. Curious to see how this run of tests turns out with 10s though

pendo324:
Weirdly, a lot of failure seem to be because this `init` command exits with a non-0 status. But [in the code](https://github.com/pendo324/finch/blob/delay-vm-state-changes/e2e/vm/vm_test.go#L57), it says it should ignore the exit status. Still investigating... Maybe you have an idea, @mharwani

pendo324:
Seems like all of the tests that use Virtualization.framework take minutes to init new VMs due to the conversion from qcow2 to raw disks. Investigating why it takes so long and possible ways to speed it up.

.github/workflows/ci-docs.yaml
```diff
@@ -58,6 +58,19 @@ jobs:
     runs-on: ${{ matrix.os }} 
     steps:
       - run: echo "Skipping CI for docs & contrib files"
+  windows-e2e-tests:
+      strategy:
+        matrix:
+          os: 
+            [
+              [self-hosted, macos, amd64, 13, test],
```
weikequ, line 66: These must be the same architectural types as the actual ci. I think from there, it looks something like this:
```yaml
      matrix:
        os:
          [
            [self-hosted, windows, amd64, test],
          ]
```

.github/workflows/ci.yaml
```diff
@@ -20,6 +20,9 @@ permissions:
   id-token: write
   contents: write
 
+env:
+  DEBUG: ${{ secrets.ACTIONS_STEP_DEBUG }}
```
weikequ, line 24: Was this just part of testing or do we still need this?

e2e/vm/config_darwin_test.go
```diff
@@ -33,9 +32,10 @@ var testConfig = func(o *option.Option, installed bool) {
 			}
 
 			limaConfigFilePath := resetVM(o, installed)
+			resetDisks(o, installed)
 			writeFile(finchConfigFilePath, []byte("memory: 4GiB\ncpus: 6\nvmType: vz\nrosetta: false"))
-			initCmdSession := command.New(o, virtualMachineRootCmd, "init").WithTimeoutInSeconds(600).Run()
-			gomega.Expect(initCmdSession).Should(gexec.Exit(0))
+			// vm init with VZ set sometimes takes 2 minutes just to convert the disk to raw
+			command.New(o, "vm", "init").WithoutCheckingExitCode().WithTimeoutInSeconds(240).Run()
```
weikequ, line 38: Saw a bit of discussion regarding the exit code stuff, but still don't quite understand, why would we want to ignore exit codes that are non 0?

weikequ (commented):
Thanks for putting all this together, hopefully it leads to less e2e flakiness. I'm still not 100% sure why for some commands we would want to ignore the exit code (for me it only makes sense in the deferred cleanup cases).

.github/workflows/ci.yaml
pendo324: I _think_ `secrets.ACTIONS_STEP_DEBUG` is what GitHub sets when you click the "run with debugging" option ([docs](https://docs.github.com/en/actions/monitoring-and-troubleshooting-workflows/enabling-debug-logging#enabling-step-debug-logging)).

Since our go e2e tests can't (easily) read the secret, I just propagate it to the environment. Forgot to update the change list in the description of the PR to include this

e2e/vm/config_darwin_test.go
pendo324: I think there's a small chance of the error code being non-zero, yet completing before the timeout. In this case, if the functionality that's being tested still works (i.e. all the stuff _after_ the `vm init` happens), I think that's good enough.

The `vm` commands has its own tests where the error codes are not ignored. Happy to discuss further.

.github/workflows/ci-docs.yaml
pendo324: Fixed in latest rev

pendo324:
@weikequ can you take another look at this?

e2e/vm/config_darwin_test.go
weikequ: That's pretty strange, so you're saying like a partially successful vm init can still allow the e2e tests to still succeed? (ie maybe everything is good except something like networking or mounting a certain directory fails?)

weikequ: > The vm commands has its own tests where the error codes are not ignored. Happy to discuss further.

Yeah thats fair, guess it's not crucial for it here since that's not what we're testing. I wonder how much vm failures affect non-vm related tests though.

weikequ (approved):

weikequ:
Oh it also looks like we need to update our required tests to account for the additional matrix options.

PR #776, opened 2024-01-26
ci: update to GH runners

weikequ:
Issue #, if available:

*Description of changes:*

*Testing done:*

#### License Acceptance


weikequ:
Seems like GH macOS runners run as a VM instance instead of bare metal. From some preliminary research, it doesn't seem like Apple supports running nested virtualization on Apple Silicon, which means that this is a no-go for now. 😞

Issue #827, opened 2024-02-21
Unable to use the "--env" option on finch container run


alecthegeek:
**Describe the bug**
 Cannot use `--env` option

**Steps to reproduce**

```sh
finch container run -it --rm --env="E=v" busybox env
```

**Expected behavior**

```
PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
TERM=xterm
HOME=/root
E=v
```

**Screenshots or logs**

Actual result

```
FATA[0000] unknown shorthand flag: 'e' in -e
```

**Additional context**

Note: Updated to Finch 1.1.2 -- error still present

As a test I ran `nerdctl` (installed via Rancher Desktop) and it worked as expected

```sh
$ nerdctl container run -it --rm --env="E=v" busybox env
PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
E=v
TERM=xterm
HOME=/root
```

_To help debug the issue as quickly as possible, we recommend generating a support bundle with `finch support-bundle generate` and attaching it to this issue. This packages all Finch-related configs and logs into one file._
[finch-support-20240222101339.zip](https://github.com/runfinch/finch/files/14366120/finch-support-20240222101339.zip)

mharwani:
Thanks for the issue. The command works properly when you use `finch run`, but fails with `finch container run`:
```
$ finch run -it --rm --env="E=v" busybox env
PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
E=v
TERM=xterm
HOME=/root
```
```
$ finch container run -it --rm --env="E=v" busybox env
FATA[0000] unknown shorthand flag: 'e' in -e            
FATA[0000] exit status 1
```

The `--debug` option shows improper ordering of nerdctl arguments, where `-e` argument is passed before the `run` command:
```
$ finch --debug container run -it --rm --env="E=v" busybox env
DEBU[0000] Creating limactl command: ARGUMENTS: [shell finch sudo nerdctl container -e E=v run -it --rm busybox env], LIMA_HOME: /Applications/Finch/lima/data
...
```

mharwani:
In the following [code](https://github.com/runfinch/finch/blob/v1.1.2/cmd/finch/nerdctl.go#L250-L259):
```go
limaArgs = append(limaArgs, append([]string{nerdctlCmdName}, strings.Fields(cmdName)...)...)

[...]

PR #830, opened 2024-02-24
fix: parse management commands with proper arguments

mharwani:
Issue #, if available: #827

*Description of changes:*
Addresses #827 by parsing management command names into a single command. For eg, `["finch", "container", "run", "alpine", ...]` will be parsed as `["finch", "container run", "alpine", ...]`, so that "run" is passed as a command rather than an argument.

*Testing done:*
Changes tested locally to ensure that the issue is fixed.
```
$ ./_output/bin/finch container run --debug -it --rm --env="E=v" busybox env
DEBU[0000] Creating limactl command: ARGUMENTS: [shell finch sudo -E nerdctl container run -e E=v -it --rm busybox env], LIMA_HOME: /Users/mharwani/work/runfinch/finch/_output/lima/data 
E=v
TERM=xterm
HOME=/root
```

#### License Acceptance

### DIFF
--- a/cmd/finch/nerdctl.go
+++ b/cmd/finch/nerdctl.go
@@ -106,6 +106,9 @@ func (nc *nerdctlCommand) run(cmdName string, args []string) error {
 	if err != nil {
 		return err
 	}
+
+	cmdName, args = parseManagementCmd(cmdName, args)
+
 	var (
 		nerdctlArgs, envs, fileEnvs       []string
 		skip, hasCmdHander, hasArgHandler bool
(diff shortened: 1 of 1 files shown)


cmd/finch/nerdctl.go
```diff
@@ -474,3 +477,43 @@ var nerdctlCmds = map[string]string{
 	"volume":    "Manage volumes",
 	"wait":      "Block until one or more containers stop, then print their exit codes",
 }
+
+var managementCmds = []string{
```
pendo324, line 481: Are management commands the only type of commands that can have sub-arguments?

cmd/finch/nerdctl.go
```diff
@@ -474,3 +477,43 @@ var nerdctlCmds = map[string]string{
...
+	"apparmor",
+	"builder",
+	"container",
+	"image",
+	"ipfs",
+	"namespace",
+	"network",
+	"system",
+	"volume",
```
pendo324, line 490: `apparmor`, `ipfs`, and `namespace` aren't in `nerdctlCmds` at [line 438 in this diff](https://github.com/runfinch/finch/pull/830/files#diff-403cda46333c3f741f8b1ff7f391901b0ccb8a5501cdd241138f89fd3c9db2c6R438).

If we add these, can we also regenerate the docs (run `./_output/bin/finch gen-docs generate -p ./docs/cmd` and commit the results)?

cmd/finch/nerdctl.go
Shubhranshu153: just to make sure i understood correctly
nerdctl compose up
in this case this is a management command as up is not an option, is this what the question about?
Note: compose is not part of this list.

pendo324: The fix in this PR is to change the way Finch handles parsing local env and then passing it into the VM context via the limactl/nerdctl command. Currently, when Finch does this, it takes a command like `finch run -it --env="E=e" alpine:latest` and transforms it into `finch run --env="E=e" -it alpine:latest`, because `run` is identified as the end of the "command".

This PR fixes the parsing logic by adding a new list of commands that have subcommands (the `managementCmds`), so that if you run a command like `finch container run --debug -it --rm --env="E=v" busybox env`, it recognizes `container run` as a single command, and adds the `--env` after `container run`, instead of after `container`.

A better approach may be to entirely rework how environment variables are handled, by not re-arranging where `--env` is in the command, and instead replacing `--env` in place.

So in the example in the PR description,

```
finch container run --debug -it --rm --env="E=v" busybox env
```

becomes

```
shell finch sudo -E nerdctl container run  -it --rm --env="E=v" busybox env
```

instead of

```
shell finch sudo -E nerdctl container run -e E=v -it --rm busybox env
```

Shubhranshu153: sure in that case let me raise a fresh cr and once that checks out we can cancel this one

Issue #747, opened 2024-01-03
finch attach not available as top level command


weikequ:
**Describe the bug**
Finch should make `finch attach` available as a top level command to keep parity with [nerdctl](https://github.com/containerd/nerdctl/blob/main/docs/command-reference.md#whale-nerdctl-attach). It is currently available as a subcommand as `finch container attach`

**Steps to reproduce**
- `finch attach --help`

**Expected behavior**
```bash
$ finch attach --help
Attach stdin, stdout, and stderr to a running container. For example:
```

**Screenshots or logs**
- `FATA[0000] unknown command "attach" for "finch"`

haytok:
Hi, Team 
Could you assign this issue to me ?
I'll address this issue.

haytok:
This feature can be added, but we can't implement the test because of a bug in nerdctl.

- https://github.com/containerd/nerdctl/issues/2877
- https://github.com/containerd/containerd/pull/10020

Further details are discussed at a below PR.

- https://github.com/runfinch/finch/pull/847

PR #847, opened 2024-03-11
fix: add finch attach command

haytok:
The finch attach command cannot be executed at the top level. Therefore, when the finch attach -h command is executed, the following error occurs.

  ```
  > finch attach -h
  FATA[0000] unknown command "attach" for "finch"
  ```

Also, this bug has been reported in issue/747 below, and it is hoped that finch attach will be executed.

- https://github.com/runfinch/finch/issues/747

Meanwhile, with the nerdctl command, it is possible to execute the nerdctl attach command.

Therefore, in this commit, we will make corrections so that the finch attach command can be executed.

Issue #, if available: #747

*Description of changes:* Details are described in commit message.

*Testing done:* No

#### License Acceptance

### DIFF
--- a/cmd/finch/nerdctl.go
+++ b/cmd/finch/nerdctl.go
@@ -473,4 +473,5 @@ var nerdctlCmds = map[string]string{
 	"update":    "Update configuration of one or more containers",
 	"volume":    "Manage volumes",
 	"wait":      "Block until one or more containers stop, then print their exit codes",
+	"attach":    "Attach stdin, stdout, and stderr to a running container",
 }


Shubhranshu153:
Thank you for the commit.
Can we add an e2e test for this. There is a separate package called common-test, where we can add this test. We have some examples in the package
https://github.com/runfinch/common-tests

For more info:
https://github.com/runfinch/finch/blob/main/CONTRIBUTING.md

haytok:
@Shubhranshu153 

Thanks for comments.




I have investigated the following, but am having trouble implementing the test for the `finch attach` command because we cannot use the `-it` option within ginkgo.
Could you please advise on how to implement the test?

### What I have investigated



First, when testing locally, we can check the operation as follows

```bash
haytok finch [add-attach-command]
> _output/bin/finch run --rm --name test -it alpine sh
/ #
/ #
/ # lbin    dev    etc    home   lib    media  mnt    opt    proc   root   run    sbin   srv    sys    tmp    usr    var
/ #
```

Other session

```bash
haytok finch [add-attach-command]
> _output/bin/finch attach test

/ #
/ # s
```



However, as you can see in the following content, it is difficult to implement a test that executes `finch attach` against a container launched interactively in the test.

- https://github.com/runfinch/common-tests/blob/main/tests/exec.go#L29

> // TODO: specifying -t flag will have error in test -> panic: provided file is not a console



So, for example, suppose that the container is launched as follows. ([ref](https://github.com/runfinch/common-tests/blob/main/tests/exec.go#L31C1-L33C6))

```golang
			ginkgo.BeforeEach(func() {
				command.Run(o, "run", "-d", "--name", testContainerName, defaultImage, "sleep", "infinity")
			})
```



Since the shell process is not running in the container launched with sleep infinity, the test for finch attach cannot be executed.

```bash
haytok finch [add-attach-command]
> _output/bin/finch run -d --name test alpine sleep infinity
0b4d9dc034ddb95335fe9685c48b7bdfaeec8d2cdfc5fe7be46fc369ddac682d
haytok finch [add-attach-command]
> _output/bin/finch ps
CONTAINER ID    IMAGE                              COMMAND             CREATED          STATUS    PORTS    NAMES
0b4d9dc034dd    docker.io/library/alpine:latest    "sleep infinity"    4 seconds ago    Up                 test
haytok finch [add-attach-command]
> _output/bin/finch attach test
FATA[0000] failed to attach to the container: failed to open stdout fifo: error creating fifo binary:///usr/local/bin/nerdctl?_NERDCTL_INTERNAL_LOGGING=%2Fvar%2Flib%2Fnerdctl%2F1935db59: no such file or directory
FATA[0000] exit status 1
```

Shubhranshu153:
for your test you might be able to use
```
finch run -d ubuntu bash -c "while true; do echo /'Hello, world/'; sleep 1; done"
```
to lunch a container that is prininting hellow world per second. Then you can able to attach the container and test if it is in stdout. there are some examples for that in Events.go 
https://github.com/runfinch/common-tests/blob/146e16fe020f6ef94139d473b78103e374180647/tests/events.go#L4

haytok:
@Shubhranshu153 

The verification was carried out by starting the container you advised on.
However, due to [a bug on the nerdctl side](https://github.com/containerd/nerdctl/issues/2877), it was not possible to attach it to the container with `finch attach`.

More details were described in the nerdctl discussion below I asked.

- https://github.com/containerd/nerdctl/discussions/2876

Therefore, it seems difficult to implement tests for `finch attach` at the moment.
So, should we put this PR pending for now?

Shubhranshu153:
thanks for the investigation. wondering if its issue with rootless only containers as your issue suggest in nerdctl?

haytok:
The issue occurs in a rootless container created with nerdctl and containerd.

The error message indicates that the nerdctl attach process cannot access the directory /var/lib/nerdctl/, which is why this error occurs.

```
[haytok@lima-finch nerdctl]$ ls /var/lib/nerdctl/1935db59/
containers etchosts names volumes
```

However, I have not been able to verify that the error does not occur when operating from a non-rootless container.

Shubhranshu153:
it occurs in sudo mode also, left comments on nerdctl channel.

PR #876, opened 2024-03-28
fix: parse management commands with proper arguments

Shubhranshu153:
Issue #, if available: https://github.com/runfinch/finch/issues/827

Description of changes:
Addresses https://github.com/runfinch/finch/issues/827 by adding all environment variables at the location of first occurrence of the environment flag

Testing done:
Changes tested locally to ensure that the issue is fixed.

$ ./_output/bin/finch container run --debug -it --rm --env="E=v" busybox env
DEBU[0000] Creating limactl command: ARGUMENTS: [shell finch sudo -E nerdctl container run -e E=v -it --rm busybox env], LIMA_HOME: /Users/mharwani/work/runfinch/finch/_output/lima/data 
E=v
TERM=xterm
HOME=/root
*Testing done:*

#### License Acceptance

### DIFF
--- a/cmd/finch/nerdctl.go
+++ b/cmd/finch/nerdctl.go
@@ -111,6 +111,8 @@ func (nc *nerdctlCommand) run(cmdName string, args []string) error {
 		skip, hasCmdHander, hasArgHandler bool
 		cmdHandler                        commandHandler
 		aMap                              map[string]argHandler
+		envArgPos                         int
+		isDebug                           int
 	)
 
 	alias, hasAlias := aliasMap[cmdName]
@@ -139,6 +141,10 @@ func (nc *nerdctlCommand) run(cmdName string, args []string) error {
 		}
 	}
 
+	// envArgPos is used to preserve the position of first environment parameter
+	envArgPos = -1
+	// if a debug flag is passed before env arg pos we reduce the env arg pos by 1 to account for skipping debug flag
+	isDebug = 0
 	for i, arg := range args {
 		// Check if command requires arg handling
 		if hasArgHandler {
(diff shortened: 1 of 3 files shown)


cmd/finch/nerdctl.go
```diff
@@ -253,10 +269,12 @@ func (nc *nerdctlCommand) run(cmdName string, args []string) error {
 	for key, val := range envVars {
 		finalArgs = append(finalArgs, "-e", fmt.Sprintf("%s=%s", key, val))
 	}
-	finalArgs = append(finalArgs, nerdctlArgs...)
```
mharwani, line 256: nit: we should rename this to something else like `envArgs`

cmd/finch/nerdctl.go
```diff
@@ -164,14 +170,24 @@ func (nc *nerdctlCommand) run(cmdName string, args []string) error {
...
+				envArgPos = i - isDebug
+			}
 			shouldSkip, addEnv := handleEnv(nc.systemDeps, arg, args[i+1])
 			skip = shouldSkip
 			if addEnv != "" {
 				envs = append(envs, addEnv)
 			}
 		case strings.HasPrefix(arg, "--env-file"):
+			if envArgPos == -1 {
+				envArgPos = i
```
mharwani, line 188: `i - isDebug`?

Shubhranshu153: 👍

pendo324:
LGTM pending the nits. Can you also update the title and description to be more accurate? It seems like it's just copied from #830. It fixes the same type of issue, but it doesn't do exactly what #830 does

pendo324:
> Thanks for the changes. There can be problems with env passed after the image name. For eg:
> 
> ```
> finch container run alpine echo -e hello
> ```
> 
> Might need another approach to handle such cases, but we can do that in a different PR.

That's a good point, we don't want to replace any random `-e`'s  that are passed in as a command since 

```bash
finch container run alpine echo -e hello
```

is a valid command already and replacing `hello` with the value of a `hello` environment variable would change the meaning of the command.

Maybe we need to special-case any commands that take command line arguments (`run` and `exec`) specifically?

Shubhranshu153:
need to think of a good solution for it and might need to refactor the code, in general i would want to avoid any fiddling with the command parsing as it creates undefined behavior for the user. But there are edge cases here.

For not would keep that separate from this PR as it doesnt seem to be directly related to the issue.

pendo324 (approved):

PR #947, opened 2024-05-15
ci: remove macOS 11 release runners

pendo324:
Issue #, if available:
- https://github.com/runfinch/finch/actions/runs/9092946617/job/24990820049#step:4:172
    - Builds are failing when running on older versions of macOS due to the new dependency on yq, which we install via brew. It is possible to install this dependency through other means (curling / chmoding / moving), but this seems like a better long-term solution.

*Description of changes:*
- Removes the need to build finch on macOS 11. This was previously done so that our binaries would be compatible with older versions of macOS. However, we can get around this by using the `CGO` flags that are in this diff, entirely removing the need to build on older macOS versions. In the future, these hosts can be completely dropped from the CI as well, as macOS 11 is no longer one of the 2 most recent versions of macOS.
- Going forward, it may be possible to use something like [crazy-max/xgo](https://github.com/crazy-max/xgo?tab=readme-ov-file) or [crazy-max/goxx](https://github.com/crazy-max/goxx?tab=readme-ov-file), with the [osxcross toolchain](https://github.com/crazy-max/goxx?tab=readme-ov-file#macosx-cross-toolchain) to containerize builds of Finch, making them more reproducible and removing the need for macOS build/release hosts entirely.

*Testing done:*
- Verified finch binaries built on macOS 14 run properly on macOS 11 hosts
- Since this isn't obvious, this is how you can at least test that the compiler used the flags:
    ```shell
    otool -l ./_output/bin/finch | grep minos
        minos 11.0
    ```

#### License Acceptance

### DIFF
--- a/.github/workflows/build-and-test-pkg.yaml
+++ b/.github/workflows/build-and-test-pkg.yaml
@@ -58,10 +58,7 @@ jobs:
           git clean -f -d
           make clean
           make download-licenses 
-          CGO_CFLAGS="-mmacosx-version-min=11.0" \
-            CGO_LDFLAGS="-mmacosx-version-min=11.0" \
-            FINCH_OS_IMAGE_LOCATION_ROOT=/Applications/Finch \
-            make
+          make
         shell: zsh {0}
 
       - name: configure aws credentials
@@ -99,10 +96,7 @@ jobs:
           git clean -f -d
           make clean
           make download-licenses 
-          CGO_CFLAGS="-mmacosx-version-min=11.0" \
-            CGO_LDFLAGS="-mmacosx-version-min=11.0" \
-            FINCH_OS_IMAGE_LOCATION_ROOT=/Applications/Finch \
-            make
+          make
         shell: zsh {0}
 
       - name: configure aws credentials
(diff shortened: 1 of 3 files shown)


.github/workflows/build-and-test-pkg.yaml
```diff
@@ -58,7 +58,10 @@ jobs:
           git clean -f -d
           make clean
           make download-licenses 
-          make FINCH_OS_IMAGE_LOCATION_ROOT=/Applications/Finch
+          CGO_CFLAGS="-mmacosx-version-min=11.0" \
+            CGO_LDFLAGS="-mmacosx-version-min=11.0" \
+            FINCH_OS_IMAGE_LOCATION_ROOT=/Applications/Finch \
+            make
```
austinvazquez, line 64: Should we consider pushing this down into the Makefile so it is easier to update the min version if required?

pendo324: Yeah, that's a good point. Do you think it makes sense to have a new target just for releases, or to just always set these parameters?

pendo324: I just added it for all mac targets, PTAL

austinvazquez: > I just added it for all mac targets, PTAL

Should be fine sense our docs state a minimum anyways. Advanced users can always override with the environment variable.

austinvazquez (approved):

PR #953, opened 2024-05-17
ci(refactor): split build-and-test-pkg for better reuse

pendo324:
Issue #, if available:

*Description of changes:*
While I was fixing the macOS pkg build error on macOS 11 (due to `yq` from brew not building properly), I noticed that most of the file was repeated (due to different archs being built/tested as different jobs) which caused it to be pretty unnecessarily unwieldy. This was the case for both the build and the test workflow jobs.

I initially thought I could combine them using matrices, but then we lose the ability to run the test jobs as soon as each respective build completes (with a matrix, it would have to wait until _both_ builds complete, which isn't even guaranteed). There were other complications that made matrices hard, like the fact that we only want to build once per arch, but we want to test on several other arch/version combinations.

Instead, I was able to split the workflows into reusable/composite workflows, so that only the inputs vary, and the script is mostly the same.

*Testing done:*

#### License Acceptance


pendo324:
Release action run test. It gets past the previously failing step (Install build dependencies)
https://github.com/runfinch/finch/actions/runs/9130235009/job/25107445223

Shubhranshu153 (approved):

.github/workflows/build-pkg.yaml
```diff
@@ -0,0 +1,78 @@
+name: build-pkg
+on:
+  workflow_call:
+    inputs:
```
swagatbora90, line 4: Is it possible to specify which values are acceptable for each input type?

.github/workflows/build-and-test-pkg.yaml
```diff
@@ -35,316 +29,59 @@ jobs:
...
-        run: |
-          ./installer-builder/tools/release-installer.sh aarch64 ${{ needs.get-tag-name.outputs.tag }} ${{ secrets.INSTALLER_PRIVATE_BUCKET_NAME }} ${{ secrets.EXECUTABLE_BUCKET }} ${{ secrets.PKG_BUCKET }} ${{ secrets.NOTARIZATION_ACCOUNT }} ${{ secrets.NOTARIZATION_CREDENTIAL }}
-        shell: zsh {0}
+    uses: ./.github/workflows/build-pkg.yaml
+    secrets: inherit
+    with:
+      os: macos
+      arch: arm64
+      output_arch: aarch64
+      version: 14
```
swagatbora90, line 41: Curious why we are not building for the other versions (11,13) but we are testing it later.

pendo324: Finch should run on multiple versions of macOS, but you only have to build the executable once per architecture in order to support that, thanks to the changes to the Makefile in https://github.com/runfinch/finch/pull/947.

.github/workflows/build-pkg.yaml
pendo324: I don't think so, according to [the docs](https://docs.github.com/en/enterprise-cloud@latest/actions/using-workflows/workflow-syntax-for-github-actions#onworkflow_callinputs), [this blog post](https://github.blog/changelog/2021-11-10-github-actions-input-types-for-manual-workflows/), and [the source code](https://github.com/actions/languageservices/blob/5dbaa884dbd889a68eb184eb0f28c0f6618264ee/workflow-parser/src/model/workflow-template.ts#L167-L180).

swagatbora90: this [options](https://github.com/actions/languageservices/blob/5dbaa884dbd889a68eb184eb0f28c0f6618264ee/workflow-parser/src/model/workflow-template.ts#L179) field?

swagatbora90: It is always good to add some input validation when creating templates. Not a blocker for this PR though. We can look into it later.

swagatbora90 (approved):
LGTM!

.github/workflows/build-pkg.yaml
pendo324: I think that only works for manually triggered workflows, to give a drop down as seen in [this blog](https://tkolber.medium.com/dropdown-input-choice-workflows-in-github-actions-e95cf140bf61). Could be wrong, though.

PR #971, opened 2024-06-06
ci: update release script to embed Info.plists

pendo324:
Issue #, if available:

*Description of changes:*
Embed Info.plists into all of the executables that we vend. Requires stripping debug information from the go executable (see [this issue](https://github.com/golang/go/issues/62577)).

*Testing done:*
Tested manually

#### License Acceptance

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -24,7 +24,7 @@ CORE_VDE_PREFIX ?= $(OUTDIR)/dependencies/vde/opt/finch
 LICENSEDIR := $(OUTDIR)/license-files
 VERSION := $(shell git describe --match 'v[0-9]*' --dirty='.modified' --always --tags)
 GITCOMMIT := $(shell git rev-parse HEAD)$(shell test -z "$(git status --porcelain)" || echo .m)
-LDFLAGS := "-X $(PACKAGE)/pkg/version.Version=$(VERSION) -X $(PACKAGE)/pkg/version.GitCommit=$(GITCOMMIT)"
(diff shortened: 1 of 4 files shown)


Kern-- (approved):

Makefile
```diff
@@ -24,7 +24,7 @@ CORE_VDE_PREFIX ?= $(OUTDIR)/dependencies/vde/opt/finch
 LICENSEDIR := $(OUTDIR)/license-files
 VERSION := $(shell git describe --match 'v[0-9]*' --dirty='.modified' --always --tags)
 GITCOMMIT := $(shell git rev-parse HEAD)$(shell test -z "$(git status --porcelain)" || echo .m)
-LDFLAGS := "-X $(PACKAGE)/pkg/version.Version=$(VERSION) -X $(PACKAGE)/pkg/version.GitCommit=$(GITCOMMIT)"
+LDFLAGS = "-w -X $(PACKAGE)/pkg/version.Version=$(VERSION) -X $(PACKAGE)/pkg/version.GitCommit=$(GITCOMMIT)"
```
swagatbora90, line 27: Curious if this reduces the size of the finch binary?
Also, what are my options if I need a debug build?

swagatbora90 (approved):
LGTM

Makefile
pendo324: Yes, it did reduce the executable size by a few megabytes. According to https://github.com/golang/go/issues/62577, Embedding debug symbols like this is wrong on macOS anyway. We can probably add `-w` only on macOS if we need the debugging symbols on other platforms. The Makefile is only really used for release builds right now though.

PR #993, opened 2024-06-20
ci: add workaround for WSL hanging in CI

austinvazquez:
Issue #, if available:
There is a known issue, https://github.com/microsoft/WSL/issues/8529, where WSL commands can hang. This can cause Windows e2e tests to block until hitting the 2 hour timeout.

*Description of changes:*
This change adds a workaround to detect the bad state and attempt to mitigate by killing the WSL service. If the issue cannot be resolved, the test will only hang for 300 seconds before failing.

*Testing done:*
CI run was successful with 8 WSL shutdown failures.
https://github.com/runfinch/finch/actions/runs/9682445232/job/26715743040

image

*Trade-off analysis*
The trade-off for this approach is the test suite can take longer with multiple reset VM calls being made. Sample runs which previously took ~15 minutes are up to ~37 minutes with the hanging mitigation; however, this is down from the 2 hour timeout failure which would occur without the mitigation.

#### License Acceptance

### DIFF
--- a/e2e/vm/vm_darwin_test.go
+++ b/e2e/vm/vm_darwin_test.go
@@ -96,3 +96,8 @@ var resetDisks = func(_ *option.Option, installed bool) {
 	}
 	gomega.Expect(os.RemoveAll(dataDiskDir)).ShouldNot(gomega.HaveOccurred())
 }
+
+var shutdownWSL = func() error {
+	// no-op on darwin
+	return nil
+}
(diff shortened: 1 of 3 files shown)


pendo324 (approved):
Hopefully this works better than the naive shutdown command 👍

austinvazquez:
> Hopefully this works better than the naive shutdown command 👍

From the microsoft/WSL issue, some users reported shutdown taking 2-3 minutes. We can also consider being more aggressive than this and killing the WSL service faster.

sondavidb (approved):

PR #1007, opened 2024-07-01
ci(release): fix release build generation

pendo324:
Issue #, if available:

*Description of changes:*
Recently in https://github.com/runfinch/finch/pull/969, the Makefile was refactored significantly, and the mechanism that was used to override the installation directory was removed ([used here](https://github.com/runfinch/finch/blob/main/.github/workflows/build-pkg.yaml#L57)). This just adds that mechanism back. There could be a better way to do this, but for now just reverting back to how it was done before the recent change.

*Testing done:*
Tested locally. The `_output/os/finch.yaml` file correctly used `/Applications/Finch/`, which was the path I overwrote, as a prefix

#### License Acceptance

### DIFF
--- a/Makefile.darwin
+++ b/Makefile.darwin
@@ -12,8 +12,11 @@ FINCH_OS_BASENAME=$(AARCH64_ARTIFACT)
 FINCH_OS_DIGEST=$(AARCH64_512_DIGEST)
 endif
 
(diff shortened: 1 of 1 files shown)


austinvazquez (approved):
Nice find.

PR #1043, opened 2024-07-22
Fix override provisioning script

Shubhranshu153:
Issue #, if available:

*Description of changes:*
Fix override provisioning script to add provisioning script only with qemu

*Testing done:*
Local Testing

#### License Acceptance


pendo324:
qemu user mode emulation binaries should still be configurable when using Virtualization.framework. There may be workloads which do not work under Rosetta emulation, but do work under qemu's emulation (which is why they have separate toggles in `finch.yaml`).

If both Rosetta and qemu-binfmt binaries are configured, Rosetta should be the one used due to Linux implementation details (I don't think its documented).

PR #1070, opened 2024-08-22
feat: support pass cred store

Shubhranshu153:
Issue #, if available:

*Description of changes:*
Feature to support pass cred store. 

TODO:
1. Add unit test
2. Add documentation 

*Testing done:*

Env:
1. sudo yum install pass -y
3. curl -fsSL https://github.com/docker/docker-credential-helpers/releases/download/v0.8.2/docker-credential-pass-v0.8.2.linux-amd64 -o docker-credential-pass
4. chmod +x /usr/local/bin/docker-credential-pass
5. sudo mv docker-credential-pass /usr/local/bin/

Test:
1. Initialize Pass: sudo finch pass-init  
2. Config finch at /home/shubhum.linux/.finch/config.json: 
```
{
        "credsStore": "pass"
}
```
3. Get Account creds (any account)
4. Run ECR login: aws ecr get-login-password --region us-west-2 | sudo DOCKER_CONFIG=/home/shubhum.linux/.finch/  ./_output/bin/finch login --username AWS --password-stdin <account>.dkr.ecr.us-west-2.amazonaws.com
6. Delete Pass: sudo finch pass-delete

- [x ] I've reviewed the guidance in CONTRIBUTING.md

#### License Acceptance


cmd/finch/pass_init_native.go
```diff
@@ -0,0 +1,25 @@
...
+
+package main
+
+import (
+	"github.com/gopasspw/gopass/pkg/pwgen"
+	"github.com/runfinch/finch/pkg/command"
+)
+
+func (pia *passInitAction) initGpgKey() command.Command {
+	passphrase := pwgen.GeneratePassword(passphraseLength, true)
```
pendo324, line 14: Since we're not even asking users to enter a passphrase, can't we do this behind the scenes? Should we instead have a mechanism that initializes the cred store whenever a user modifies their `creds_helpers` to include `pass`?

Shubhranshu153: deletion is problematic,
i would prefer not taking responsibility of key management, if its done with a command customer has responsibility to manage key lifecycle.

pendo324: We can still have the commands, but if the commands require 0 user input, I don't see the harm in doing it for them

Shubhranshu153: i still feel if we create it the lifecycle is on us to manage for example key rotation etc. if it is automatically done finch has to won the management of the key.

pendo324: What's the purpose of key rotation for a key where we don't even care about the password?

cmd/finch/pass_init_native.go
```diff
@@ -0,0 +1,25 @@
...
+	ecc := command.NewExecCmdCreator()
+	cmd := ecc.Create(
+		"gpg2", "--batch", "--passphrase", passphrase, "--quick-gen-key", "finch")
```
swagatbora90, line 17: Does this key expire after some default period?

Shubhranshu153: key doesnt expire, the key creation is owned by the customer so they would need to remove it.

cmd/finch/pass_init_native.go
Shubhranshu153: if the key is compromised, we probably want to rotate it, and having set time to rotate it probably makes it harder to breach within certain time period.

PR #1071, opened 2024-08-23
build(linux): add Linux packaging and e2e tests

pendo324:
Issue #, if available:

*Description of changes:*
- Add packaging and e2e runners which use the packages to run tests
- Before merging this, we need to:
  - add linux runners to this repository (by merging https://github.com/runfinch/infrastructure/pull/683)
  - remove the change in this PR which makes it so CI runs on all branches, not just main
  - uncomment the disabling of the macOS/Windows runners

*Testing done:*
- Tested on this PR to my fork, which has linux runners added https://github.com/pendo324/finch/pull/7

#### License Acceptance


e2e/container/container_test_other.go
```diff
@@ -0,0 +1,15 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
Shubhranshu153, line 1: what is other mean?

e2e/container/container_test.go
```diff
@@ -59,9 +66,13 @@ func TestContainer(t *testing.T) {
 			// containerd expects it at /sys/fs/cgroup based on
 			// https://github.com/containerd/cgroups/blob/cc78c6c1e32dc5bde018d92999910fdace3cfa27/utils.go#L36
 			tests.Run(&tests.RunOption{BaseOpt: o, CGMode: tests.Hybrid, DefaultHostGatewayIP: hostIP})
-		} else {
+		case "darwin":
 			tests.Run(&tests.RunOption{BaseOpt: o, CGMode: tests.Unified, DefaultHostGatewayIP: "192.168.5.2"})
+		default:
```
Shubhranshu153, line 71: nit: i am assuming this is for linux but may be good to be explicit, default seems to be a case of unknown

contrib/packaging/rpm/finch-buildkit.service
```diff
@@ -0,0 +1,12 @@
+[Unit]
```
Shubhranshu153, line 1: are the service launch etc are tested somewhere else already?

Shubhranshu153 (approved):

.github/workflows/e2e-linux.yaml
```diff
@@ -0,0 +1,104 @@
...
+          # We need to get all the git tags to make version injection work. See VERSION in Makefile for more detail.
+          fetch-depth: 0
+          persist-credentials: false
+          submodules: recursive
+      - name: Check repo out manually
+        if: ${{ (startsWith(inputs.os, 'amazon') && inputs.version == '2' ) }}
+        run: |
+          pwd
+          echo "${GITHUB_WORKSPACE}"
+          ls -lah
```
swagatbora90, line 52: what is the purpose of this `ls`

.github/workflows/e2e-linux.yaml
```diff
@@ -0,0 +1,104 @@
...
+        if: always()
+        run: |
+          sudo systemctl stop finch.service || true
+          sudo systemctl stop finch-buildkit.service || true
+          sudo systemctl stop finch-soci.service || true
+          sudo sudo rpm -e runfinch-finch || true
+          sudo systemctl daemon-reload
+          sudo rm -rf /etc/finch
+          sudo rm -rf /var/finch
+          sudo rm -rf /var/containerd
```
swagatbora90, line 81: This is not where containerd's stores its data, right? I see that we are not restarting containerd service later, which may be required if we are removing the entire store here.

contrib/packaging/rpm/finch.spec
```diff
@@ -0,0 +1,266 @@
...
+%build
+
+# build finch
+pushd "%{finch_src}"
+VERSION=%{finch_release} GITCOMMIT=%{finch_commit} make finch
+popd
+
+# build buildkit
+pushd "%{buildkit_src}"
+go build -mod=vendor -buildmode=pie -ldflags '-w -X %{buildkit_package}/version.Version=%{buildkit_rpm_version}' -o _output/buildkitd %{buildkit_package}/cmd/buildkitd
```
swagatbora90, line 131: do we need to define a min go version?

contrib/packaging/rpm/finch.spec
```diff
@@ -0,0 +1,266 @@
...
+%{_libexecdir}/finch/cosign
+
+# nerdctl
+%{_libexecdir}/finch/nerdctl
+
+%pre
+# Stop the agent before the upgrade
+STATUS="$(systemctl is-active --quiet finch.service)"
+if [ "${STATUS}" = "active" ]; then
+  systemctl stop finch.service
```
swagatbora90, line 241: Use macros for systemctl?

swagatbora90 (commented):
Minor nits. Looks good otherwise.

e2e/container/container_test_other.go
pendo324: It means "not Linux". `getCGroupMode` doesn't really make sense unless you're running on Linux

contrib/packaging/rpm/finch.spec
pendo324: I think we will just keep it as low as possible, it will depend on when different dependencies upgrade

contrib/packaging/rpm/finch-buildkit.service
pendo324: Currently when installing Finch, all of these services are launched automatically by the `%post` scripts in the RPM spec. I'm going to switch to using [RPM systemd scriptlet macros](https://docs.fedoraproject.org/en-US/packaging-guidelines/Scriptlets/#_systemd), so we will now have to manually start finch.service, which will also start all of its related services (currently, buildkit and soci)

.github/workflows/e2e-linux.yaml
pendo324: debug, it can be removed

.github/workflows/e2e-linux.yaml
pendo324: You're right, this should be `/var/lib/containerd`, and we can also restart the service before starting finch

pendo324:
The e2e test names changed due to renaming / splitting them into different files. After merging, I'll have to change the "required checks" in the GitHub settings.

Issue #745, opened 2024-01-02
devcontainer cli - support --sig-proxy=false


jsamuel1:
**What is the problem you're trying to solve?.**
Running @devcontainer/cli to launch a devcontainer, the process fails at the run step due to devcontainer/cli passing the option `--sig-proxy=false` to the run action when running a single container.

```
[4483 ms] Start: Run: finch run --sig-proxy=false -a STDOUT -a STDERR --mount type=bind,source=XXXXXXXX,target=/workspaces/XXXXXXXXX,consistency=cached -l devcontainer.local_folder=XXXXXXXX -l devcontainer.config_file=XXXXXXXX/.devcontainer/devcontainer.json -e NODE_OPTIONS=--max-old-space-size=8192 --cap-add=SYS_ADMIN --entrypoint /bin/sh vsc-xxxxxxxxxxxxxxxxxx -c echo Container started
FATA[0000] unknown flag: --sig-proxy
FATA[0001] exit status 1
```

**Describe the feature you'd like**
Would like the option to either be support or ignored.

**Additional context**
https://github.com/devcontainers/cli/blob/f7d4c853bf8c284d784173f3e915a34d961b0b55/src/spec-node/singleContainer.ts#L396

Reproducing steps:
From a project with a .devcontainer folder/setup that has a single Dockerfile and devcontainer.json, 
run:
`npx devcontainer up --docker-path finch --docker-compose-path "finch compose" --workspace-folder . `

jsamuel1:
```
devcontainer --version
0.55.0
```

jsamuel1:
Note the -- the sig-proxy=false argument is to prevent signals (eg. Ctrl-C) being sent along to the process inside the container.   Default was to pass the signal in.

jsamuel1:
Corresponding item in vscode devcontainer extension issues:  https://github.com/microsoft/vscode-remote-release/issues/6014

Note - this linked issue is for the extension, not the base @devcontainer packages.

PR #1100, opened 2024-09-19
Dockercompat

chews93319:
Issue #, if available: #387, #745

*Description of changes:*

*Testing done:*

#### License Acceptance


chews93319:
With commit 999371ea, I resolved the `buildx` bug and confirmed that the `handleCache` is actually working correctly and extracting the "consistency" strings from the args.

Unfortunately, my devContainers experiment  was now able to build an image, but unable to open it due to the length of identifier.

```
FATA[0000] invalid name "vsc-arbtest_devcontainers_features-f31e83368f02f49b07b37f8b815ef72280fdebe4e5bb5de366af337b5e70ae4a-features-62cc1": identifier "vsc-arbtest_devcontainers_features-f31e83368f02f49b07b37f8b815ef72280fdebe4e5bb5de366af337b5e70ae4a-features-62cc1" greater than maximum length (76 characters): invalid argument 
```

cmd/finch/nerdctl_remote.go
```diff
@@ -501,3 +539,134 @@ func handleEnvFile(fs afero.Fs, systemDeps NerdctlCommandSystemDeps, arg, arg2 s
...
+
+	for _, arg := range limaArgs {
+		if arg == "buildx" {
+			buildx = true
+			newLimaArgs = append(newLimaArgs, "build")
+			logrus.Warn("buildx is not supported. using standard buildkit instead...")
+			continue
+		}
+
+		// Find standard vsc container identifier to truncate to less than 76chars
```
Shubhranshu153, line 648: this change we should not be doing and atleast not here, any change to identifier is a risky change and can impact other things, 

Tracing the origin:
This portion is from https://github.com/devcontainers/cli/blob/9ba1fdaa11dee087b142d33e4ac13c5788392e34/src/spec-node/utils.ts#L527

and  that is getting called in: https://github.com/devcontainers/cli/blob/9ba1fdaa11dee087b142d33e4ac13c5788392e34/src/spec-node/singleContainer.ts#L121


We can configure this under devcontainer.json

config {
 image: <name>
}
This can be part of our docs instead of doing it like this.

cmd/finch/nerdctl_remote.go
```diff
@@ -157,25 +170,38 @@ func (nc *nerdctlCommand) run(cmdName string, args []string) error {
 					envs = append(envs, addEnv)
 				}
 			case shortFlagBoolSet.Has(arg) || longFlagBoolSet.Has(arg):
-				// exact match to a short flag: -?
+				// exact match to a short no argument flag: -?
 				// or exact match to: --<long_flag>
 				nerdctlArgs = append(nerdctlArgs, arg)
+			case longFlagBoolSet.Has(strings.Split(arg, "=")[0]):
```
Shubhranshu153, line 176: we can add this to the or statement to the above case

cmd/finch/nerdctl_remote.go
```diff
@@ -501,3 +539,134 @@ func handleEnvFile(fs afero.Fs, systemDeps NerdctlCommandSystemDeps, arg, arg2 s
...
+	case "":
+		break
+	default:
+		newArgList = append(newArgList[:inspectIndex], append([]string{inspectType}, newArgList[inspectIndex+1:]...)...)
+	}
+
+	return newArgList
+}
+
+func handleBuildx(fc *config.Finch, limaArgs []string) (bool, []string) {
```
Shubhranshu153, line 630: am thinking if we should first check for buildx and create a flag so that we dont enter this loop again, we already have all the info from the first loop

Shubhranshu153: the intial arg loop

cmd/finch/nerdctl_remote.go
chews93319: You were correct about the devcontainers code reference.
I also followed it to the `getFolderImageName` and found the specific concatenation that was creating the "extra long name" (including folderHash): https://github.com/devcontainers/cli/blob/9ba1fdaa11dee087b142d33e4ac13c5788392e34/src/spec-node/utils.ts#L527

cmd/finch/nerdctl_remote.go
chews93319: Agreed, the recent update has refactored into a handleBuildx mechanism and modifies-in-place the "buildx <sub_cmd>" scenarios.

cmd/finch/nerdctl_remote.go
chews93319: All of the switch case statements were intentionally broken out for explicit commenting for future maintenance.
Nevertheless, based on your feedback and feedback from Justin, I am currently refactoring the "handlers" and "processing loop" to move away from the large quantity of processing case scenarios.

chews93319:
The update to commit 9672022 is an incremental improvement:
 - repaired the compile error for linux e2e tests
 - refactored handleBuildx for use in all 3 platforms (Darwin, Windows, Native)
 - refactored Docker build --load handler for use in all 3 platforms (Darwin, Windows, Native)

cmd/finch/nerdctl_remote.go
```diff
@@ -501,3 +539,57 @@ func handleEnvFile(fs afero.Fs, systemDeps NerdctlCommandSystemDeps, arg, arg2 s
...
+			skipNext = false
+			continue
+		}
+		if arg == "inspect" {
+			isInspect = true
+			inspectIndex = idx
+			newArgList = append(newArgList, arg)
+		}
+
+		if strings.Contains(arg, "--type") && (idx < len(args)-1) && isInspect {
```
pendo324, line 565: Should we be using the slices package to check if the args slice has a string `--type` in case the `inspect` arg comes after the `--type` arg?

pkg/config/config.go
```diff
@@ -38,6 +38,7 @@ type SharedSystemSettings struct {
 type SharedSettings struct {
 	Snapshotters []string `yaml:"snapshotters,omitempty"`
 	CredsHelpers []string `yaml:"creds_helpers,omitempty"`
+	Mode         *string  `yaml:"mode,omitempty"`
```
pendo324, line 41: I wonder if `mode` is too generic of a name here. Maybe `dockercompat: true` makes more sense?

chews93319: Acknowledged!
Changing in next revision...

pendo324 (approved):
There are still linting errors, but LGTM as long as all of the automated approval workflows pass

Shubhranshu153 (approved):

cmd/finch/nerdctl_windows.go
```diff
@@ -197,7 +247,7 @@ func handleBindMounts(systemDeps NerdctlCommandSystemDeps, nerdctlCmdArgs []stri
 }
 
 // handles --output/-o for build command.
-func handleOutputOption(systemDeps NerdctlCommandSystemDeps, nerdctlCmdArgs []string, index int) error {
+func handleOutputOption(systemDeps NerdctlCommandSystemDeps, _ *config.Finch, nerdctlCmdArgs []string, index int) error {
```
Shubhranshu153, line 250: nit: any particular reason to prefer pointers, Nothing wrong with it just want to know the reasoning.

cmd/finch/nerdctl_darwin.go
```diff
@@ -46,3 +49,70 @@ func resolveIP(host string, logger flog.Logger, _ command.Creator) (string, erro
 	}
 	return host, nil
 }
+
+// removes the consistency key-value entity from --mount.
+func handleBindMounts(_ NerdctlCommandSystemDeps, _ *config.Finch, nerdctlCmdArgs []string, index int) error {
```
Shubhranshu153, line 54: do we need to pass args that are not getting used or they needed for different architectures?

chews93319: The handler interface is set by [nerdctl.go line46](https://github.com/runfinch/finch/pull/1100/files#diff-403cda46333c3f741f8b1ff7f391901b0ccb8a5501cdd241138f89fd3c9db2c6R46), so to be "least impactful" these new handlers conformed to the same signature, but did not necessarily require the same args.

cmd/finch/nerdctl_windows.go
chews93319: If I'm reading the GitHub highlighting correctly, I think that you commenting on the `*config.Finch` pointer.

Since the config was originally defined within the `fc` as such, I replicated it rather than trying to convert it into a new type. Additionally, I thought if there were other future opportunities for the Finch Config to change handler behavior, then the entire object would already be available.

cmd/finch/nerdctl.go
```diff
@@ -215,3 +243,99 @@ var cmdFlagSetMap = map[string]map[string]sets.Set[string]{
...
+			*args = append([]string{}, savedArgs...)
+		}
+	case "container":
+		*cmdName = "inspect"
+		*args = append([]string{modeDockerCompat}, savedArgs...)
+	case "":
+		*cmdName = "inspect"
+		*args = append([]string{modeDockerCompat}, savedArgs...)
+	default:
+		return fmt.Errorf("unsupported inspect type: %s", inspectType)
```
Shubhranshu153, line 337: default should not be not supported i think, it should be returned as it is and throw an error as per nerdctl

generic order that nerdctl support is

finch <type> inspect <options> identifier (this is what is supported by nerdctl)
I would say have another pr later this is fine for now for devcontainer use case

Shubhranshu153 (approved):

cmd/finch/nerdctl_remote.go
Shubhranshu153: for now resolving it,  for semantics, if its too concerning will fix in my pr which will be based on top of it.

PR #1109, opened 2024-09-23
fix: allow finch commands to run using default config

pendo324:
Issue #, if available:

*Description of changes:*
- Makes it so config requirements are not as strict. This allows basic commands, like `finch help` and `finch version` to work on Linux, even when not running as root.
- Also added a warning message when the running user does not have permission to read the config file.

*Testing done:*

#### License Acceptance

### DIFF
--- a/cmd/finch/main_native.go
+++ b/cmd/finch/main_native.go
@@ -44,7 +44,7 @@ func xmain(logger flog.Logger,
 	)
 	if err != nil {
 		if errors.Is(err, os.ErrPermission) {
-			logger.Warnf("Failed to load config, using default values. Are you running as root? (%s)", err)
+			logger.Warnf("Failed to load config, using default values. You may need to be root or use sudo. (%s)", err)
 		} else {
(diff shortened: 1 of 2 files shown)


cmd/finch/main_native.go
```diff
@@ -41,7 +43,11 @@ func xmain(logger flog.Logger,
 		ecc,
 	)
 	if err != nil {
-		return fmt.Errorf("failed to load config: %w", err)
+		if errors.Is(err, os.ErrPermission) {
+			logger.Warnf("Failed to load config, using default values. Are you running as root? (%s)", err)
```
swagatbora90, line 47: Slight nitpick to make the message clear: Prefer "You may need to be root or use sudo"

swagatbora90 (approved):

cmd/finch/main_native.go
pendo324: Updated message PTAL

swagatbora90 (approved):

PR #1119, opened 2024-10-02
feat: Add BindMount handler for Linux

chews93319:
Add handleBindMount to remove consistency entity
Revise comments for linting and clarity

Issue #, if available: #387

*Description of changes:*
When using DevContainers on Finch within Linux, the operating specific handleBindMounts will remove the "consistency" entity from the argument set.

*Testing done:*
- draft PR Github CI Tests.

#### License Acceptance


cmd/finch/nerdctl_native.go
```diff
@@ -89,6 +87,73 @@ func (nc *nerdctlCommand) run(cmdName string, args []string) error {
...
-var osArgHandlerMap = map[string]map[string]argHandler{}
+var osArgHandlerMap = map[string]map[string]argHandler{
+	"container run": {
+		"--mount": handleBindMounts,
+	},
+}
 
 var osCommandHandlerMap = map[string]commandHandler{}
+
+func handleBindMounts(_ NerdctlCommandSystemDeps, _ *config.Finch, nerdctlCmdArgs []string, index int) error {
```
pendo324, line 98: Is it possible to combine the 3 separate `handleBindMounts` functions in the `_darwin` `_windows` and `_native` files? They seem similar

chews93319: Agreed, they are very similar.
I believe that there is an opportunity to move the `handleBindMounts` to the level of `nerdctl.go`, but the WSL path manipulation is the detail currently anchoring them at the OS level.

pendo324: There are two ways (which are roughly the same):
1. You could add a new function which encapsulates all of the OS specific logic (e.g. `handleBindmountPath`, or something), and does nothing on Linux (this is what I would prefer)
1. You could use runtime.GOOS to gate the logic

chews93319: Please refer to the [commit 6f4f775](https://github.com/runfinch/finch/commit/6f4f77561567c5d331d2f24dbf80c4fc98786d5a) as resolution to this comment.

cmd/finch/nerdctl_native.go
```diff
@@ -92,3 +89,17 @@ var osAliasMap = map[string]string{}
 var osArgHandlerMap = map[string]map[string]argHandler{}
 
 var osCommandHandlerMap = map[string]commandHandler{}
+
+func mapToString(m map[string]string) string {
```
pendo324, line 93: Can de-dupe this code too? Not a priority and can be done in a followup PR.

pendo324 (approved):

cmd/finch/nerdctl_native.go
chews93319: +1 Agreed!

In another iteration, it could be possible to completely refactor all of the handlers into the generalized level with OS-specific handlerHelpers and simplifying the handlerMap aggregation.
This de-dupe and optimization can be included then.

PR #1121, opened 2024-10-03
feat: handle dockercompat inspect for devcontainers

Shubhranshu153:
Issue #, if available:

*Description of changes:*
feat: handle dockercompat inspect for devcontainers

*Testing done:*
local dev container testing

#### License Acceptance


cmd/finch/nerdctl.go
```diff
@@ -45,7 +45,7 @@ type nerdctlCommandCreator struct {
 
 type (
 	argHandler     func(systemDeps NerdctlCommandSystemDeps, fc *config.Finch, args []string, index int) error
-	commandHandler func(systemDeps NerdctlCommandSystemDeps, fc *config.Finch, cmdName *string, args *[]string) error
+	commandHandler func(systemDeps NerdctlCommandSystemDeps, fc *config.Finch, cmdName *string, args *[]string, inspectType *string) error
```
chews93319, line 48: This is an interesting mechanism to enable the specific `handleDockerCompatInspect` to affect operation of the parent process.
Perhaps, in the future, we can generalize this var type so that it is more useful for any of the `cmdHandler` functions, such as a `map[string]string`.

cmd/finch/devcontainer_patch.go
```diff
@@ -0,0 +1,100 @@
...
+	}
+
+	fmt.Println(string(finalJSON))
+}
+
+func inspectContainerOutputHandler(cmd command.Command) error {
+	var stdoutBuf bytes.Buffer
+	cmd.SetStdout(&stdoutBuf)
+
+	err := cmd.Run()
```
chews93319, line 96: Justin had advised that we could use `cmd.Output()` to capture the output for unmarshalling the json, rather than relying upon the interpretation of the stdout buffer.

Shubhranshu153: the string output from nerdctl cannot be unmarshalled as it is not a valid json. once we fix upstream this patch wont be required.

cmd/finch/nerdctl.go
Shubhranshu153: may be if required, this one will be removed once we fix it upstream.

chews93319 (approved):

cmd/finch/devcontainer_patch.go
```diff
@@ -0,0 +1,100 @@
...
+				arrays = append(arrays, input[start:i+1])
+			}
+		}
+	}
+
+	return arrays
+}
+
+func prettyPrintJSON(input string) {
+	jsonArrays := extractOuterArrays(input)
```
pendo324, line 44: Why can't we import the type from nerdctl and then unmarshall it?

Shubhranshu153: the output from nerdctl is of the form
```
[
container1 inspect
]
[

container 2 inspect
}
]
```

this is not a valid json, so first we tract the comma and convert it to a single valid json array that docker outputs, 

Post that i think it would be feasible so that the type cast can be avoided but the conversion has to be made prior.

pendo324: What if we append `--format '{{json .}}'` to the `container inspect` command?

chews93319 (dismissed):

chews93319 (approved):

PR #1133, opened 2024-10-07
feat: allow user configurable docker compose version env variable

Shubhranshu153:
Issue #, if available:

*Description of changes:*

Dev containers depends on docker compose version to name their applications. So sending nerdctl compose version breaks functionality

*Testing done:*
1. Run with dockercompat flag and env variable set
2. Run with dockercompat flag and no env variable set
3. Run without dockercompat flag. 

#### License Acceptance

### DIFF
--- a/cmd/finch/devcontainer_patch.go
+++ b/cmd/finch/devcontainer_patch.go
@@ -8,12 +8,14 @@ package main
 import (
 	"bytes"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"strings"
 
 	"github.com/containerd/nerdctl/v2/pkg/inspecttypes/dockercompat"
 	"github.com/docker/go-connections/nat"
 	"github.com/sirupsen/logrus"
+	"golang.org/x/exp/slices"
 
 	"github.com/runfinch/finch/pkg/command"
 )
@@ -107,4 +109,17 @@ func inspectContainerOutputHandler(cmd command.Command) error {
 	}
 	prettyPrintJSON(stdoutBuf.String())
 	return err
+}
+
+func handleDockerCompatComposeVersion(cmdName string, nc nerdctlCommand, runArgs []string) error {
+	if cmdName == "compose" && nc.fc.DockerCompat && slices.Contains(runArgs, "version") {
+		ver := nc.systemDeps.Env("DOCKER_COMPOSE_VERSION")
+		if ver != "" {
+			logrus.Warn("Displaying docker compose version set as environment variable DOCKER_COMPOSE_VERSION...")
+			fmt.Println(ver)
+			return nil
+		}
+		return errors.New("DOCKER_COMPOSE_VERSION environment variable is not set")
+	}
+	return errors.New("")
 }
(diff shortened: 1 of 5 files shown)


cmd/finch/devcontainer_patch.go
```diff
@@ -108,3 +109,14 @@ func inspectContainerOutputHandler(cmd command.Command) error {
 	prettyPrintJSON(stdoutBuf.String())
 	return err
 }
+
+func handleDockerCompatComposeVersion(cmd command.Command) error {
+	logrus.Warn("Displaying a corresponding docker version set as environment variable DOCKER_COMPOSE_VERSION...")
+	ver := os.Getenv("DOCKER_COMPOSE_VERSION")
```
pendo324, line 115: We should be using `stdlib.Env` function here for easier mocking

cmd/finch/nerdctl_remote.go
```diff
@@ -340,6 +340,11 @@ func (nc *nerdctlCommand) run(cmdName string, args []string) error {
 		return inspectContainerOutputHandler(cmd)
 	}
 
+	if cmdName == "compose" && nc.fc.DockerCompat && slices.Contains(runArgs, "version") {
+		cmd := nc.ncc.Create(runArgs...)
+		return handleDockerCompatComposeVersion(cmd)
+	}
+
```
pendo324, line 347: nit: extract this entire block into a a function and put it in `nerdctl.go`

cmd/finch/devcontainer_patch.go
```diff
@@ -108,3 +109,14 @@ func inspectContainerOutputHandler(cmd command.Command) error {
...
+	if ver != "" {
+		fmt.Println(ver)
+		return nil
+	}
+	logrus.Warn("Environment variable DOCKER_COMPOSE_VERSION not found falling back to finch compose version")
```
pendo324, line 120: What does this mean? We're using the default compose version? Do we have to set any fallback environment variable?

Shubhranshu153: not really we are just setting nerdctl/finch compose version

cmd/finch/devcontainer_patch.go
Shubhranshu153: ok will do that

cmd/finch/nerdctl_remote.go
Shubhranshu153: i thought about it but was a bit on the fence as this is more of logical thing but we can move it out.

Shubhranshu153: thinking about it, it creates the issue of return statements so if the above portion is true we would need to account the handling of return here vs later. that makes it need to add additional logic in the main code either way

cmd/finch/devcontainer_patch.go
Shubhranshu153: is the package named exactly like that, i dont seem to be able to import this package.

cmd/finch/nerdctl_remote.go
Shubhranshu153: nvm i can refactor that.

pendo324 (changes requested):
LGTM, you just need to regenerate the mocks for `NerdctlCommandSystemDeps`

pendo324 (approved):

PR #1146, opened 2024-10-22
ci: fix AL2 checkouts

pendo324:
Issue #, if available:

AL2 CI checkouts failing: https://github.com/runfinch/finch/actions/runs/11462038498/job/31899467633?pr=1144#step:3:942

*Description of changes:*
- Updated the logic to more closely match what's done in the official checkout action

*Testing done:*

#### License Acceptance

### DIFF
--- a/.github/workflows/e2e-linux.yaml
+++ b/.github/workflows/e2e-linux.yaml
@@ -45,9 +45,12 @@ jobs:
         if: ${{ (startsWith(inputs.os, 'amazon') && inputs.version == '2' ) }}
         run: |
           git clone https://github.com/${GITHUB_REPOSITORY}.git .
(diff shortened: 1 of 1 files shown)


.github/workflows/e2e-linux.yaml
austinvazquez, line 1: @pendo324 , any idea why now?

austinvazquez (approved):

.github/workflows/e2e-linux.yaml
pendo324: I'm not really sure why it happened. I think the hosts were replaced. My suspicion is that somehow there was config on those hosts which caused this to work previously. Really weird timing, and I'm not totally convinced by that. This new method should work going forward though. I pretty much copied if from the checkout action

Issue #516, opened 2023-08-04
Inconsistent messaging in `--help` output


RomainMuller:
**Describe the bug**

When I do `finch build --help`, the output ends with:
```
See also 'finch --help' for the global flags such as '--namespace', '--snapshotter', and '--cgroup-manager'.
```

However running `finch --help` does not mention the `--napespace`, `--snapshotter` and `--cgroup-manager` options, which is confusing:

```console
$ finch --help
Finch: open-source container development tool

Usage:
  finch [command]

Available Commands:
  build          Build an image from Dockerfile
  builder        Manage builds
  commit         Create a new image from a container's changes
  completion     Generate the autocompletion script for the specified shell
  compose        Compose
  container      Manage containers
  cp             Copy files/folders between a running container and the local filesystem
  create         Create a new container
  events         Get real time events from the server
  exec           Run a command in a running container
  help           Help about any command
  history        Show the history of an image
  image          Manage images
  images         List images
  info           Display system-wide information
  inspect        Return low-level information on Docker objects
  kill           Kill one or more running containers
  load           Load an image from a tar archive or STDIN
  login          Log in to a container registry
  logout         Log out from a container registry
  logs           Fetch the logs of a container
  network        Manage networks
  pause          Pause all processes within one or more containers
  port           List port mappings or a specific mapping for the container
  ps             List containers
  pull           Pull an image from a registry
  push           Push an image or a repository to a registry
  restart        Restart one or more containers
  rm             Remove one or more containers
  rmi            Remove one or more images
  run            Run a command in a new container
  save           Save one or more images to a tar archive (streamed to STDOUT by default)
  start          Start one or more stopped containers
  stats          Display a live stream of container(s) resource usage statistics
  stop           Stop one or more running containers
  support-bundle Support bundle management
  system         Manage containerd
  tag            Create a tag TARGET_IMAGE that refers to SOURCE_IMAGE
  top            Display the running processes of a container
  unpause        Unpause all processes within one or more containers
  update         Update configuration of one or more containers
  version        Shows Finch version information
  vm             Manage the virtual machine lifecycle
  volume         Manage volumes
  wait           Block until one or more containers stop, then print their exit codes

Flags:
      --debug     running under debug mode
  -h, --help      help for finch
  -v, --version   version for finch

Use "finch [command] --help" for more information about a command.
```

**Steps to reproduce**

```
$ finch build --help
$ finch --help
```

**Expected behavior**

[...]

PR #1157, opened 2024-10-29
fix: add persistent flags to root finch command [WIP]

coderbirju:
Issue #, if available:
#516 

*Description of changes:*

*Testing done:*

#### License Acceptance


chews93319 (dismissed):

cmd/finch/main_native.go
```diff
@@ -79,6 +79,9 @@ var newApp = func(
 	// TODO: Decide when to forward --debug to the dependencies
 	// (e.g. nerdctl for container commands and limactl for VM commands).
 	rootCmd.PersistentFlags().Bool("debug", false, "running under debug mode")
+	rootCmd.PersistentFlags().String("namespace", "", `containerd namespace, such as "moby" for Docker, "k8s.io" for Kubernetes`)
```
austinvazquez, line 82: It's a bit odd to reference these other namespaces instead of documenting the default "finch" namespace.

cmd/finch/main_native.go
```diff
@@ -79,6 +79,9 @@ var newApp = func(
...
+	rootCmd.PersistentFlags().String("cgroup-manager", "", `Cgroup manager to use ("cgroupfs"|"systemd")`)
```
austinvazquez, line 83: Do we actually support cgroupfs cgroup manager?

cmd/finch/main_native.go
```diff
@@ -79,6 +79,9 @@ var newApp = func(
...
+	rootCmd.PersistentFlags().String("snapshotter", "", "containerd snapshotter")
```
austinvazquez, line 84: Does this have precedent over the finch.yaml snapshotter configuration?

coderbirju: I think yes, according to the docs [here](https://github.com/runfinch/finch/blob/main/docs/snapshotters.md#snapshotters-option)
> The default snapshotter will be used whenever a snapshotter is needed and not specified for a command. Any other listed snapshotters can be used by specifying them with the --snapshotter flag

cmd/finch/main_native.go
coderbirju: > Do we actually support cgroupfs cgroup manager?

I was under the assumption that nerdctl handles this under the hood, I was just referencing [this](https://github.com/containerd/nerdctl/blob/cabd5bfef2f63ecf29d61a0fd50a740e5d359bc0/cmd/nerdctl/main.go#L120)

Apologies for the delayed response, I thought I replied here.

PR #1163, opened 2024-10-31
fix: add ci.yaml to exclude from ci-docs triggers

Shubhranshu153:
Issue #, if available:

*Description of changes:*
Add ci.yaml to exclude from ci-docs triggers

*Testing done:*

#### License Acceptance

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -12,6 +12,7 @@ on:
       - '.github/workflows/e2e-macos.yaml'
       - '.github/workflows/e2e-windows.yaml'
       - '.github/workflows/e2e-linux.yaml'
+      - '.github/workflows/ci.yaml'
       - 'contrib/packaging/**'
       - 'deps/**'
       - 'finch.yaml.d/**'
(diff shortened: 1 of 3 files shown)


pendo324:
Is the PR title wrong? I see more changes than just excluding ci.yaml in ci-docs

.github/workflows/ci.yaml
```diff
@@ -201,7 +203,36 @@ jobs:
       arch: ${{ matrix.arch }}
       version: ${{ matrix.version }}
       runner-type: ${{ matrix.runner-type }}
-
+    macos-unit-test-coverage:
+    runs-on: macos-latest
```
pendo324, line 207: We should just add a `unit-test-coverage` step to [the existing `unit-tests` workflow](https://github.com/runfinch/finch/blob/9c72ad1dcc7829b2421ab8de07024201057dab1b/.github/workflows/ci.yaml#L80), instead of creating an entirely new workflow.

We also don't even need to use multiple runners for this, we can use the default `ubuntu-latest` and set GOOS as needed, per platform, like the linting workflow

Shubhranshu153: noted, this i meant to try out things only intention was the exclude, but if we have it in the unit test flow we probably dont need it will cancel this one.

PR #1180, opened 2024-11-12
feat: make finch config available in VM at $HOME/.finch

Kern--:
Issue #, if available:
N/A

*Description of changes:*

NerdctlConfigApplier updates the user's bash rc in the VM to point DOCKER_CONFIG to the mounted .finch dir from the host OS. This works when the bash rc is loaded, but doesn't work for, e.g., systemd user services.

This change makes the finch config available available at $HOME/.finch in the VM so there is a clear place to load config via systemd user services without relying on any bash rcs.

*Testing done:*
Verified the updated unit test

#### License Acceptance


Shubhranshu153 (dismissed):

Shubhranshu153:
Unit test needs to be fixed, others are flaky, rerun will fix it.

chews93319 (approved):

PR #1181, opened 2024-11-13
feat: add finch-daemon

pendo324:
Issue #, if available:

## Description of changes

Adds finch-daemon, so that it starts inside the VM and is exposed to processes inside the VM (via `/run/finch.sock` and `/var/run/docker.sock`), and processes on the native macOS host (via `/path/to/lima/data/finch/sock/finch.sock`).

To make this possible, these changes had to be made:
1. Added finch-daemon specific logic to `./finch.yaml.d/mac.yaml`
    1. This logic makes it so that the `finch-daemon` binary and service file are copied into the VM, and then starts the daemon using systemd
1. In `Makefile.darwin`, changed the order so that the finch-daemon specific provisioning script would be the final script, after being merged
    1. Also added a `add-daemon-mount` target, which makes it so that duplicate mounts are not added to Lima. This is done to avoid the default lima provisioning scripts failing due to trying to mount/unmount a directory twice (the end result is users may see `cloud-final.service` fail after shelling into the VM. although this should only ever really impact developers, it was easy enough to fix)
1. Added finch@.service

This change also currently relies on a fork of finch-core, which actually has the code necessary for building finch-daemon. Once https://github.com/runfinch/finch-core/pull/424 is merged, I'll update this PR to remove the use of a different finch-core upstream.

### Testing done
- finch-daemon works as expected on macOS:
```
DOCKER_HOST=unix:///path/to/finch/_output/lima/data/finch/sock/finch.sock docker images
REPOSITORY                                                                   TAG                IMAGE ID       CREATED        SIZE
public.ecr.aws/amazonlinux/amazonlinux                                       2023               75763d26a280   3 weeks ago    51.4MB
public.ecr.aws/docker/library/registry                                       latest             12120425f07d   2 months ago   9.47MB
public.ecr.aws/lambda/nodejs                                                 18-x86_64          97f94536a3af   3 weeks ago    180MB
```

#### License Acceptance


pendo324:
Turns out `git.exe` isn't in the Windows runner's PATH. Added it here: https://github.com/runfinch/infrastructure/pull/769

After that gets deployed to the runners, it should resolve the Windows build failures we're seeing.

.gitmodules
```diff
@@ -1,3 +1,3 @@
 [submodule "finch-core"]
 	path = deps/finch-core
-	url = https://github.com/runfinch/finch-core.git
+	url = https://github.com/pendo324/finch-core.git
```
coderbirju, line 3: This points to your personal account, is this intentional?

coderbirju (approved):

finch@.service
```diff
@@ -0,0 +1,17 @@
+[Unit]
+Description=Finch daemon %I
+Documentation=https://runfinch.com https://builderhub.corp.amazon.com/docs/docker.html#finch https://github.com/runfinch/finch-daemon
+After=network.target local-fs.target
+
+[Service]
+ExecStart=/usr/local/bin/finch-daemon --socket-owner %i
+ExecStartPost=-rm -rf /var/run/docker.sock
+ExecStartPost=ln -s /run/finch.sock /var/run/docker.sock
```
Shubhranshu153, line 9: this i am not sure is a good idea by default, there might be case where customer wants to run both finch and docker, so any reference to docker doesnt seem correct.

pendo324: This is only inside the VM, which users should not be modifying

finch@.service
```diff
@@ -0,0 +1,17 @@
+[Unit]
+Description=Finch daemon %I
+Documentation=https://runfinch.com https://builderhub.corp.amazon.com/docs/docker.html#finch https://github.com/runfinch/finch-daemon
+After=network.target local-fs.target
```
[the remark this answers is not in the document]
pendo324: Added in latest revision

Shubhranshu153 (dismissed):

Makefile
```diff
@@ -75,7 +78,7 @@ endif
 
 FINCH_CORE_DIR := $(CURDIR)/deps/finch-core
 
-remote-all: arch-test finch install.finch-core-dependencies finch.yaml networks.yaml config.yaml
+remote-all: arch-test finch install.finch-core-dependencies finch.yaml networks.yaml config.yaml $(OUTDIR)/finch-daemon/finch@.service
```
swagatbora90, line 81: Do we really need separate targets for `networks.yaml`, `config.yaml`,  `$(OUTDIR)/finch-daemon/finch@.service`?

Makefile.darwin
```diff
@@ -18,9 +18,21 @@ FINCH_OS_IMAGE_LOCATION_ROOT ?= $(DEST)
 FINCH_IMAGE_LOCATION ?= $(FINCH_OS_IMAGE_LOCATION_ROOT)/os/$(FINCH_OS_BASENAME)
 FINCH_IMAGE_DIGEST ?= "sha512:$(FINCH_OS_DIGEST)"
 
+# check if finch-daemon socket is in a default path
+SHOULD_ADD_DAEMON_MOUNT = $(shell if [[ $(FINCH_DAEMON_LOCATION_ROOT) = ^\/Users\/.* ]]; then echo "0"; else echo "1"; fi)
```
Shubhranshu153, line 22: if its an install why dont we overwrite it?

finch.yaml.d/common.yaml
```diff
@@ -85,8 +85,13 @@ provision:
       printf '[Unit]\nDescription=Delete hanging data on boot\nDefaultDependencies=no\nBefore=basic.target\n\n[Service]\nType=oneshot\nExecStart=/bin/bash -c "sudo rm -rf /var/lib/cni/networks/bridge/**; sudo rm -rf /var/lib/cni/results/bridge-finch-*"\n\n[Install]\nWantedBy=basic.target\n' | sudo tee /usr/local/lib/systemd/system/finch-cleanup-on-boot.service
       sudo systemctl enable --now finch-cleanup-on-boot.service
 
-      sudo systemctl restart containerd.service
+      # Set a default ulimit for number of files in containerd
+      sudo mkdir -p /usr/local/lib/systemd/system/containerd.service.d/
+      printf '[Service]\nLimitNOFILE=1048576\n' | sudo tee /usr/local/lib/systemd/system/containerd.service.d/finch.conf
```
Shubhranshu153, line 90: kernel default seems to be 1048576
so can we set no limit and it auto sets it?

pendo324: systemd's limit is different:

```
systemctl show | grep LimitNOFILE
DefaultLimitNOFILE=524288
DefaultLimitNOFILESoft=1024
```

Makefile.darwin
pendo324: This also gets run when user's build development versions (likely in their user directory)

In this case, we don't want to add an extra mount, because it will cause problems when Lima tries to unmount (explained better in the PR description)

Makefile
pendo324: I think anything that writes a file should probably have its own rule so that its not re-created unless the file is deleted

Shubhranshu153 (approved):

finch@.service
```diff
@@ -0,0 +1,16 @@
+[Unit]
+Description=Finch daemon %I
+Documentation=https://runfinch.com https://github.com/runfinch/finch-daemon
+After=network.target local-fs.target containerd.service buildkit.service
```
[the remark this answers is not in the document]
swagatbora90: Not blocking. We can do it as a follow up.

swagatbora90 (approved):

Shubhranshu153 (approved):

pendo324:
Tracking finch-core from https://github.com/pendo324/finch-core/tree/with-updated-finch-daemon while finch-daemon changes are under review (https://github.com/runfinch/finch-daemon/pull/199 and https://github.com/runfinch/finch-daemon/pull/203)

finch@.service
Shubhranshu153: Not Blocking: 
Why do we need this?

Shubhranshu153 (approved):

Shubhranshu153 (approved):

Makefile.darwin
```diff
@@ -18,9 +18,21 @@ FINCH_OS_IMAGE_LOCATION_ROOT ?= $(DEST)
...
+# only add the finch-daemon mount when its not in a default path
+# this scenario is common in dev, where the typical path is /Users/...
+ifeq ($(SHOULD_ADD_DAEMON_MOUNT),1)
+finch.yaml: add-daemon-mount
+endif
+
 $(OS_OUTDIR)/finch.yaml: $(OS_OUTDIR) finch.yaml.d/common.yaml finch.yaml.d/mac.yaml
 	# merge the appropriate YAMLs into a temporary finch.yaml file on the current working directory
-	cd finch.yaml.d && yq eval-all '. as $$item ireduce ({}; . *+ $$item)' mac.yaml common.yaml > ../finch.yaml.temp
+	cd finch.yaml.d && yq eval-all '. as $$item ireduce ({}; . *+ $$item)' common.yaml mac.yaml > ./../finch.yaml.temp
```
chews93319, line 35: nit: (non-blocking)
Should this comment emphasize the significance of "mac.yaml" values will overwrite "common.yaml" due to order of evaluation?

chews93319 (dismissed):

chews93319:
From exploring the CI / macos-e2e-tests, it seems like neither of the [finch-daemon network_create.go tests L188-L234](https://github.com/runfinch/finch-daemon/blob/460b73f5872c7579be20572dde98bfd69ee9af14/e2e/tests/network_create.go#L188-L234) are invoking `iptables` quite right.

`enable_icc set to true` test case expects an unsuccessful exit, but perhaps it is unsuccessful for a reason other than what we are hoping for...

`enable_icc set to false` test case expects a successful exit, but an invocation option seems to be incorrect/invalid or `iptables` itself has an issue. Is it possible that the runner has an older version of `iptables`, which does not support `-C`?

pendo324:
> Should this PR be held for these tests?
Yeah, probably. Good catch. The weirdest part is that the tests succeed, but only sometimes. I haven't been able to get to the root cause because I haven't been able to make them fail on my local machine.

Do you mind creating the PRs to fix the tests in the finch-daemon repo since you've already got a good handle on some fixes?

chews93319:
> Do you mind creating the PRs to fix the tests in the finch-daemon repo since you've already got a good handle on some fixes?

Yeah, I can try to do that to unblock this PR.

Shubhranshu153 (dismissed):

pendo324:
CI looks good with the new finch-daemon version coming in from the latest merge from main!

I'm going to manually run the installer/nightly tests against this branch, since last time I did that, I actually found some issues.

pendo324:
`OUTDIR=/Applications/Finch make test-e2e-daemon` tests ran successfully after I installed an installer built from this commit 🎊

chews93319 (dismissed):

Makefile
```diff
@@ -303,6 +309,32 @@ test-e2e-container: create-report-dir
...
+
+ifeq ($(INSTALLED),true)
+DAEMON_DOCKER_HOST := "unix:///Applications/Finch/lima/data/finch/sock/finch.sock"
+else
+DAEMON_DOCKER_HOST := "unix://$(OUTDIR)/lima/data/finch/sock/finch.sock"
+endif
+
+.PHONY: test-e2e-daemon
+test-e2e-daemon:
+	-@$(OUTDIR)/bin/$(BINARYNAME) vm stop -f || true
```
Shubhranshu153, line 324: does the make file need to ensure finch is running or test runner should ensure.

swagatbora90 (approved):

finch.yaml.d/common.yaml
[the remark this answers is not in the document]
pendo324: Removed

Makefile
pendo324: I didn't add it as a workflow target, like our other tests. Technically I can move the logic to a workflow though, you're right. Can be done in a follow up

Makefile.darwin
pendo324: Yeah, the comment can be improved. Will do in a follow up PR

PR #1183, opened 2024-11-14
ci: use setup-go for Windows and hardcode GO_VERSION

pendo324:
Issue #, if available:

## Description of changes
- Adds actions/setup-go to Windows workflows
- Does not use `check-latest: true` because our dependencies require 1.22.0, and go considers 1.22 (without the patch version) to not satisfy that criteria. We will have to periodically manually update `GO_VERSION` to align with new Go releases.

## Testing done
None, need to test this on the runners themselves.

## Contribution agreement

## License Acceptance

### DIFF
--- a/.github/workflows/build-and-test-msi.yaml
+++ b/.github/workflows/build-and-test-msi.yaml
@@ -16,6 +16,7 @@ on:
     - cron: '0 9 * * *'
 env:
   GO111MODULE: on
+  GO_VERSION: '1.22.9'
 
 permissions:
   # This is required for configure-aws-credentials to request an OIDC JWT ID token to access AWS resources later on.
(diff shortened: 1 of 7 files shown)


Shubhranshu153 (approved):

austinvazquez:
Curious if the enable cache change is intentional? For macOS e2e testing workflows cache writes takes 6-7 minutes due to cache size. These are also evicted quite quickly as we only have the default 10GB GitHub cache.

pendo324:
> Curious if the enable cache change is intentional? For macOS e2e testing workflows cache writes takes 6-7 minutes due to cache size. These are also evicted quite quickly as we only have the default 10GB GitHub cache.

I thought that issue was resolved upstream, but I see they only have a workaround for Windows, not macOS. I'll make a new revision with cahceing disabled.

austinvazquez (approved):
Glad to hear upstream is working on it. LGTM.

PR #1186, opened 2024-11-15
chore: add longform boolean unit tests

coderbirju:
Issue #, if available:

*Description of changes:*

*Testing done:*

#### License Acceptance

### DIFF
--- a/cmd/finch/nerdctl_darwin_test.go
+++ b/cmd/finch/nerdctl_darwin_test.go
@@ -1171,6 +1171,41 @@ func TestNerdctlCommand_run(t *testing.T) {
 				c.EXPECT().Run()
 			},
 		},
+		{
+			name:    "with long-form boolean flags",
+			cmdName: "run",
+			fc:      &config.Finch{},
+			args: []string{
+				"--env", "ARG1=val1", "-p", "8080:8080",
+				"--name", "myContainer", "--interactive=true", "--detach", "--rm=true", "--init=false",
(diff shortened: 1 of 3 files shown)


cmd/finch/nerdctl_darwin_test.go
```diff
@@ -1171,6 +1171,37 @@ func TestNerdctlCommand_run(t *testing.T) {
 				c.EXPECT().Run()
 			},
 		},
+		{
```
Shubhranshu153, line 1174: we should include all long form flags supported by nerdctl in the check, irrespective of the ones present in our dictionary.

coderbirju:
```
--debug                   
--debug-full
--experimental
--insecure-registry
--attach
--detach
--interactive
--oom-kill-disable
--privileged
--read-only
--init
--rm
--sig-proxy
--tty

``` 
These are the long form boolean commands available on nerdctl. (` nerdctl run --help | grep -- '--' | grep -v -E 'string|int|float|uint|uint16' | grep -v '=' `)

I have included everything except
```
--debug - since debug-full is already present
--insecure-registry - this is a pull specific flag
--attach - contradicts detach
```

It is important to note that certain global commands ``--experimental`` and ``--debug-full`` result in the boolean value being sent as argument instead of an equality sign (e.g ``--debug-full false``), it is valid nerdctl command though. Just something to note.

PR #1200, opened 2024-11-29
feat: refactor cli parsing

Shubhranshu153:
Issue #, if available:

*Description of changes:*

*Testing done:*

- [ ] I've reviewed the guidance in CONTRIBUTING.md

#### License Acceptance

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #1251, opened 2025-01-21
fix: conflict in parsing of -h flag for help and hostname in container run 

Shubhranshu153:
…r run

Issue #, if available:

*Description of changes:*
1. Fix  conflict in parsing of -h flag for help and hostname in container
2. Fix invalid ptr access error at different places of the parsing logic

*Testing done:*
1. Run with -h flag and check --help works
3. finch run -it --name edge -h edge alpine sh
4. finch compose -h
5. finch compose run --help

#### License Acceptance

### DIFF
--- a/cmd/finch/nerdctl.go
+++ b/cmd/finch/nerdctl.go
@@ -110,8 +110,10 @@ func (nc *nerdctlCommand) shouldReplaceForHelp(cmdName string, args []string) bo
 		}
 	}
 
+	// this needs to handle cases of -h except for `container run`,`run`,`create`. For these options -h represent hostname
+	// TODO: Refactor this function.
 	for _, arg := range args {
-		if arg == "--help" || arg == "-h" {
+		if arg == "--help" || (arg == "-h" && !(cmdName == "container run" || cmdName == "run" || cmdName == "create")) {
 			return true
 		}
 	}
--- a/cmd/finch/nerdctl_remote.go
+++ b/cmd/finch/nerdctl_remote.go
@@ -141,10 +141,12 @@ func (nc *nerdctlCommand) run(cmdName string, args []string) error {
 			case arg == "--help":
 				nerdctlArgs = append(nerdctlArgs, arg)
 			case arg == "--add-host":
-				// exact match to --add-host
-				args[i+1], err = resolveIP(args[i+1], nc.logger, nc.ecc)
-				if err != nil {
-					return err
+				// exact match to --add-host. resolve ip if param passed
+				if len(args) > i+1 {
+					args[i+1], err = resolveIP(args[i+1], nc.logger, nc.ecc)
+					if err != nil {
+						return err
+					}
 				}
 				nerdctlArgs = append(nerdctlArgs, arg)
 			case strings.HasPrefix(arg, "--add-host"):
@@ -158,21 +160,33 @@ func (nc *nerdctlCommand) run(cmdName string, args []string) error {
 			case strings.HasPrefix(arg, "--env-file"):
 				// exact match to --env-file
 				// or arg begins with --env-file
-				shouldSkip, addEnvs, err := handleEnvFile(nc.fs, nc.systemDeps, arg, args[i+1])
-				if err != nil {
-					return err
+				if len(args) > i+1 {
+					shouldSkip, addEnvs, err := handleEnvFile(nc.fs, nc.systemDeps, arg, args[i+1])
+					if err != nil {
+						return err
+					}
+					skip = shouldSkip
+					fileEnvs = append(fileEnvs, addEnvs...)
+				} else {
+					// if --env-file is at the end of the args, its refers to entrypoint command
+					// which need not be handled
+					nerdctlArgs = append(nerdctlArgs, arg)
 				}
-				skip = shouldSkip
-				fileEnvs = append(fileEnvs, addEnvs...)
 			case argIsEnv(arg):
 				// exact match to either -e or --env
 				// or arg begins with -e or --env
 				//     -e="<value>", -e"<value>"
 				//     --env="<key>=<value>", --env"<key>=<value>"
-				shouldSkip, addEnv := handleEnv(nc.systemDeps, arg, args[i+1])
-				skip = shouldSkip
-				if addEnv != "" {
-					envs = append(envs, addEnv)
+				if len(args) > i+1 {
+					shouldSkip, addEnv := handleEnv(nc.systemDeps, arg, args[i+1])
+					skip = shouldSkip
+					if addEnv != "" {
+						envs = append(envs, addEnv)
+					}
+				} else {
+					// if -e or --env is at the end of the args, its refers to entrypoint command
+					// which need not be handled
+					nerdctlArgs = append(nerdctlArgs, arg)
 				}
 			case shortFlagBoolSet.Has(arg) || longFlagBoolSet.Has(arg):
 				// exact match to a short no argument flag: -?
@@ -195,11 +209,17 @@ func (nc *nerdctlCommand) run(cmdName string, args []string) error {
 				// or begins with a short arg flag:
 				//     short arg flag concatenated to value: -?"<value>"
 				//     short arg flag equated to value: -?="<value>" or -?=<value>
-				shouldSkip, addKey, addVal := nc.handleFlagArg(arg, args[i+1])
-				skip = shouldSkip
-				if addKey != "" {
-					nerdctlArgs = append(nerdctlArgs, addKey)
-					nerdctlArgs = append(nerdctlArgs, addVal)
+				if len(args) > i+1 {
+					shouldSkip, addKey, addVal := nc.handleFlagArg(arg, args[i+1])
+					skip = shouldSkip
+					if addKey != "" {
+						nerdctlArgs = append(nerdctlArgs, addKey)
+						nerdctlArgs = append(nerdctlArgs, addVal)
+					}
+				} else {
+					// no value found for short arg flag
+					// pass the arg as a nerdctl command argument
+					nerdctlArgs = append(nerdctlArgs, arg)
 				}
 			case strings.HasPrefix(arg, "--"):
 				// exact match to a long arg flag: -<long_flag>
@@ -207,11 +227,17 @@ func (nc *nerdctlCommand) run(cmdName string, args []string) error {
 				// or begins with a long arg flag:
 				//     long arg flag concatenated to value: --<long_flag>"<value>"
 				//     long arg flag equated to value: --<long_flag>="<value>" or --<long_flag>=<value>
-				shouldSkip, addKey, addVal := nc.handleFlagArg(arg, args[i+1])
-				skip = shouldSkip
-				if addKey != "" {
-					nerdctlArgs = append(nerdctlArgs, addKey)
-					nerdctlArgs = append(nerdctlArgs, addVal)
+				if len(args) > i+1 {
+					shouldSkip, addKey, addVal := nc.handleFlagArg(arg, args[i+1])
+					skip = shouldSkip
+					if addKey != "" {
+						nerdctlArgs = append(nerdctlArgs, addKey)
+						nerdctlArgs = append(nerdctlArgs, addVal)
+					}
+				} else {
+					// if --<long flag> is at the end of the args, its refers to entrypoint command
+					// which need not be handled
+					nerdctlArgs = append(nerdctlArgs, arg)
 				}
 			default:
 				// arg other than a flag ("-?","--<long_flag>") or a skipped <flag_value>
--- a/cmd/finch/nerdctl_test.go
+++ b/cmd/finch/nerdctl_test.go
@@ -30,48 +30,76 @@ func TestNerdctlCommand_shouldReplaceForHelp(t *testing.T) {
 	t.Parallel()
 
 	testCases := []struct {
-		name    string
-		cmdName string
-		args    []string
-		mockSvc func(*mocks.NerdctlCmdCreator, *mocks.Logger, *gomock.Controller)
+		name     string
+		cmdName  string
+		args     []string
+		expected bool
+		mockSvc  func(*mocks.NerdctlCmdCreator, *mocks.Logger, *gomock.Controller)
 	}{
 		{
-			name:    "with --help flag",
-			cmdName: "pull",
-			args:    []string{"test:tag", "--help"},
+			name:     "with --help flag",
+			cmdName:  "pull",
+			args:     []string{"test:tag", "--help"},
+			expected: true,
 		},
 		{
-			name:    "with -h",
-			cmdName: "pull",
-			args:    []string{"test:tag", "-h"},
+			name:     "with -h",
+			cmdName:  "pull",
+			args:     []string{"test:tag", "-h"},
+			expected: true,
 		},
 		{
-			name:    "system returns help",
-			cmdName: "system",
+			name:     "system returns help",
+			cmdName:  "system",
+			expected: true,
 		},
 		{
-			name:    "builder returns help",
-			cmdName: "builder",
+			name:     "builder returns help",
+			cmdName:  "builder",
+			expected: true,
 		},
 		{
-			name:    "container returns help",
-			cmdName: "container",
+			name:     "container returns help",
+			cmdName:  "container",
+			expected: true,
 		},
 		{
-			name:    "image returns help",
-			cmdName: "image",
+			name:     "image returns help",
+			cmdName:  "image",
+			expected: true,
 		},
 		{
-			name:    "network returns help",
-			cmdName: "network",
+			name:     "network returns help",
+			cmdName:  "network",
+			expected: true,
 		},
 		{
-			name:    "volume returns help",
-			cmdName: "volume",
+			name:     "volume returns help",
+			cmdName:  "volume",
+			expected: true,
 		},
 		{
-			name:    "compose returns help",
-			cmdName: "compose",
+			name:     "compose returns help",
+			cmdName:  "compose",
+			expected: true,
+		},
+		{
+			name:     "-h argument for hostname-related command",
+			cmdName:  "run",
+			args:     []string{"-h"},
+			expected: false,
+		},
+		{
+			name:     "-h argument for hostname-related command",
+			cmdName:  "container run",
+			args:     []string{"-h"},
+			expected: false,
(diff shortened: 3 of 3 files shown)


cmd/finch/nerdctl_remote.go
```diff
@@ -142,11 +142,17 @@ func (nc *nerdctlCommand) run(cmdName string, args []string) error {
 				nerdctlArgs = append(nerdctlArgs, arg)
 			case arg == "--add-host":
 				// exact match to --add-host
-				args[i+1], err = resolveIP(args[i+1], nc.logger, nc.ecc)
-				if err != nil {
-					return err
+				if len(args) > i+1 {
```
chews93319, line 145: The comment added within the else statement was useful for clarification.
We can add the "inverse note" within the if clause:

```
Found an arg after 'add-host', so passing to resolveIP to validate/update it as a valid value
```

cmd/finch/nerdctl_remote.go
```diff
@@ -158,21 +164,33 @@ func (nc *nerdctlCommand) run(cmdName string, args []string) error {
...
-				skip = shouldSkip
-				if addEnv != "" {
-					envs = append(envs, addEnv)
+				if len(args) > i+1 {
+					shouldSkip, addEnv := handleEnv(nc.systemDeps, arg, args[i+1])
+					skip = shouldSkip
+					if addEnv != "" {
+						envs = append(envs, addEnv)
+					}
+				} else {
```
chews93319, line 190: This is a valid else condition because it indicates that:
 - ` arg begins with -e or --env, like -e="<value>", -e"<value>" or  --env="<key>=<value>", --env"<key>=<value>"`

As such, the `arg[i]` must be split and processed into `Env` values, which are in line467-485 of `handleEnv`

This else condition should change to:
```
// No arg after `e`/`env`, so passing current arg to handleEnvFile to validate/process it as a valid value

shouldSkip, addEnv := handleEnv(nc.systemDeps, arg, '')
```

-------------

Alternatively, the splitting logic could be refactored to here and a simplified version of `handleEnv` could be invoked.

cmd/finch/nerdctl_remote.go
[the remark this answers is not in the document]
Shubhranshu153: There is an open pr to entirely refactor the flow that would make lot of this redundant. Would add a unit test but don't think the e2e test is necessary.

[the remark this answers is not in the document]
Shubhranshu153: https://github.com/runfinch/finch/pull/1200

cmd/finch/nerdctl_remote.go
```diff
@@ -195,23 +213,35 @@ func (nc *nerdctlCommand) run(cmdName string, args []string) error {
...
-					nerdctlArgs = append(nerdctlArgs, addKey)
-					nerdctlArgs = append(nerdctlArgs, addVal)
+				if len(args) > i+1 {
+					shouldSkip, addKey, addVal := nc.handleFlagArg(arg, args[i+1])
+					skip = shouldSkip
+					if addKey != "" {
+						nerdctlArgs = append(nerdctlArgs, addKey)
+						nerdctlArgs = append(nerdctlArgs, addVal)
+					}
+				} else {
```
chews93319, line 223: If and only if all cases of Finch responsibility are addressed prior to this case, then the "simple" or "complex" argument can be passed directly into `nerdctlArgs`.

cmd/finch/nerdctl_remote.go
```diff
@@ -195,23 +213,35 @@ func (nc *nerdctlCommand) run(cmdName string, args []string) error {
 				// or begins with a short arg flag:
 				//     short arg flag concatenated to value: -?"<value>"
 				//     short arg flag equated to value: -?="<value>" or -?=<value>
-				shouldSkip, addKey, addVal := nc.handleFlagArg(arg, args[i+1])
-				skip = shouldSkip
-				if addKey != "" {
-					nerdctlArgs = append(nerdctlArgs, addKey)
-					nerdctlArgs = append(nerdctlArgs, addVal)
+				if len(args) > i+1 {
+					shouldSkip, addKey, addVal := nc.handleFlagArg(arg, args[i+1])
```
chews93319, line 217: The `handleFlagArg` method is the purest form of "argument separation", since it does not seem to be doing anything "Finch specific" (Refer to line428).

In the effort to remove null pointers, the if/else conditions would split to:
```
if len(args) > i+1 {
   shouldSkip, addKey, addVal := nc.handleFlagArg(arg, args[i+1])
} else {
   shouldSkip, addKey, addVal := nc.handleFlagArg(arg, "")
}
```

-----------

As a candidate for "pass-through to nerdctl", then this case block could be simplified to leave responsibility to `nerdctl`

```
if len(args) > i+1 {
   nerdctlArgs = append(nerdctlArgs, arg)
   nerdctlArgs = append(nerdctlArgs, arg[i+1])
   skip = true
} else {
   nerdctlArgs = append(nerdctlArgs, arg)
}
```

cmd/finch/nerdctl_remote.go
```diff
@@ -195,23 +213,35 @@ func (nc *nerdctlCommand) run(cmdName string, args []string) error {
...
 				//     next arg must be the <value>
 				// or begins with a long arg flag:
 				//     long arg flag concatenated to value: --<long_flag>"<value>"
 				//     long arg flag equated to value: --<long_flag>="<value>" or --<long_flag>=<value>
-				shouldSkip, addKey, addVal := nc.handleFlagArg(arg, args[i+1])
-				skip = shouldSkip
-				if addKey != "" {
-					nerdctlArgs = append(nerdctlArgs, addKey)
-					nerdctlArgs = append(nerdctlArgs, addVal)
+				if len(args) > i+1 {
```
chews93319, line 234: Similar to the "general short flag" case above, this is the "general long flag" case.

The `handleFlagArg` method is the purest form of "argument separation", since it does not seem to be doing anything "Finch specific" (Refer to line428).

In the effort to remove null pointers, the if/else conditions would split to:
```
if len(args) > i+1 {
   shouldSkip, addKey, addVal := nc.handleFlagArg(arg, args[i+1])
   ....
} else {
   shouldSkip, addKey, addVal := nc.handleFlagArg(arg, "")
   ....
}
```

-----------

As a candidate for "pass-through to nerdctl", then this case block could be simplified to leave responsibility to `nerdctl`

```
if len(args) > i+1 {
   nerdctlArgs = append(nerdctlArgs, arg)
   nerdctlArgs = append(nerdctlArgs, arg[i+1])
   skip = true
} else {
   nerdctlArgs = append(nerdctlArgs, arg)
}
```

cmd/finch/nerdctl.go
```diff
@@ -110,8 +110,9 @@ func (nc *nerdctlCommand) shouldReplaceForHelp(cmdName string, args []string) bo
 		}
 	}
 
+	// this only needs to handle `container run`.
 	for _, arg := range args {
-		if arg == "--help" || arg == "-h" {
+		if arg == "--help" || (arg == "-h" && !strings.Contains(cmdName, "run")) {
```
chews93319, line 115: The `-h` short flag conflict exists in our runfinch.com CLI Reference for both `run` and `create`.
However, everywhere else `-h, --help` is standard.

If we proceed with this modification, then please add the "not create" condition as well. 

------

I am concerned that overriding `-h` for only `run` and `create` would get lost in "future maintenance".
Perhaps the right mechanism would be to commit to either: 
 - reducing help to long flag only
 - committing hostname to long flag only

cmd/finch/nerdctl_remote.go
```diff
@@ -142,11 +142,17 @@ func (nc *nerdctlCommand) run(cmdName string, args []string) error {
...
-				args[i+1], err = resolveIP(args[i+1], nc.logger, nc.ecc)
-				if err != nil {
-					return err
+				if len(args) > i+1 {
+					args[i+1], err = resolveIP(args[i+1], nc.logger, nc.ecc)
+					if err != nil {
+						return err
+					}
+					nerdctlArgs = append(nerdctlArgs, arg)
+				} else {
```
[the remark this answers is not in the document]
Shubhranshu153: we need not handle err nerdctl will complain if something is wrong.
Finch cli will act as pass through as possible

cmd/finch/nerdctl_remote.go
Shubhranshu153: we need not this len is just taking care of cases to ensure the case where we have
finch command -<arg>
so as to not enter invalid pointer logic.

cmd/finch/nerdctl.go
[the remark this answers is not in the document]
Shubhranshu153: agreed

Shubhranshu153: good catch. Yes create case needs to be handled as well .

Shubhranshu153: both the suggestions are restrictive. 
We probably refactor the help replace part as part of the refactor code. 
Idea is we donot try to get in the way of nerdctl. Finch being as dumb as possible is the best way for us.

cmd/finch/nerdctl_remote.go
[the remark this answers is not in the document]
Shubhranshu153: @pendo324  yes that should be possible will fix that.

cmd/finch/nerdctl_remote.go
```diff
@@ -158,21 +164,33 @@ func (nc *nerdctlCommand) run(cmdName string, args []string) error {
...
-				if err != nil {
-					return err
+				if len(args) > i+1 {
+					shouldSkip, addEnvs, err := handleEnvFile(nc.fs, nc.systemDeps, arg, args[i+1])
+					if err != nil {
+						return err
+					}
+					skip = shouldSkip
+					fileEnvs = append(fileEnvs, addEnvs...)
+				} else {
```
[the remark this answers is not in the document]
Shubhranshu153: it cannot really happen --env-file is the last argument.
once case can be where we have is after the image arg, in that case it should be passed without any processing any way.

cmd/finch/nerdctl_remote.go
Shubhranshu153: it cannot really happen --env-file is the last argument.
once case can be where we have is after the image arg, in that case it should be passed without any processing any way.

cmd/finch/nerdctl_remote.go
Shubhranshu153: this case only happens if its the last param without any following arg, which in any case should not be handled 
am i missing some logic here  in the sense that the handle flag arg doesnt.

adding a else here is just escaping null reference the code is not changing any logic here. if its advartently changing then lets discuss.  the bug might be somewhere else.

cmd/finch/nerdctl_remote.go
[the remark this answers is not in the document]
chews93319: The value must be passed through `resolveIP`, which is Finch's responsibility prior to passing the values into the VM. The value must be manipulated prior to sending to nerdctl.

cmd/finch/nerdctl_remote.go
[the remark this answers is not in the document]
chews93319: That's a great clarification; thanks!
In fact, the comment for this else conditional should reflect that implication.

```
// The last arg has a prefix of '--env-file=', which must be an image argument
// and should be passed directly to nerdctl as a command argument.
```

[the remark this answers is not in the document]
Shubhranshu153: will add the comment.

cmd/finch/nerdctl_remote.go
chews93319: (Similar to above...)  That's a great clarification; thanks!
The comment for this else conditional should reflect that implication.

```
// The last arg has a prefix of '--env-file=', which must be an image argument
// and should be passed directly to nerdctl as a command argument.
```

--------------

What would the implications be for nerdctl to receive an environment variable file as an image argument?
Does that mean the image would have required the file to be loaded into the vm file system?

cmd/finch/nerdctl_remote.go
Shubhranshu153: not really, so here it means its the last argument which we pass it to nerdctl to handle as explained in the above case the reasoning behind it.

cmd/finch/nerdctl_remote.go
Shubhranshu153: not sure i understand why the current implementation won't work.

cmd/finch/nerdctl_remote.go
chews93319: ( sorry for poor text communication.... )

I agree with the intent of this change: the last argument really shouldn't be manipulated and should be passed to nerdctl.

This was a discussion comment (not intended to block approval) and I was trying to offer feedback options of:
 - a "consistent" change that **_only_** addresses the nullptr issue
 - an incremental removal/refactor step.

cmd/finch/nerdctl.go
chews93319: unfortunately, the `help` and the `hostname` must be Finch's responsibility for "remote" (vm) use cases due to the IP resolution manipulation.

cmd/finch/nerdctl_remote.go
Shubhranshu153: all these options will go away in the refactor pr so i thought not to further refactor these.

cmd/finch/nerdctl_remote.go
chews93319: Perhaps I am misunderstanding something, but I currently understand this as Finch's responsibility for the "remote" (VM) use cases.

cmd/finch/nerdctl_remote.go
Shubhranshu153: no it means the [cmd] has env
finch run -d alpine:latest python3 test.py --env-file=test.env

cmd/finch/nerdctl.go
Shubhranshu153: due to the IP resolution manipulation.? is for --add-host.
Not sure i get this point.

cmd/finch/nerdctl_remote.go
Shubhranshu153: finch will communicate the error from nerdctl. it acts as a bridge to enable running nerdctl inside a remote vm. but it doesnt need to handle the error that are from nerdctl. it can be transperant to that.

cmd/finch/nerdctl.go
chews93319: Sorry, yes.... I was "crossing" the topic of `add-host` and `hostname`.
Only catching the Finch `help` is our responsibility.

cmd/finch/nerdctl_remote.go
chews93319: Finch is responsible for executing the `resolveIP` method on the value being passed into the VM.
 - For Windows: nerdctl_windows.go#L343
 - For MacOS: nerdctl_darwin.go#L33
 
Hence my suggestion to iterate on the comment.

cmd/finch/nerdctl_remote.go
chews93319: For the given case:
> finch run -d alpine:latest python3 test.py --env-file=test.env

I understand this to mean that there will be an existing environment file `test.env` within the VM mounted path, which `nerdctl` would pass into the container for processing.

Is my understanding correct?

cmd/finch/nerdctl_remote.go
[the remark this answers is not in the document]
chews93319: Okay, I am convinced that we can rely upon the nerdctl error response when `--add-host` is the last argument position rather than finch trying to detect and raise the error.

cmd/finch/nerdctl_remote.go
chews93319: Also, I am convinced that we can rely upon the nerdctl error response when `--add-host` is the last argument position rather than finch trying to detect and raise the error.

cmd/finch/nerdctl_remote.go
chews93319: My suggestion was merely a style/non-functional feedback.
It is non-blocking for this pr.

cmd/finch/nerdctl_remote.go
chews93319: Okay, I am convinced that we can rely upon the nerdctl error response when `--add-host` is the last argument position rather than finch trying to detect and raise the error.

cmd/finch/nerdctl_remote.go
chews93319: My suggestion was merely a style/non-functional feedback.
It is non-blocking for this pr.

Shubhranshu153: it will be a throw away refactor so will skip it. 
the ptr check is functionally relevant so i think its worth doing a patch

cmd/finch/nerdctl_remote.go
Shubhranshu153: it will be a throw away refactor so will skip it.
the ptr check is functionally relevant so i think its worth doing a patch

cmd/finch/nerdctl_remote.go
Shubhranshu153: test.env here is a env file inside the container, nerdctl or finch neednot worry about it. 
if its something that is mounted by user, user is responsible for creating it and finch merely mounts it in the vm

cmd/finch/nerdctl_remote.go
```diff
@@ -158,21 +164,33 @@ func (nc *nerdctlCommand) run(cmdName string, args []string) error {
 			case strings.HasPrefix(arg, "--env-file"):
 				// exact match to --env-file
 				// or arg begins with --env-file
-				shouldSkip, addEnvs, err := handleEnvFile(nc.fs, nc.systemDeps, arg, args[i+1])
-				if err != nil {
-					return err
+				if len(args) > i+1 {
```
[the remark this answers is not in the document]
Shubhranshu153: on second thought we neednot add it as all this going to become redundant with the refactor.

cmd/finch/nerdctl_remote.go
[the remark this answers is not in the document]
Shubhranshu153: same response as above.

chews93319 (approved):

pendo324 (approved):

Issue #800, opened 2024-02-08
Support BYO credential helpers


ollypom:
**What is the problem you're trying to solve?.**
Today Finch only "natively" supports the [ecr-login credential helper](https://github.com/awslabs/amazon-ecr-credential-helper) https://github.com/runfinch/finch/pull/462. However there are scenarios where folks may want to use other credential helpers or wrap `ecr-login`. 

Today you can put your own helpers in `~/.finch/cred-helpers` and update your DOCKER_CONFIG `config.json` and everything will be successfully passed into the lima VM, however the BYO credential helper would not be in your path.

```
$ cat .finch/config.json
{
        "credsStore": "ecr-login-new"
}

$ ls -l .finch/cred-helpers
total 6512
-rwxrwxrwx 1 root root 6668288 Feb  1 13:42 docker-credential-ecr-login
-rwxrwxrwx 1 root root      42 Feb  8 11:24 docker-credential-ecr-login-new

$ finch run 111222333444.dkr.ecr.eu-west-1.amazonaws.com/nginx:latest
WARN[0000] cannot get auth config for authConfigHostname="111222333444.dkr.ecr.eu-west-1.amazonaws.com" (refHostname="111222333444.dkr.ecr.eu-west-1.amazonaws.com")  error="error getting credentials - err: exec: \"docker-credential-ecr-login-new\": executable file not found in $PATH, out: ``"
```

Today we create a symlink explicitly for the `ecr-login` provider so that's its in your path.

https://github.com/runfinch/finch/blob/52fb8b2fe5fb4523076a4bfbd2da8861df120adf/pkg/config/nerdctl_config_applier.go#L87-L92

I wonder if we could instead add `"$FINCH_DIR"/cred-helpers/` to the users path. Therefore all credentials providers could be picked up 🤔 

**Describe the feature you'd like**
Support BYO credential providers / wrapper scripts.

**Additional context**

PR #1265, opened 2025-01-31
feat: allow other credential helpers in Finch VM

austinvazquez:
Issue #, if available:

Addresses https://github.com/runfinch/finch/issues/800

*Description of changes:*
This change updates nerdctl commands to have other credential helpers on the path for nerdctl commands.

*Testing done:*
1. Added `creds_helpers: - "ecr-login"` to `_output/config.yaml`
2. Put a second credential helper in cred-helpers directory.
3. Started the virtual machine.
```
_output/bin/finch vm init
LIMA_HOME=$(pwd)/_output/lima/data _output/bin/lima/limactl shell finch
```
4. Verified both credential helpers were on the path and linked to `/usr/local/bin`

#### License Acceptance


pkg/config/nerdctl_config_applier.go
```diff
@@ -91,8 +91,8 @@ func addLineToBashrc(fs afero.Fs, profileFilePath string, profStr string, cmd st
 func updateEnvironment(fs afero.Fs, fc *Finch, finchDir, homeDir, limaVMHomeDir string) error {
 	cmdArr := []string{
 		`export DOCKER_CONFIG="$FINCH_DIR"`,
-		"[ -L /usr/local/bin/docker-credential-ecr-login ] " +
-			`|| sudo ln -s "$FINCH_DIR"/cred-helpers/docker-credential-ecr-login /usr/local/bin/`,
+		`for f in "$FINCH_DIR"/cred-helpers/*; do ` +
+			`[ -L "/usr/local/bin/$(basename "$f")" ] || sudo ln -s "$f" "/usr/local/bin/$(basename "$f")"; done`,
```
pendo324, line 95: Any reason why it isn't more like this? The quoting/unquoting seems odd to me
```suggestion
			`[ -L "/usr/local/bin/$(basename ${f})" ] || sudo ln -s "${f}" "/usr/local/bin/$(basename ${f}"); done`,
```

pkg/config/nerdctl_config_applier.go
```diff
@@ -91,8 +91,8 @@ func addLineToBashrc(fs afero.Fs, profileFilePath string, profStr string, cmd st
 func updateEnvironment(fs afero.Fs, fc *Finch, finchDir, homeDir, limaVMHomeDir string) error {
 	cmdArr := []string{
 		`export DOCKER_CONFIG="$FINCH_DIR"`,
-		"[ -L /usr/local/bin/docker-credential-ecr-login ] " +
-			`|| sudo ln -s "$FINCH_DIR"/cred-helpers/docker-credential-ecr-login /usr/local/bin/`,
+		`for f in "$FINCH_DIR"/cred-helpers/docker-credential-*; do ` +
+			`[ -L "/usr/local/bin/$(basename ${f})" ] || sudo ln -s "${f}" "/usr/local/bin/$(basename ${f})"; done`,
```
swagatbora90, line 95: The logic seems to be only checking for if a symbolic link exists. If there is a regular file by the same name, this may still fail.

pkg/config/nerdctl_config_applier.go
```diff
@@ -91,8 +91,8 @@ func addLineToBashrc(fs afero.Fs, profileFilePath string, profStr string, cmd st
 func updateEnvironment(fs afero.Fs, fc *Finch, finchDir, homeDir, limaVMHomeDir string) error {
 	cmdArr := []string{
 		`export DOCKER_CONFIG="$FINCH_DIR"`,
-		"[ -L /usr/local/bin/docker-credential-ecr-login ] " +
-			`|| sudo ln -s "$FINCH_DIR"/cred-helpers/docker-credential-ecr-login /usr/local/bin/`,
+		`for f in "$FINCH_DIR"/cred-helpers/docker-credential-*; do ` +
```
swagatbora90, line 94: Do we want to validate the files before copying? I am thinking if it will be better to verify the files with the cred_helpers config value. This way it makes the configuration clear and we can also throw an error here if the particular cred-helper binary is not found. wdyt?

austinvazquez: Discussed offline. Instead of linking all files from credential helpers matching pattern; will iterate over all credential helpers in configuration and only link those files to the path. The result is any configuration which lists a binary that is not in cred-helpers directory will fail.

pkg/config/nerdctl_config_applier.go
```diff
@@ -91,11 +91,18 @@ func addLineToBashrc(fs afero.Fs, profileFilePath string, profStr string, cmd st
...
 		`export DOCKER_CONFIG="$FINCH_DIR"`,
-		"[ -L /usr/local/bin/docker-credential-ecr-login ] " +
-			`|| sudo ln -s "$FINCH_DIR"/cred-helpers/docker-credential-ecr-login /usr/local/bin/`,
 		`[ -L /root/.aws ] || sudo ln -fs "$AWS_DIR" /root/.aws`,
 	}
 
+	for _, credHelper := range fc.CredsHelpers {
+		cmdArr = append(cmdArr,
+			fmt.Sprintf(
+				`[ -L /usr/local/bin/docker-credential-%s ] || sudo ln -s "$FINCH_DIR"/cred-helpers/docker-credential-%s /usr/local/bin/`,
```
swagatbora90, line 100: Can we add a explicit error/warning here if the source binary is not present in the cred-helpers dir?

pendo324 (dismissed):

pkg/config/nerdctl_config_applier.go
austinvazquez: Addressed by https://github.com/runfinch/finch/pull/1265/commits/360785dbbf1bf430e224ae2da2e298b10e85ce37. As symbolic link would be successful even if the source did not exist, I added logic to first check if the source exists before checking if the symbolic link exists.

pendo324: This is pretty much just an info/warning message though, right? Since we aren't checking if there's an error and then returning it back in Go. That's probably fine for now

pendo324 (approved):

pkg/config/nerdctl_config_applier.go
austinvazquez: @pendo324 pinged me offline because I missed this comment. For completeness, yes it is best effort to inform users the configuration would not work as expected. The issue is this logic is added to the user's guest machine bashrc and is invoked on each command. Note: container finch commands shell (SSH) into the guest machine and execute nerdctl. The message would not be propagated up through Finch CLI though.

swagatbora90 (approved):

PR #1273, opened 2025-02-05
ci: consolidate upload test report into a reusable action

pendo324:
Issue #, if available:

*Description of changes:*
- Uses a standard ubuntu-latest runner instead of self-hosted runners for uploading test reports and consolidates all of the logic into one reusable action

*Testing done:*

#### License Acceptance

### DIFF
--- a/.github/workflows/e2e-linux.yaml
+++ b/.github/workflows/e2e-linux.yaml
@@ -134,93 +134,23 @@ jobs:
         if: ${{ steps.vars.outputs.has_creds == 'true' && (!(startsWith(inputs.os, 'amazon') && inputs.version == '2' ))}}
         uses: actions/upload-artifact@65c4c4a1ddee5b72f698fdd19549f0f0fb45cf08 # v4.6.0
         with:
-          name: linux-${{ inputs.version }}-${{ inputs.test-command }}-${{ inputs.arch }}-${{ github.run_id }}-${{ github.run_attempt }}-e2e-reports
+          name: linux-${{ inputs.version }}-test-e2e-${{ inputs.arch }}-${{ github.run_id }}-${{ github.run_attempt }}-e2e-reports
           path: ${{ github.workspace }}/reports/${{ github.run_id }}-${{ github.run_attempt }}-*.json
       - name: Clean up repo AL2
         if: ${{ (startsWith(inputs.os, 'amazon') && inputs.version == '2' && always() ) }}
         run: |
           rm -rf "${GITHUB_WORKSPACE}"
-  upload-linux-e2e-test-report:
+  upload-e2e-test-reports:
     needs: test
     if: always()
-    runs-on:
-      [
-        "self-hosted",
-        "linux",
-        "${{ inputs.arch }}",
-        "${{ inputs.version }}",
-        "${{ inputs.runner-type }}",
-      ]
-    steps:
-      - name: Set output variables to check AL2
-        id: linux2
-        run: |
-          is_al2=${{ (startsWith(inputs.os, 'amazon') && inputs.version == '2' ) }}
-          echo "is_al2=$is_al2" >> $GITHUB_OUTPUT
-      - name: configure aws credentials
-        uses: aws-actions/configure-aws-credentials@4fc4975a852c8cd99761e2de1f4ba73402e44dd9 # v4.0.3
-        if: |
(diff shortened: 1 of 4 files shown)


.github/workflows/e2e-windows.yaml
```diff
@@ -111,64 +116,17 @@ jobs:
...
-            Write-Host "VM serial report uploaded successfully."
-          } else {
-            Write-Host "VM serial report file does not exist. Skipping upload."
-          }
+    uses: ./.github/workflows/upload-test-report.yaml
+    secrets: inherit
+    with:
+      os: "windows"
+      arch: ${{ inputs.arch }}
+      version: "2022"
```
chews93319, line 127: (nit) For consistency, should this be changed into a variable like the mac&linux and passed by the `ci.yaml` in lines 180-183?

.github/workflows/e2e-windows.yaml
```diff
@@ -101,7 +106,7 @@ jobs:
         if: always()
         uses: actions/upload-artifact@65c4c4a1ddee5b72f698fdd19549f0f0fb45cf08 # v4.6.0
         with:
-          name: windows-${{ inputs.test-command }}-${{ inputs.arch }}-${{ github.run_id }}-${{ github.run_attempt }}-e2e-reports
+          name: windows-${{ inputs.test-command }}-2022-${{ inputs.arch }}-${{ github.run_id }}-${{ github.run_attempt }}-e2e-reports
```
chews93319, line 109: (nit)
 - `2022` magic value could become a variable for consistency
 - naming scheme is a different order than mac&linux
     - win: `test-command`-`version`-`arch`
     - mac&linux: `version`-`test-command`-`arch`

.github/workflows/e2e-linux.yaml
```diff
@@ -135,94 +135,23 @@ jobs:
...
-          else
-            echo "VM serial report file does not exist. Skipping upload."
-          fi
+    uses: ./.github/workflows/upload-test-report.yaml
+    secrets: inherit
+    with:
+      os: ${{ inputs.os }}
+      arch: ${{ inputs.arch }}
+      version: ${{ inputs.version }}
+      test-command: "test-e2e"
```
chews93319, line 153: (nit) For consistency, should this be changed into a variable like the mac&win and passed by the ci.yaml in lines 193-197?

.github/workflows/e2e-linux.yaml
```diff
@@ -135,94 +135,23 @@ jobs:
         if: ${{ steps.vars.outputs.has_creds == 'true' && (!(startsWith(inputs.os, 'amazon') && inputs.version == '2' ))}}
         uses: actions/upload-artifact@65c4c4a1ddee5b72f698fdd19549f0f0fb45cf08 # v4.6.0
         with:
-          name: linux-${{ inputs.version }}-${{ inputs.test-command }}-${{ inputs.arch }}-${{ github.run_id }}-${{ github.run_attempt }}-e2e-reports
+          name: linux-${{ inputs.version }}-test-e2e-${{ inputs.arch }}-${{ github.run_id }}-${{ github.run_attempt }}-e2e-reports
```
chews93319, line 138: (nit)
 - `test-e2e` magic value could become a variable for consistency

chews93319 (approved):

.github/workflows/e2e-windows.yaml
pendo324: It's just a little hard to change it into a variable. For macOS/Linux its a variable because we test on multiple versions of those systems, where as on Windows, we only test on one version

.github/workflows/e2e-linux.yaml
pendo324: On Linux, we run all of the e2e tests at once since they don't require use of a VM and take much less time, whereas on macOS/Windows, that's not the case (so we split it up into two different runs)

.github/workflows/e2e-windows.yaml
pendo324: That'll probably cause issues, but I'd really like to just merge this to get the better utilization and fix it in a follow up PR. Is that ok with you?

chews93319: Yes, I'm ok with that. I would also be willing to do the "fast-follow" CR to make this update.

PR #1285, opened 2025-02-11
fix: container runtime full archive parse artifact path variable

austinvazquez:
Issue #, if available:
Found in #1249 on Windows if environment variable is not defined then Make is not handling the same as on macOS and Linux.

*Description of changes:*
This change adds strip to environment variable check which will return an empty string if environment variable does not exist which is more robust on all platforms.

#### License Acceptance


pendo324 (approved):

austinvazquez:
@pendo324, yep I should have opened as draft. I can rebase and try to get this in after the v1.7.0 release.

PR #1367, opened 2025-04-15
fix: Add Buildctl cache prune

Shubhranshu153:
Issue #, if available:

*Description of changes:*
Finch system prune didn't clean up build cache once vm was restarted or re-init as buildkit dir was not mounted to persistent disk. This fix, mounts the buildkit dir to persistent disk while provisioning and adds a test to verify the clean up of cache on a re-init

*Testing done:*
Local Testing done to verify system prune is able to clean up the cache.

#### License Acceptance

### DIFF
--- a/e2e/vm/vm_cache_cleanup_test.go
+++ b/e2e/vm/vm_cache_cleanup_test.go
@@ -1,0 +1,78 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+//go:build darwin
+
+package vm
+
+import (
+	"fmt"
+	"strings"
+	"time"
+
+	"github.com/onsi/ginkgo/v2"
+	"github.com/onsi/gomega"
+	"github.com/runfinch/common-tests/command"
+	"github.com/runfinch/common-tests/ffs"
+	"github.com/runfinch/common-tests/option"
+)
+
+var testVMPrune = func(o *option.Option, installed bool) {
+	ginkgo.Describe("Test System Prune Cleans Up Build Cache, Images, Containers", func() {
+		ginkgo.It("check prune cleans up build cache, images and containers", func() {
+			limaCtlO, err := limaCtlOpt(installed)
+			gomega.Expect(err).ShouldNot(gomega.HaveOccurred())
+
+			command.Stdout(limaCtlO, "shell", "finch", "mount")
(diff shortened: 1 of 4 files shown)


swagatbora90 (approved):

swagatbora90 (approved):

pendo324 (dismissed):

e2e/vm/vm_cache_cleanup_test.go
```diff
@@ -0,0 +1,59 @@
...
+	"time"
+
+	"github.com/onsi/ginkgo/v2"
+	"github.com/onsi/gomega"
+	"github.com/runfinch/common-tests/command"
+	"github.com/runfinch/common-tests/ffs"
+	"github.com/runfinch/common-tests/option"
+)
+
+var testVMPrune = func(o *option.Option, installed bool) {
```
Kern--, line 20: Should this test also verify that containerd doesn't have any stray content/leases/snapshots to verify that buildkit's cache isn't forgetting about resources?

Shubhranshu153: Added to check the leases, snapshots identifying may be harder. As long as gc.flats leases are clean we should be fine otherwise it will be an error in containerd gc.

Kern--: What we really care about is that space is reclaimed. Whether the bug is in buildkit or containerd gc doesn't really matter to people using `finch system prune`. 

That said, you're probably right that checking for the buildkit cache leases is a good proxy based on what you mentioned offline: if the cache forgot about its objects the second lease check would fail and if it stopped using the "gc.flat" label the first check would fail and we could fix the test.

Kern-- (dismissed):

Shubhranshu153:
Updates:
1. Skipped the test from windows due to this error i see, need a separate fix for it.
```
/*
			TODO: Fix issue with prune in windows. Error:
			failed marshalling json out of network configuration file \"/etc/cni/net.d/nerdctl-bridge.conflist\":
			error parsing configuration list:unexpected end of JSON input\nFor details on the schema"
		*/
```

2. Simplified the test to verify the buildctl prune. Containerd snapshots doesnt seem to be reliably cleared for the test. The test probably wont cover all test cases but would catch the issue with builder cache. 

Will create couple of backlog tickets for fixing  the above errors incrementally

pendo324 (approved):

PR #1369, opened 2025-04-16
feat: Support disk resizing via VM settings[WIP]

subashkotha:
Issue #, if available:

*Description of changes:* 

Added support for disk resize by introducing a new `diskSize` field in the VM settings command and persisting it in the Finch config file.

*Testing done:* Added unit-tests, e2e tests are WIP

- [ X] I've reviewed the guidance in CONTRIBUTING.md

#### License Acceptance

### DIFF
--- a/cmd/finch/virtual_machine_settings_darwin.go
+++ b/cmd/finch/virtual_machine_settings_darwin.go
@@ -38,6 +38,11 @@ func newSettingsVMCommand(
 		config.DefaultMemory,
 		"the amount of memory to dedicate to the virtual machine (restart the vm when applying this change.)",
 	)
+	settingsVMCommand.Flags().String(
+		"disk-size",
+		config.DefaultDiskSize,
+		"the size of the virtual machine disk (restart the vm when applying this change.)",
+	)
 
 	return settingsVMCommand
 }
@@ -74,9 +79,15 @@ func (sva *settingsVMAction) runAdapter(cmd *cobra.Command, _ []string) error {
 		return err
 	}
 
+	diskSize, err := cmd.Flags().GetString("disk-size")
+	if err != nil {
+		return err
+	}
+
 	opts := config.VMConfigOpts{
-		CPUs:   cpus,
-		Memory: memory,
+		CPUs:     cpus,
+		Memory:   memory,
+		DiskSize: diskSize,
 	}
 
 	return sva.run(opts)
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -59,14 +59,16 @@ type Nerdctl struct {
 
 // VMConfigOpts represents the Options for finch vm settings command.
 type VMConfigOpts struct {
-	CPUs   int
-	Memory string
+	CPUs     int
+	Memory   string
+	DiskSize string
 }
 
 // Default values for the command line arguments --cpus and --memory.
 const (
-	DefaultCPUs   = 0
-	DefaultMemory = ""
+	DefaultCPUs     = 0
+	DefaultMemory   = ""
+	DefaultDiskSize = ""
 )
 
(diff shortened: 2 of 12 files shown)


pkg/config/validate_darwin.go
```diff
@@ -54,5 +52,19 @@ func validate(cfg *Finch, log flog.Logger, systemDeps LoadSystemDeps, mem fmemor
 		)
 	}
 
+	// Todo confirm if we need more validations
+	diskSize, err := units.FromHumanSize(*cfg.DiskSize)
+	if err != nil {
+		return fmt.Errorf("invalid requested disk size format: %w", err)
+	}
+
+	const maxRecommendedSize = 200 * units.GB
```
Shubhranshu153, line 61: This i feel is opinionated in the sense the sense this will depend on the size of the system it is run, would suggest us not to control it.

pkg/config/config_darwin.go
```diff
@@ -10,6 +10,7 @@ import (
 	"strconv"
 	"strings"
 
+	"github.com/docker/go-units"
```
Shubhranshu153, line 13: do we need this extra package?

cmd/finch/virtual_machine_settings_darwin.go
```diff
@@ -38,6 +38,11 @@ func newSettingsVMCommand(
 		config.DefaultMemory,
```
Shubhranshu153, line 38: need to add readme and also need to add it to the website

pkg/config/config_darwin.go
```diff
@@ -81,8 +83,30 @@ func ModifyFinchConfig(fs afero.Fs, logger flog.Logger, finchConfigPath string,
...
+		currentDiskSize := *finchCfg.DiskSize
+
+		currentSize, _ := units.FromHumanSize(currentDiskSize)
+		requestedSize, err := units.FromHumanSize(opts.DiskSize)
+		if err != nil {
+			return false, fmt.Errorf("invalid requested disk size format: %w", err)
+		}
+
+		// Throw error if attempting to shrink disk size
+		if requestedSize <= currentSize {
```
Shubhranshu153, line 96: lima already does it so we need not handle it.
if lima in future will support we will also allow so i dont think we need to do it.

pkg/config/config.go
```diff
@@ -59,14 +59,16 @@ type Nerdctl struct {
 
 // VMConfigOpts represents the Options for finch vm settings command.
```
Shubhranshu153, line 60: In general to test:
Do we need to modify all the tests? 
DiskSize should not be a mandatory parameter

pkg/disk/disk_darwin.go
```diff
@@ -35,11 +34,22 @@ func (m *userDataDiskManager) EnsureUserDataDisk() error {
...
-				return err
+		// Get current disk info for size comparison
+		info, err := m.getDiskInfo(diskPath)
+		if err != nil {
+			return err
+		}
+
+		// Check if resize is needed
+		configuredSize, _ := units.FromHumanSize(*m.config.DiskSize)
+		if configuredSize > int64(info.VirtualSize) {
```
Shubhranshu153, line 45: this logic should already be in limactl?

pkg/disk/disk_darwin_test.go
```diff
@@ -114,35 +116,39 @@ func TestUserDataDiskManager_InitializeUserDataDisk(t *testing.T) {
 			name: "disk already exists",
```
Shubhranshu153, line 116: The existing test need not be modified we need to test one case when no size is there we are able to configure it as 50, if some size in yaml we are able to configure correctly, if we run finch vm disk resize ...  we can configure the correct size. 

Need a vm integ test for it, where we modify the size and we are able to see the change.

PR #1386, opened 2025-04-30
fix: image name parsing

Shubhranshu153:
Issue #, if available:

*Description of changes:*
We are accessing arg[:2] which does a null ptr reference for args of len=1

*Testing done:*
Did a manual test and added an unit test to capture the case. Verified the test fails without my changes.

- [x ] I've reviewed the guidance in CONTRIBUTING.md

#### License Acceptance

### DIFF
--- a/cmd/finch/nerdctl_darwin_test.go
+++ b/cmd/finch/nerdctl_darwin_test.go
@@ -1204,6 +1204,33 @@ func TestNerdctlCommand_run(t *testing.T) {
 				c.EXPECT().Run()
 			},
 		},
+		{
+			name:    "with single letter entry in args",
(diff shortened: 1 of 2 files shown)


pendo324 (approved):

pendo324:
Requires minor lint fixes but lgtm. We should add an e2e test case for this, but that can be follow up

Shubhranshu153:
e2e test for all cases is kind of bloat i feel. Let me check after this release i can fix my CLI refactor then we dont need to worry about all these edge cases.

pendo324 (approved):

subashkotha (approved):

PR #1437, opened 2025-06-24
ci: Add upload verified artifacts to S3 action

cartrius:
*Description of changes:*
- This workflow allows tracking on the latest "verified" base OS/rootfs artifacts and digests
- This action will not run until Finch-core has been updated for Finch which typically happens after running the "Sync Submodules and Dependencies" workflow. This would imply that the latest configurations (containing artifact and digest information) have passed checks and are compatible with Finch
- The latest artifact and digest information will updated in latest-os-artifacts.json and latest-rootfs-artifacts.json in S3

*Testing done:*
- Simulated the workflow on my own fork
- Checked for the JSON files within a test S3 bucket

#### License Acceptance

### DIFF
--- a/.github/workflows/upload-verified-artifacts-to-s3.yaml
+++ b/.github/workflows/upload-verified-artifacts-to-s3.yaml
@@ -1,0 +1,70 @@
+name: Upload Verified Artifacts to S3
+
+on:
+  push:
+    branches:
+      - main
+    paths:
+      - 'deps/finch-core'
+
+permissions:
+  id-token: write
+  contents: read
+
+jobs:
+  upload-artifacts:
+    name: Upload Artifacts to S3
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+        with:
+          submodules: recursive
+          
+      - name: Configure AWS credentials
+        uses: aws-actions/configure-aws-credentials@b47578312673ae6fa5b5096b330d9fbac3d116df # v4.2.1
+        with:
+          aws-region: ${{ secrets.REGION }}
+          role-to-assume: ${{ secrets.ROLE }}
+          role-session-name: update-latest-os-artifacts-in-s3
+          
+      - name: Update latest base OS artifact information in S3
+        run: |
+          # Update the latest verified base OS artifact information in S3
+          echo "Uploading the verified base OS artifact information to S3"
+
+          # Extract updated artifact information
+          source deps/finch-core/deps/full-os.conf
+          
+          # Create manifest file to hold artifact information
+          MANIFEST_FILE="latest-os-artifacts.json"
+          echo '{
+            "aarch64": {
+              "filename": "'$AARCH64_ARTIFACT'",
+              "sha512sum": "'$AARCH64_512_DIGEST'"
+            },
+            "x86_64": {
+              "filename": "'$X86_64_ARTIFACT'",
+              "sha512sum": "'$X86_64_512_DIGEST'"
+            }
+          }' > $MANIFEST_FILE
+          
+          # Upload to S3
+          aws s3 cp $MANIFEST_FILE s3://${{ secrets.ARTFACT_BUCKET_NAME }}/manifest/latest-os-artifacts.json --content-type "application/json"
+
+      - name: Update latest rootfs information in S3
+        run: |
+          # Update the latest verified rootfs information in S3
+          echo "Uploading the latest verified rootfs information in S3"
+          # Extract updated artifact information
+          source deps/finch-core/deps/rootfs.conf
+          
+          # Create manifest file to hold artifact information
+          MANIFEST_FILE="latest-rootfs-artifacts.json"
+          echo '{
+            "filename": "'$X86_64_ARTIFACT'",
+            "sha512sum": "'$X86_64_512_DIGEST'"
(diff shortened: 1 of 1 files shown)


pendo324:
> This action will not run until Finch-core has been updated for Finch which typically happens after running the "Sync Submodules and Dependencies" workflow. This would imply that the latest configurations (containing artifact and digest information) have passed checks and are compatible with Finch

We also have nightly test runs, which test a mock installed version of Finch. Would it be even safer to run this workflow as part of the nightly run?

.github/workflows/upload-verified-artifacts-to-s3.yaml
```diff
@@ -0,0 +1,70 @@
...
+          
+          # Create manifest file to hold artifact information
+          MANIFEST_FILE="latest-rootfs-artifacts.json"
+          echo '{
+            "filename": "'$X86_64_ARTIFACT'",
+            "sha512sum": "'$X86_64_512_DIGEST'"
+          }' > $MANIFEST_FILE
+          
+          # Upload to S3
+          aws s3 cp $MANIFEST_FILE s3://${{ secrets.ARTIFACT_BUCKET_NAME }}/manifest/latest-rootfs-artifacts.json --content-type "application/json"
```
sondavidb: nit: Can you add an extra newline at the end for consistency?

sondavidb:
> We also have nightly test runs, which test a mock installed version of Finch. Would it be even safer to run this workflow as part of the nightly run?

I was thinking it might be a bit of an expensive operation but we run the whole testing suite and upload installers to S3 every night anyway, so I think this is a relatively light operations all things considered. @cartrius any opinions on this?

pendo324:
> > We also have nightly test runs, which test a mock installed version of Finch. Would it be even safer to run this workflow as part of the nightly run?
> 
> I was thinking it might be a bit of an expensive operation but we run the whole testing suite and upload installers to S3 every night anyway, so I think this is a relatively light operations all things considered. @cartrius any opinions on this?

Yeah just to be clear, I'm proposing tacking this on top of the existing nightly release workflow, not running the entire "release validation steps" an additional time

cartrius:
@pendo324 @sondavidb Since we are already running a whole mock installed version of Finch, I think adding this on would be great and more safe. I wasn't aware of the nightly test runs, thanks for the suggestion!

sondavidb (dismissed):
LGTM, I'm not super familiar with the workflow run syntax but I xref'd [here](https://github.com/orgs/community/discussions/66512) and it looks fine.

I'm a little concerned that we can't test this but I think we can followup on this. Maybe add an extra trigger on PR and have it upload to a test file, but I'm not sure that's a valuable test anyway.

.github/workflows/upload-verified-artifacts-to-s3.yaml
```diff
@@ -0,0 +1,77 @@
+name: Upload Verified Artifacts to S3
+
+on:
+  push:
+    branches:
+      - main
+    paths:
+      - 'deps/finch-core'
+  workflow_run:
+    workflows:
```
pendo324: Won't this make it so only one of the nightly workflows needs to complete to trigger both the macOS and Windows rootfs to be updated?

I think it would be better to add this logic to build-and-test-msi.yaml and build-and-test-pkg.yaml

cartrius: You're correct. This current implementation would only require one to complete not both. Just to clarify, are you suggesting that I split and add the base OS and rootfs steps into build-and-test-msi.yaml and build-and-test-pkg.yaml? Then, I should omit the path check for deps/finch-core too?

pendo324: > are you suggesting that I split and add the base OS and rootfs steps into build-and-test-msi.yaml and build-and-test-pkg.yaml

Yeah, I think this ends up being simpler/easier to reason about

> Then, I should omit the path check for deps/finch-core too?

Not sure what you mean about this?

cartrius: On L3-L8, I am also checking to see if there was a change to the deps/finch-core directory which could happen after someone many pushes the finch-core changes to finch or if we use the Sync Submodules and Deps action. Though, we have that scheduled to happen at the same time as the nightly test runs. If I attach these steps to the nightly build, I am thinking that I no longer have to check for this change to deps/finch-core.

sondavidb: Moving this to those two workflows would mean that this no longer gets run on every change to finch-core, but instead only when the build workflows happen, which would be nightly or manually, and to me whether or not that's ok depends on at what point we consider it "safe" to push it to S3.

IMO, if it's changed in finch-core and the change gets merged to Finch, then it's safe, and the installers here are more for the Finch application, versus the base image we're running on, so I'm not convinced that we should move it behind the nightly builds, unless there's something I'm missing that ties the OS image to these workflows? I guess I can't imagine a scenario where it passes the PR tests but fails the packaging into an installer step.

cartrius: I decided to revert my changes to run the workflow based on push to deps/finch-core. I believe that it would be "safe" enough to use the pushed configurations because they would have passed tests within finch-core and finch. I agree that it would make some sense to wait for the nightly tests, but a lot of the verification is already done by that point. This would cause some redundancy and potential delays.

sondavidb (approved):

PR #1451, opened 2025-07-09
feat: Add finch update-finch command

cartrius:
*Description of changes:*
- Added new `finch update-finch` command to allow users to update Finch to the newest release through the CLI. This will download and install the pkg/msi with the newest version of Finch. This feature supports users on MacOS and Windows. Linux users should continue to use package managers
- Added `--check-only` or `-c` flag to check for updates without installing

*Testing done:*
- Added unit-tests for newly created `finch update-finch`
- Updated finch locally (on darwin/aarch64)

#### License Acceptance


cmd/finch/updatefinch.go
```diff
@@ -0,0 +1,86 @@
...
+func newUpdateCommand(logger flog.Logger, stdOut io.Writer) *cobra.Command {
+	updateCommand := &cobra.Command{
+		Use:   "update-finch",
+		Args:  cobra.NoArgs,
+		Short: "Update Finch to the latest version",
+		Long:  "Check for and install the latest version of Finch",
+		RunE:  newUpdateAction(logger, stdOut).runAdapter,
+	}
+
+	updateCommand.Flags().BoolP("check-only", "c", false, "Check if an update is available without installing it")
```
sondavidb: I wonder if this should be the default behavior, and instead lock the actual installation process behind a flag. I'm envisioning a use case where a user is curious about what certain commands do, run `finch update` when an update happens to be available, and it suddenly installs it for them.

pkg/version/version_update_checker.go
```diff
@@ -0,0 +1,73 @@
...
+}
+
+// IsUpdateAvailable checks if the current version is different from the latest version
+func IsUpdateAvailable() (bool, string, error) {
+	latestVersion, err := GetLatestVersion()
+	if err != nil {
+		return false, "", err
+	}
+
+	if Version != latestVersion {
```
sondavidb: We should also do a check and ensure that Version is not modified, e.g. if someone compiles from source and makes changes they probably won't want to update. We can make the argument that they should still be notified that an update is available, but if we go down that route we should call that out in a comment here.

cmd/finch/updatefinch.go
sondavidb, line 1: Would be good to stick with our convention so maybe rename to `update_finch.go`

cmd/finch/main_remote_test.go
```diff
@@ -195,7 +195,7 @@ func TestNewApp(t *testing.T) {
 	assert.Equal(t, cmd.SilenceUsage, true)
 	assert.Equal(t, cmd.SilenceErrors, true)
 	// confirm the number of command, comprised of nerdctl commands + finch commands
-	assert.Equal(t, len(cmd.Commands()), len(nerdctlCmds)+4)
+	assert.Equal(t, len(cmd.Commands()), len(nerdctlCmds)+5)
```
sondavidb, line 198: I'm sure this is right if the tests pass with this change but curious on the reasoning for this?

cmd/finch/updatefinch.go
```diff
@@ -0,0 +1,86 @@
...
+	checkOnly, err := cmd.Flags().GetBool("check-only")
+	if err != nil {
+		return err
+	}
+
+	return ua.run(checkOnly)
+}
+
+func (ua *updateAction) run(checkOnly bool) error {
+	fmt.Fprintln(ua.stdOut, "Checking for updates...")
```
sondavidb: I think we should use `ua.logger.Info` here, took a look at finch vm remove [func](https://github.com/runfinch/finch/blob/330c19803b642233624d9709fb526afea6059e90/cmd/finch/virtual_machine_remove.go#L79) and looks like this is what they use to print to console. Probably any instance of `fmt.Println` should be replaced with that. (And if there's any additional debug messages you can probably do `ua.logger.Debug` to log those ([ref](https://github.com/runfinch/finch/blob/330c19803b642233624d9709fb526afea6059e90/cmd/finch/nerdctl_darwin.go#L42))

cmd/finch/updatefinch_test.go
```diff
@@ -0,0 +1,279 @@
...
+
+import (
+	"bytes"
+	"errors"
+	"testing"
+	"fmt"
+
+	"github.com/runfinch/finch/pkg/mocks"
+
+	"github.com/golang/mock/gomock"
```
sondavidb: Ah, checker failed here because we use `go.uber.org/mock/gomock`. I *think* that should have similar behavior to this one, so just change the import name and hopefully none of the behavior breaks 🤞

pkg/updatefinch/updatefinch_test.go
```diff
@@ -0,0 +1,368 @@
...
+		{
+			name:        "linux unsupported",
+			version:     "1.0.0",
+			goos:        "linux",
+			goarch:      "amd64",
+			expectedURL: "",
+			expectError: true,
+		},
+		{
+			name:        "version with v prefix",
```
sondavidb: nit: Do we run into this case? Not a bad check either way but curious

pkg/updatefinch/updatefinch_test.go
sondavidb, line 1: Same here — `update_finch_test.go` (or whatever snake case name you decide on) and also the folder. (Though, maybe you can just call the folder update? Up to you 🤷 )

cmd/finch/updatefinch_test.go
```diff
@@ -0,0 +1,279 @@
...
+	}
+
+	cmd := newUpdateCommand(logger, &stdOut)
+	cmd.SetArgs([]string{"--check-only"})
+	
+	err := cmd.Execute()
+	assert.NoError(t, err)
+	
+	output := stdOut.String()
+	assert.Contains(t, output, "Checking for update")
```
sondavidb: I think you might also want to move all the `assert.Contains` to `logger.EXPECT().Info` (or `Infof`, or `Debug`, or whatever) once you output to the logger instead of stdout

pkg/updatefinch/updatefinch.go
```diff
@@ -0,0 +1,144 @@
...
+// DownloadAndInstallUpdate downloads and installs the specified verison of Finch
+func DownloadAndInstallUpdate(latestVersion string) error {
+	// Get the appropriate download URL based on platform (MacOS or Windows)
+	downloadURL, err := getDownloadURL(latestVersion)
+	if err != nil {
+		return fmt.Errorf("failed to get download URL: %w", err)
+	}
+	
+	// Create a temporary file to store the download
+	tempDir := os.TempDir()
```
sondavidb: Don't forget to remove it once we're done (just do `defer os.Remove(tempDir)` after this line)

pkg/updatefinch/updatefinch.go
```diff
@@ -0,0 +1,144 @@
...
+	// Install the update
+	if err := installUpdate(filePath); err != nil {
+		return fmt.Errorf("failed to install update: %w", err)
+	}
+	
+	return nil
+}
+
+// getDownloadURL returns the appropriate download URL for the current platform
+func getDownloadURL(version string) (string, error) {
```
sondavidb: So I was looking in your unit test for this and you only run this if we are on the correct arch, which makes sense but makes testing on a singular system a little difficult. What if we take another string variable called "os" and then in `DownloadAndInstallUpdate` we can populate this var with `runtime.GOOS`. Then our unit tests can just pass in the desired os/arch.

pkg/updatefinch/updatefinch.go
```diff
@@ -0,0 +1,144 @@
...
+	_, err = io.Copy(out, resp.Body)
+	return err
+}
+
+func installUpdate(filePath string) error {
+	var cmd *exec.Cmd
+	
+	switch runtime.GOOS {
+	case "darwin":
+		cmd = exec.Command("sudo", "installer", "-pkg", filePath, "-target", "/")
```
sondavidb: Don't use sudo here. If we run into any issues with permission issues we can probably print an error message saying "try running with sudo" or something like that, but we should never run as admin for the user.

sondavidb (commented):
Left quite a few comments (it's a pretty big change!) — overall really happy with the general flow/direction of it, just a few parts that I think need a little more attention. Excited to get this change out 🙂

cmd/finch/updatefinch.go
cartrius: I was thinking of that use case as well. At first, I was thinking about adding a yes/no confirmation, but I decided to switch over to a `--install` flag.

pkg/version/version_update_checker.go
cartrius: I just realized that this code snippet is actually redundant because I am also making checking for updates in pkg/update/update.go at L112-130. I removed this section and renamed the file to version_latest.go. Also, I have added a conditional that checks for the `.modified` suffix and warns users who may be on a modified build of Finch.

cmd/finch/main_remote_test.go
cartrius: I did have to change this to pass the unit-test, but I was also curious. This change was made long ago by @pendo324. Could you explain this please?

pkg/updatefinch/updatefinch_test.go
cartrius: This test case runs, but it isn't supposed to actually happen because I have implemented splitting to get the version in a format like #.#.#

pkg/updatefinch/updatefinch_test.go
cartrius: I have changed the folder name to "update" and adjusted the file names as well. Though, I have kept the cmd files related to this as update_finch.go/update_finch_test.go.

cmd/finch/main_remote_test.go
sondavidb: Synced offline, our best guess is that this checks that the amount of commands we have are the same as the amount of nerdctl commands we support, plus however many non-nerdctl commands we support, and since update is not related to nerdctl this gets incremented

pkg/updatefinch/updatefinch.go
cartrius: I have changed this to print an error message suggesting to run with sudo. Though, I was wondering what your thoughts may be on using AppleScript via `oascript`. This way, we could have GUI authentication instead and wouldn't have to make the user type in the command again with sudo.

pendo324: oascript sounds like a neat solution

sondavidb: +1, this is also probably a better experience for the user

pkg/update/update.go
```diff
@@ -0,0 +1,178 @@
...
+	downloadURL, err := getDownloadURL(latestVersion, runtime.GOOS, runtime.GOARCH)
+	if err != nil {
+		return fmt.Errorf("failed to get download URL: %w", err)
+	}
+
+	// Create a temporary file to store the download
+	tempDir := os.TempDir()
+	defer func() {
+		if err := os.Remove(tempDir); err != nil {
+			return
```
sondavidb: This effectively just ignores the error, which maybe is ok but we should probably log the error in some way. (I think we can just print an error message?)

pkg/update/update.go
```diff
@@ -0,0 +1,178 @@
...
+func downloadFile(url, filePath string) error {
+	filePath = filepath.Clean(filePath)
+	out, err := os.Create(filePath)
+	if err != nil {
+		return err
+	}
+
+	defer func() {
+		if err := out.Close(); err != nil {
+			return
```
sondavidb: Same here

pkg/update/update.go
```diff
@@ -0,0 +1,178 @@
...
+	}()
+
+	resp, err := http.Get(fmt.Sprintf("https://%s", path.Clean(url)))
+	if err != nil {
+		return err
+	}
+
+	defer func() {
+		if err := resp.Body.Close(); err != nil {
+			return
```
sondavidb: Same here

pkg/updatefinch/updatefinch.go
cartrius: I decided to incorporate both suggestions of the error message and `osascript` with a flag. I think this would be helpful for future extensions of Finch.

cmd/finch/update_finch_test.go
```diff
@@ -0,0 +1,247 @@
...
+			if tc.wantErr != nil {
+				assert.Equal(t, tc.wantErr.Error(), err.Error())
+			} else {
+				assert.Equal(t, tc.wantErr, err)
+			}
+		})
+	}
+}
+
+//nolint:paralleltest // This function manipulates variables to facilitate mocking.
```
sondavidb, line 119: It would probably be good practice if we can get this parallelized. Maybe we can take it as a followup?

pkg/update/update.go
```diff
@@ -0,0 +1,163 @@
...
+}
+
+func installUpdate(filePath string) error {
+	var cmd *exec.Cmd
+
+	switch runtime.GOOS {
+	case "darwin":
+		script := fmt.Sprintf(`do shell script "installer -pkg %s -target /" with administrator privileges with `+
+			`prompt "Finch is requesting permission to install version update"`, filePath)
+		cmd = exec.Command("osascript", "-e", script) //nolint:gosec // osascript subprocess
```
sondavidb: I think we should at least call `filepath.Clean()` on the filepath passed in, this might also make the linter happy.

sondavidb:
ubuntu tests failing I think are not an issue with the PR since we're seeing failures across the board

pkg/update/update.go
sondavidb: Oh, could you also remove the `nolint` here? I'm curious if the filepath.Clean will make the linter happy with this. (Or if it has any suggestions on how to make this safe...)

sondavidb: Synced offline, the linter doesn't like the fact that we call it via a variable, I think technically we can make it happy by just putting the `fmt.Sprintf` inline but that seems clunky and weird...I think this approach is probably fine.

sondavidb (dismissed):

PR #1466, opened 2025-07-18
fix(vm): Don't fail if settings match existing one

sondavidb:
Issue #, if available:

*Description of changes:*
Currently if we run `finch vm settings` we can set the CPU and memory via flags. However, if the flags do not change anything, we print an error `the number of CPUs or the amount of memory should be at least one valid value`. IMO, this is not a particularly clear error message, and it might be better to print a more benign error in its place, so this PR does that.

Not sure if we're opinionated yet on this but IMO this is a little cleaner and removes some confusion, feel free to close if we don't want to do this.

*Testing done:*
`make test-unit`

#### License Acceptance

### DIFF
--- a/cmd/finch/virtual_machine_settings_darwin.go
+++ b/cmd/finch/virtual_machine_settings_darwin.go
@@ -74,14 +74,20 @@ func (sva *settingsVMAction) runAdapter(cmd *cobra.Command, _ []string) error {
 		return err
 	}
 
+	cpusChanged := cmd.Flags().Changed("cpus")
+	memoryChanged := cmd.Flags().Changed("memory")
+
 	// check if any flags were provided by the user
-	if !cmd.Flags().Changed("cpus") && !cmd.Flags().Changed("memory") {
+	if !cpusChanged && !memoryChanged {
 		return cmd.Help()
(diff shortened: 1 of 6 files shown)


Kern-- (approved):
It's still a little odd that you can specify one of cpu or memory to be the empty value which means no change. e.g. `finch vm settings --cpus 0 --memory 4GiB` is allowed and doesn't change the CPU setting. 

I think this is a good change overall.

sondavidb:
> It's still a little odd that you can specify one of cpu or memory to be the empty value which means no change. e.g. `finch vm settings --cpus 0 --memory 4GiB` is allowed and doesn't change the CPU setting.
> 
> I think this is a good change overall.

Yeah, with the existing code it is a little weird. Let me see if I can make it work a little better.

sondavidb:
Think I made it better in behavior, not super happy with the output of the code but I guess it kind of works. The big issue here is that we have no good way of knowing if the flag was set as a bad value or a default value, since both are the same to us.

Maybe I could have made this a pointer...I actually kind of like that. Let me re-simplify this and hopefully it will be a little easier to work with then 🙂

PR #1475, opened 2025-07-28
fix(ci): run linter in ci-docs.yaml

sondavidb:
We had a case where the linter got updated but failed to report that it didn't run because it was a YAML-only change that didn't run the linter. This should help mitigate the issue as well as hopefully catch any YAML-only changes that also affect our code.

Issue #, if available:

*Description of changes:*

*Testing done:*

#### License Acceptance

### DIFF
--- a/.github/workflows/ci-docs.yaml
+++ b/.github/workflows/ci-docs.yaml
@@ -77,12 +77,9 @@ jobs:
     timeout-minutes: 2
     steps:
       - run: echo "Skipping CI for docs & contrib files"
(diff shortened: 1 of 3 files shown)


sondavidb:
Ugh, AFAICT there's no good way to remove the annoying name-wrapping that's being done....

Is it worth changing our required jobs to be `go-linter/lint`?

sondavidb:
I just decided to not use a reusable workflow to get around our required checks. Maybe we should change that down the line but for now this will do.

PR #1491, opened 2025-08-04
feat(macOS): Add finch vm disk usage command

cartrius:
*Description of changes:*
- New `finch vm disk usage` command to allow users to see detailed disk information (filesystem, mount-point, total size, used space, available space, usage percentage) when their VM is running

*Testing done:*
- Added unit-tests for newly created `finch vm disk usage`
- Ran command locally on macOS

#### License Acceptance

### DIFF
--- a/cmd/finch/virtual_machine_disk_usage.go
+++ b/cmd/finch/virtual_machine_disk_usage.go
@@ -1,0 +1,98 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+//go:build darwin
+
+package main
+
+import (
+	"fmt"
+	"strings"
+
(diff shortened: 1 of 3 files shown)


cmd/finch/virtual_machine_disk_usage.go
```diff
@@ -0,0 +1,98 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
Shubhranshu153, line 1: have we looked into implementing equivalent of 
https://docs.docker.com/reference/cli/docker/system/df/

I think this implementation will be appropriately set there.

sondavidb: I looked into the code base and this seems a lot more complex. I think it's probably a lot more maintainable to just use CLI commands.

PR #1550, opened 2025-09-09
ci(fix): race condition when building and testing installers

pendo324:
Issue #, if available:

## Description of changes
When building and testing the installers, it's possible for the ref which we are building the installer for (e.g. `main`), to change during the build process. This is less likely when building off of a tag, but recently when testing various fixes to the release workflow, I had this happen several times when building installers off of `main`.

The workflow will now checkout the ref once at the start, and propagate the exact commit to checkout to the rest of the steps which re-checkout.

*Testing done:*

#### License Acceptance

### DIFF
--- a/.github/workflows/build-and-test-deb.yaml
+++ b/.github/workflows/build-and-test-deb.yaml
@@ -4,6 +4,7 @@ on:
   workflow_dispatch:
     inputs:
       ref_name:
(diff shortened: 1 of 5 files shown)


pendo324:
Re-created PR off of a branch of origin, rather than my fork, so I can test this workflow: https://github.com/runfinch/finch/pull/1551

PR #1561, opened 2025-09-19
fix: update benchmark supported versions

sondavidb:
Issue #, if available:

*Description of changes:*
We haven't supported MacOS for a few months now. Changed our benchmarking page to show MacOS 13 + 14 instead of 12 + 13, and moved MacOS 12 benchmarks behind a drop-down menu.

*Testing done:*

Checks if the local HTML looks good, it's a bit annoying that opening also moves around all the other elements but this seems well out of my scope to fix this.

#### License Acceptance

### DIFF
--- a/dev/bench/index.html
+++ b/dev/bench/index.html
@@ -37,9 +37,15 @@
 </head>
 <body>
 <h1>Finch Benchmark</h1>
-<a href="./macOS/12/arm64/index.html">macOS 12 arm64</a>
-<a href="./macOS/12/x86_64/index.html">macOS 12 x86_64</a>
 <a href="./macOS/13/arm64/index.html">macOS 13 arm64</a>
 <a href="./macOS/13/x86_64/index.html">macOS 13 x86_64</a>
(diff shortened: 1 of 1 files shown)


dev/bench/index.html
```diff
@@ -37,9 +37,15 @@
...
 <body>
 <h1>Finch Benchmark</h1>
-<a href="./macOS/12/arm64/index.html">macOS 12 arm64</a>
-<a href="./macOS/12/x86_64/index.html">macOS 12 x86_64</a>
 <a href="./macOS/13/arm64/index.html">macOS 13 arm64</a>
 <a href="./macOS/13/x86_64/index.html">macOS 13 x86_64</a>
+<a href="./macOS/14/arm64/index.html">macOS 14 arm64</a>
+<a href="./macOS/14/x86_64/index.html">macOS 14 x86_64</a>
+<br>
+<details>
```
sondavidb, line 45: We can alternatively just remove any reference to it. We can just get to the data if we need to by changing the URL manually.

PR #1582, opened 2025-10-10
ci: add debian canaries

coderbirju:
Issue #, if available:
This workflow 
- verifies that finch is available via self hosted apt package in ubuntu
- runs every 5 mins

*Description of changes:*

*Testing done:*

#### License Acceptance

### DIFF
--- a/.github/workflows/canary-deb.yaml
+++ b/.github/workflows/canary-deb.yaml
@@ -1,0 +1,90 @@
+name: Deb Canary
+
+on:
+  # This workflow will run from the release-automation.yaml automation on each merge
+  workflow_dispatch:
+    inputs:
+      ref_name:
+        required: true
+        type: string
+  workflow_call:
+    inputs:
+      ref_name:
+        required: true
+        type: string
+
+  # This workflow will run every 5 min
+  schedule:
+    - cron: '*/5 * * * *'
+  
+  # This workflow will run when the workflow file is updated
+  pull_request:
+    paths:
+      - '.github/workflows/canary-deb.yaml'
+  
+jobs:
+  get-latest-tag:
+    name: Get the latest release tag
+    runs-on: ubuntu-latest
+    timeout-minutes: 2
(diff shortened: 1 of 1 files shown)


.github/workflows/canary-deb.yaml
```diff
@@ -0,0 +1,85 @@
+name: Deb Canary
+
+on:
```
sondavidb, line 3: Can we also add a condition to run it every time this file is modified?

.github/workflows/canary-deb.yaml
```diff
@@ -0,0 +1,90 @@
...
+      ref_name:
+        required: true
+        type: string
+
+  # This workflow will run every 5 min
+  schedule:
+    - cron: '*/5 * * * *'
+  
+  # This workflow will run when the workflow file is updated
+  push:
```
sondavidb: `pull_request` would be more suitable here

.github/workflows/canary-deb.yaml
```diff
@@ -0,0 +1,90 @@
...
+      - uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
+        with:
+          fetch-depth: 0
+      - name: 'Get the latest tag'
+        id: latest-tag
+        uses: "WyriHaximus/github-action-get-previous-tag@04e8485ecb6487243907e330d522ff60f02283ce" # v1.4.0
+
+  canary-deb:
+    name: Test Finch APT installation
+    runs-on: ubuntu-latest
```
sondavidb, line 42: Should we also run this on self-hosted runners? (I'm not certain myself...)

coderbirju: We're setting up the apt repository ourselves in the workflow and this is going to be the same in either case. Any reason why we want to run this in self-hosted runners?

pendo324 (approved):

PR #1593, opened 2025-10-20
ci(canary): move to script, use own downloader

sondavidb:
Issue #, if available:

*Description of changes:*
I found that getting the release tag takes quite a while since we clone the whole repo. The release downloader action has a way to get the tag output so I'd like to reuse that so our canaries run even faster.

Additionally, added an optimization so that we only update the runfinch sources repo. This would also take some time in some runs of this.

These optimizations combined should bring our canary runs pretty consistently under 15 seconds.

*Testing done:*

#### License Acceptance

### DIFF
--- a/.github/workflows/canary-deb.yaml
+++ b/.github/workflows/canary-deb.yaml
@@ -21,55 +21,18 @@ on:
   pull_request:
     paths:
       - '.github/workflows/canary-deb.yaml'
+      - 'scripts/canary-deb.sh'
   
 jobs:
   canary-deb:
     name: Test Finch APT repo health
     runs-on: ubuntu-latest
-    timeout-minutes: 3
+    timeout-minutes: 2
     steps:
-      - name: Setup environment variables
-        run: |
-          ARCH=$(dpkg --print-architecture)
-          echo "ARCH=${ARCH}" >> $GITHUB_ENV
-      - name: Download latest Finch release from GitHub
-        id: release-download
-        uses: "robinraju/release-downloader@daf26c55d821e836577a15f77d86ddc078948b05" # v1.12.0
+      - name: Checkout canary script
+        uses: actions/checkout@v5
         with:
-          latest: true
-          fileName: '*${{ env.ARCH }}.deb'
-          out-file-path: github-release
-      - name: Add Finch APT Repository
-        run: |
-          echo "Detected architecture: ${{ env.ARCH }}"
-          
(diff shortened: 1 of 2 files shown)


.github/workflows/canary-deb.yaml
```diff
@@ -23,62 +23,46 @@ on:
...
-
-          # Strip v from tag
-          tag=${{ needs.get-latest-tag.outputs.tag }}
-          version=${tag/v/}
-          echo "version=${version}" >> $GITHUB_ENV
-
-          echo "filename=runfinch-finch_${version}_${ARCH}.deb" >> $GITHUB_ENV
+      - name: Download latest Finch release from GitHub
+        id: release-download
+        uses: "robinraju/release-downloader@daf26c55d821e836577a15f77d86ddc078948b05" # v1.12.0
```
Shubhranshu153: this one seems to be a personal github action. I am not sure we should be using it.

pendo324: It's not that hard to `curl https://api.github.com/repos/runfinch/finch/releases/latest` and use `jq` to get some data from the release as an alternative

sondavidb: If the concern is licensing/maintainability I'd argue a script doing exactly what we want might be a little more cumbersome (though still feasible). I did also check and licensing isn't an issue (MIT license).

pendo324: I think the concern is just taking an unnecessary dependency for something trivial. With SHA pinning, it's unlikely we could be impacted by a supply chain attack, but it is still a non-zero chance, compared to creating our own implementation.

sondavidb: Fair enough. I can do that then.

.github/workflows/canary-deb.yaml
```diff
@@ -23,65 +23,56 @@ on:
...
     name: Test Finch APT repo health
     runs-on: ubuntu-latest
-    timeout-minutes: 3
-    needs: get-latest-tag
+    timeout-minutes: 2
     steps:
-      - name: Setup environment variables
+      - name: Download latest Finch release from GitHub
         run: |
           ARCH=$(dpkg --print-architecture)
```
pendo324: One nit not necessarily related to this PR: can we also move this logic into it's own bash script (maybe in the `scripts` directory)? It should allow for easier local testing

sondavidb: The whole process or just this particular step?

pendo324 (approved):

Shubhranshu153 (approved):

PR #1600, opened 2025-10-23
fix: Change permissions for package.sh

Shubhranshu153:
Issue #, if available:

*Description of changes:*
Change permissions for package.sh to enable execution.

*Testing done:*

#### License Acceptance

### DIFF
--- a/contrib/packaging/deb/update-dependencies.sh
+++ b/contrib/packaging/deb/update-dependencies.sh
@@ -79,6 +79,8 @@ update_dependency() {
         -e "s/${name}_COMMIT=\"[^\"]*\"/${name}_COMMIT=\"$new_commit\"/" \
         "$PACKAGE_SH" > "$temp_file"
     
+    # Use package file permissions
+    chmod --reference="$PACKAGE_SH" "$temp_file"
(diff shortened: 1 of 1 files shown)


Kern--:
Why not add execute to the script and commit it to the repo instead?

pendo324:
When I look at this file in my locally checked out repo, it seems to already be marked as executable. Why do we need this `chmod +x` at runtime?

```
$ ls -l ./contrib/packaging/deb/package.sh
-rwxr-xr-x  1 alvajus  staff   9.0K Sep  3 10:26 ./contrib/packaging/deb/package.sh
```

Shubhranshu153:
I think  this operation used in update-debian dependencies is causing the permissions issues we see: https://github.com/runfinch/finch/blob/da20d2085a80508f09c98091fd06fb78d4a75e64/contrib/packaging/deb/update-dependencies.sh#L71

This is overriding the permissions. Will modify that instead to have the permissions here.

Kern-- (approved):
Would `sed --in-place` remove the need for the temp file all together? I don't think it matters that much, though.

PR #1630, opened 2025-11-24
feat: add journalctl logs to support-bundle

sondavidb:
Issue #, if available:

*Description of changes:*
Added support for journalctl logs to be added to the support bundle. Default logs include containerd and finch-daemon logs.

*Testing done:*
I wasn't sure of the best way to test this so the best I could do was add something to e2e tests to ensure that the logs folder exists. Ideally I wanted to check if the files were populated with the right info.

#### License Acceptance

### DIFF
--- a/cmd/finch/main_native_test.go
+++ b/cmd/finch/main_native_test.go
@@ -88,7 +88,6 @@ func TestXmain(t *testing.T) {
 	}
 
 	for _, tc := range testCases {
-		tc := tc
 		t.Run(tc.name, func(t *testing.T) {
 			t.Parallel()
 
--- a/docs/cmd/finch_support-bundle_generate.md
+++ b/docs/cmd/finch_support-bundle_generate.md
@@ -9,7 +9,7 @@ Generates a collection of logs and configs that can be uploaded to a Github issu
 ## Options
 
 ```text
-      --exclude stringArray   files to exclude from the support bundle. if you specify a base name, all files matching that base name will be excluded. if you specify an absolute or relative path, only exact matches will be excluded
+      --exclude stringArray   files to exclude from the support bundle. If you specify a base name, all files matching that base name will be excluded. If you specify an absolute or relative path, only exact matches will be excluded. To exclude journal logs for a service, prefix the file path with "service":. To exclude all journal logs, use "service:all"
   -h, --help                  help for generate
-      --include stringArray   additional files to include in the support bundle, specified by absolute or relative path. to include a file from the VM, prefix the file path with "vm:"
+      --include stringArray   additional files to include in the support bundle, specified by absolute or relative path. To include journal logs for a service, prefix the file path with "service:". To include a file from the VM, prefix the file path with "vm:"
 ```
--- a/pkg/support/config_native_linux.go
+++ b/pkg/support/config_native_linux.go
@@ -16,3 +16,7 @@ func (bc *bundleConfig) ConfigFiles() []string {
 		bc.finch.ConfigFilePath(),
 	}
 }
+
+func (bc *bundleConfig) JournalServices() []string {
+	return []string{"service:containerd", "service:finch", "service:buildkit", "service:soci"}
+}
--- a/pkg/support/support.go
+++ b/pkg/support/support.go
@@ -14,6 +14,7 @@ import (
 	"path"
 	"path/filepath"
 	"runtime"
+	"slices"
 	"strings"
 	"time"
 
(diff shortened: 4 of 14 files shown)


pkg/support/config_native_linux.go
```diff
@@ -16,3 +16,7 @@ func (bc *bundleConfig) ConfigFiles() []string {
 		bc.finch.ConfigFilePath(),
 	}
 }
+
+func (bc *bundleConfig) JournalServices() []string {
+	return []string{"service:containerd", "service:finch@503.service"}
```
Kern--: 503 is a UID and shouldn't be hardcoded (in fact, I think 501 is first user's UID).

pendo324: In addition, we should also get buildkit and soci logs here

pkg/support/support.go
```diff
@@ -219,10 +233,23 @@ func (bb *bundleBuilder) copyAndRedactFile(writer io.Writer, reader bufReader) e
...
 
 func (bb *bundleBuilder) copyInFile(writer *zip.Writer, fileName string, prefix string) error {
-	// check filename validity?
-	f, err := bb.fs.Open(fileName)
-	if err != nil {
-		return err
+	var f io.Reader
+	if isService(fileName) {
+		service := strings.TrimPrefix(fileName, "service:")
+		cmd := bb.ecc.Create("journalctl", "-xu", service)
```
pendo324: Looks like we already have https://github.com/coreos/go-systemd in our go.mod file, and it uses a more official programmatic method of reading logs. Is it possible or worth it to use this library? I'm not sure since we're running finch on the host and go-systemd would need access to files or sockets (dbus) inside the Linux VM

pkg/support/config_native_linux.go
sondavidb: Would "service:finch" suffice here?

pkg/support/support.go
sondavidb: I can look into this tomorrow.

pkg/support/config_native_linux.go
pendo324: Unless you can somehow do a prefix search, I think you just need to insert the UID.

Kern--: I didn't realize this is the native linux file, "service:finch" should work here.

It looks like journactl does have wildcards - "service:finch@*" should work for macos. Does windows put the UID into the service name?

pkg/support/support.go
sondavidb: Looking at the [journal package](https://github.com/coreos/go-systemd/tree/main/journal) I don't think this would be extensible for operations within the VM, at least from looking at the code...

sondavidb:
Specifying no-pager seemed to help with the weird hangs I saw.

From some offline discussions, I added a couple extra things
- Allow a keyword `service:all` to exclude all default journalctl logs. Notably this doesn't override any additional logs put in with `--include` which I think is fine but LMK if you think it should be different. Added test cases for the new flag
- Realized journalctl logs will be populated even with no entries so I removed the systemctl startup test and just amended the initial default bundle test

sondavidb:
Even more changes....

So some of the tests are failing which leads to the support bundles not being cleaned up after tests. The way that we find the support bundle is that we check anything with the prefix, but we have the info for the specific one that got created anyway, so I refactored the tests to do this so that we are always checking the right zip files. If nothing else this should make it a little easier to see exactly which tests are failing...

Shubhranshu153:
nit: Squash the commits

cmd/finch/main_native_test.go
```diff
@@ -88,7 +88,6 @@ func TestXmain(t *testing.T) {
 	}
 
 	for _, tc := range testCases {
-		tc := tc
```
Shubhranshu153, line 91: is this a general cleanup? 
If yes ideally we have it in a separate commit.

docs/cmd/finch_support-bundle_generate.md
```diff
@@ -9,7 +9,7 @@ Generates a collection of logs and configs that can be uploaded to a Github issu
 ## Options
 
 ```text
-      --exclude stringArray   files to exclude from the support bundle. if you specify a base name, all files matching that base name will be excluded. if you specify an absolute or relative path, only exact matches will be excluded
+      --exclude stringArray   files to exclude from the support bundle. If you specify a base name, all files matching that base name will be excluded. If you specify an absolute or relative path, only exact matches will be excluded. To exclude journal logs for a service, prefix the file path with "service":. To exclude all journal logs, use "service:all"
```
Shubhranshu153, line 12: any reason we want the include and exclude options at all, seems like it is adding more complexity for customers.

cmd/finch/main_native_test.go
sondavidb: I actually have no idea. These just got randomly thrown in when I ran the linter or something🤷 can remove if you'd like

sondavidb:
I'll merge this soon but have a couple concerns I'll be discussing offline. Squashed commits for now.

sondavidb:
Followup concern: truncate log length (right now we just collect the entire systemd logs, we can truncate to ~100 or so). Will merge for now and have a followup

PR #1634, opened 2025-12-05
fix: finch run panic without arguments

### LINKED ISSUE #1633: `finch run` with no arguments results in a panic
**Please do not report security issues through this template.** Refer to [Security](https://github.com/runfinch/finch?tab=security-ov-file#reporting-security-issues) for the proper reporting channels.

---

**Steps to reproduce**

Run on MacOS:

```
$ finch run   
panic: runtime error: index out of range [0] with length 0

goroutine 1 [running]:
main.(*nerdctlCommand).run(0x1400040aba0, {0x1036d8cc0, 0x3}, {0x104357080, 0x0, 0x0})
        /Users/ec2-user/ar/_work/finch/finch/cmd/finch/nerdctl_remote.go:97 +0x36f8
main.(*nerdctlCommand).runAdapter(0x1400040aba0, 0x1036d9351?, {0x104357080, 0x0, 0x0})
        /Users/ec2-user/ar/_work/finch/finch/cmd/finch/nerdctl.go:96 +0x4c
github.com/spf13/cobra.(*Command).execute(0x1400044c908, {0x104357080, 0x0, 0x0})
        /Users/ec2-user/go/pkg/mod/github.com/spf13/cobra@v1.10.1/command.go:1015 +0x844
github.com/spf13/cobra.(*Command).ExecuteC(0x14000442008)
        /Users/ec2-user/go/pkg/mod/github.com/spf13/cobra@v1.10.1/command.go:1148 +0x384
github.com/spf13/cobra.(*Command).Execute(...)
        /Users/ec2-user/go/pkg/mod/github.com/spf13/cobra@v1.10.1/command.go:1071
main.xmain({0x103b062b8, 0x104357080}, {0x103b00740, 0x104357080}, {0x103b06c20, 0x104357080}, {0x103af1778, 0x104357080}, {0x103af1878, 0x104357080}, ...)
        /Users/ec2-user/ar/_work/finch/finch/cmd/finch/main_remote.go:68 +0x298
main.main()
        /Users/ec2-user/ar/_work/finch/finch/cmd/finch/main.go:29 +0x8c
```

**Expected behavior**

Finch should show me the help prompt if I pass no arguments.

coderbirju:
Issue #, if available:
fixes #1633 

*Description of changes:*
Added a length check to prevent panic when run without arguments

*Testing done:*
Before 
```
~ finch run    
panic: runtime error: index out of range [0] with length 0
..
~ finch compose
panic: runtime error: index out of range [0] with length 0
..
~ finch exec
panic: runtime error: index out of range [0] with length 0
..
```
After 
```
~ finch run 
Run a command in a new container. Optionally specify "ipfs://" or "ipns://" scheme to pull image from IPFS.

Usage: finch container run [flags] IMAGE [COMMAND] [ARG...]

Flags:
      --add-host strings                               Add a custom host-to-IP mapping (host:ip)
      --annotation stringArray                         Add an annotation to the container (passed through to the OCI runtime)
  -a, --attach strings                                 Attach STDIN, STDOUT, or STDERR
      --blkio-weight uint16                            Block IO (relative weight), between 10 and 1000, or 0 to disable (default 0)
      --blkio-weight-device stringArray                Block IO weight (relative device weight) (default [])
      --cap-add strings                                Add Linux capabilities
      --cap-drop strings                               Drop Linux capabilities
      --cgroup-conf strings                            Configure cgroup v2 (key=value)
      --cgroup-parent string                           Optional parent cgroup for the container
      --cgroupns string                                Cgroup namespace to use, the default depends on the cgroup version ("host"|"private") (default "private")
      --cidfile string                                 Write the container ID to the file
     .....
~ finch exec
Run a command in a running container

Usage: finch exec [flags] CONTAINER COMMAND [ARG...]

Flags:
  -d, --detach             Detached mode: run command in the background
  -e, --env stringArray    Set environment variables
      --env-file strings   Set environment variables from file
  -h, --help               help for exec
  -i, --interactive        Keep STDIN open even if not attached
      --privileged         Give extended privileges to the command
  -t, --tty                Allocate a pseudo-TTY
  -u, --user string        Username or UID (format: <name|uid>[:<group|gid>])
  -w, --workdir string     Working directory inside the container

See also 'finch --help' for the global flags such as '--namespace', '--snapshotter', and '--cgroup-manager'. 

~ finch compose 
Compose

Usage: finch compose [flags] COMMAND

Commands:
  build    Build or rebuild services
  config   Validate and view the Compose file
  cp       Copy files/folders between a service container and the local filesystem
  create   Creates containers for one or more services
  down     Remove containers and associated resources
  exec     Execute a command in a running container of the service
  images   List images used by created containers in services
  kill     Force stop service containers
  logs     Show logs of running containers
  pause    Pause all processes within containers of service(s). They can be unpaused with finch compose unpause
  port     Print the public port for a port binding
  ps       List containers of services
  pull     Pull service images
  push     Push service images
  restart  Restart containers of given (or all) services
  rm       Remove stopped service containers
  run      Run a one-off command on a service
  start    Start existing containers for service(s)
  stop     Stop running containers without removing them.
  top      Display the running processes of service containers
  unpause  Unpause all processes within containers of service(s).
  up       Create and start containers
  version  Show the Compose version information
....
```

#### License Acceptance

### DIFF
--- a/cmd/finch/nerdctl_remote_test.go
+++ b/cmd/finch/nerdctl_remote_test.go
@@ -14,7 +14,6 @@ import (
 	"github.com/stretchr/testify/assert"
 	"go.uber.org/mock/gomock"
 
-	"github.com/runfinch/finch/pkg/command"
 	"github.com/runfinch/finch/pkg/config"
 	"github.com/runfinch/finch/pkg/mocks"
 )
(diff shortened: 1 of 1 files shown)


Shubhranshu153:
Can we add a test, if its breaking that means we are missing a test for it somewhere.

cmd/finch/nerdctl_remote_test.go
```diff
@@ -132,3 +133,96 @@ func TestNerdctlCommand_withVMErrors(t *testing.T) {
 		})
 	}
 }
+
+func TestNerdctlCommand_withEmptyArgs(t *testing.T) {
```
sondavidb: I wonder if we can simplify this test by just ensuring that creating and running the command doesn't result in a panic. It feels like this does a lot to ensure that certain variables are set but all we want to do is check the command output IMO.

sondavidb: This is a unit test so I don't think it has to run per se, but I do believe the path it follows to mock command runs should panic if you run it without your changes and pass if you run it with your changes

cmd/finch/nerdctl_remote_test.go
```diff
@@ -132,3 +133,96 @@ func TestNerdctlCommand_withVMErrors(t *testing.T) {
...
+			args:    []string{},
+		},
+		{
+			name:    "finch exec with no arguments",
+			cmdName: "exec",
+			args:    []string{},
+		},
+	}
+
+	for _, tc := range testCases {
```
Shubhranshu153: The fact that unit testing requires this much setup suggest to me we need to refactor this code.

cmd/finch/nerdctl_remote_test.go
coderbirju: This is the simplest way to do a unit test unfortunately, however I think there is a case to be made that this is not the right place to test it. 
[common-tests](https://github.com/runfinch/common-tests) is the more appropriate place to have an integration test for this use case. That should not be a blocker to merge this PR though.
[Draft PR](https://github.com/runfinch/common-tests/pull/275) yet to be tested

sondavidb: I think it can still be in the Finch repo. I just don't think it has to do everything it does right now. I'm OK to have the test as a followup.

I'm just thinking that since the unit test calls some functions of the actual command for parsing it should just make sure we don't panic. We don't have to test too much besides that behavior.

sondavidb (approved):

PR #1653, opened 2025-12-30
feat: default to OS credstore for storing registry creds

ayush-panta:
*Description of changes:* This PR enables Finch to integrate with native OS credential stores (Keychain on macOS, Credential Manager on Windows) for container registry authentication.

*How it works:*
- finch login/logout commands execute on host using native credential helpers
- When running container commands, Finch creates a temporary credential socket between host and VM
- The VM's docker-credential-finch helper connects to this socket to retrieve credentials
- Socket is automatically cleaned up when the command completes

*Dependencies:*
- Requires [finch-daemon PR #382](https://github.com/runfinch/finch-daemon/pull/382) for credential helper updates
- Requires finch-core to be updated to point to finch-daemon commit

*Testing done:* Tested locally; works alongside changes to finch-daemon for standard workflows. Added E2E tests for macOS, which worked locally (see below). Must add tests for Windows.
```
% go test -v -ginkgo.focus="Native Credential Helper" -timeout=30m
=== RUN   TestVM
Running Suite: Finch Virtual Machine E2E Tests - ../finch/e2e/vm
==================================================================================================
Random Seed: 1767117182

Will run 4 of 225 specs
SS••••SSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSS

Ran 4 of 225 Specs in 453.266 seconds
SUCCESS! -- 4 Passed | 0 Failed | 0 Pending | 221 Skipped
--- PASS: TestVM (453.29s)
PASS
ok  	github.com/runfinch/finch/e2e/vm	454.011s
```

#### License Acceptance


pkg/config/nerdctl_config_applier.go
```diff
@@ -95,12 +95,24 @@ func updateEnvironment(fs afero.Fs, fc *Finch, finchDir, homeDir, limaVMHomeDir
...
+	//nolint:gosec // G101: Potential hardcoded credentials false positive - this is a template string
+	const configureFinchCredHelperTemplate = `[ -L /usr/local/bin/docker-credential-finch ] || \
+  (sudo ln -s /usr/bin/docker-credential-finch /usr/local/bin/docker-credential-finch)`
+
 	for _, credHelper := range fc.CredsHelpers {
-		cmdArr = append(cmdArr, fmt.Sprintf(configureCredHelperTemplate, credHelper, credHelper, credHelper, credHelper))
+		if credHelper == "finch" {
+			cmdArr = append(cmdArr, configureFinchCredHelperTemplate)
+			// Add as default in credStore to work with nerdctl command flow. 
+			// TODO: support multi-registry workflows. Will involve set registries as credStore vs credHelpers in config.json.
```
pendo324, line 111: How much more work is this? It'd be nice to have this in the initial implementation so we don't have to think about what happens when user's update

cmd/finch/cred_helper.go
```diff
@@ -0,0 +1,86 @@
...
+	if err != nil {
+		return "", fmt.Errorf("failed to get home directory: %w", err)
+	}
+
+	var helperName string
+	switch runtime.GOOS {
+	case "darwin":
+		helperName = "docker-credential-osxkeychain"
+	case "windows":
+		helperName = "docker-credential-wincred.exe"
```
pendo324, line 32: Are we expecting the user to install the credential helpers specifically in the finch path? Other tools just require the credential helpers to be in the PATH. This might not be expected.

cmd/finch/cred_socket.go
```diff
@@ -0,0 +1,113 @@
+//go:build darwin || windows
```
pendo324, line 1: Can we create a new directory/package for all of the credential handling files? Namely: `cred_socket.go` and `cred_helper.go`?

ayush-panta: Will do

cmd/finch/cred_helper.go
ayush-panta: The user does not have to install anything. I just symlinked the binaries that are packaged with finch-core to .finch/cred-helpers, and then depend on these for credhelper operations. 

I did this because as it is right now, if someone configures ecr-login in finch.yaml (for example), docker-credential-ecr-login gets installed into ./finch/cred-helpers. 

However, I can try to make it just check if helpers are in PATH while also creating the symlink.

pkg/config/nerdctl_config_applier.go
ayush-panta: I'm giving this some thought right now. I think it is a matter of making ecr-login behave like acr-login:
```
{
...
"credHelpers": {
    "dkr.acr.builder-tools.aws.dev": "acr-login",
    "dkr.ecr.public.aws....": "ecr-login",
}
...
}
```
docker-credential-acr-login has a --setup flag that allows it to populate the config file by itself, but docker-credential-ecr-login does not. 

I think what I will do is make it so that if ecr-login is set as a credhelper in finch.yaml, we store it as a credHelper for generic ecr links

ayush-panta: ^ Going to make this a mini-feature after the project.

cmd/finch/cred_helper.go
```diff
@@ -0,0 +1,86 @@
...
+	default:
+		return "", fmt.Errorf("unsupported OS: %s", runtime.GOOS)
+	}
+
+	helperPath := filepath.Join(homeDir, ".finch", "cred-helpers", helperName)
```
Shubhranshu153, line 37: This path neednot be in the home dir for example
```
/opt/homebrew/bin/docker-credential-osxkeychain
```

i would say we neednot handle packaging them at all. Just have the process in the docs.

Shubhranshu153: to find the path depending on OS you can do a which or where

cmd/finch/cred_helper.go
```diff
@@ -0,0 +1,86 @@
...
+	helperPath, err := getHelperPath()
+	if err != nil {
+		return nil, err
+	}
+
+	// #nosec G204 - helperPath is validated and comes from trusted source
+	cmd := exec.Command(helperPath, action)
+
+	// Set input based on action
+	if action == "store" {
```
Shubhranshu153, line 55: if else is not usually desirable in function, try to have single functionality per function

Shubhranshu153: Makes it easier to test also.

cmd/finch/cred_helper.go
```diff
@@ -0,0 +1,86 @@
...
+		cred := dockerCredential{
+			ServerURL: serverURL,
+			Username:  username,
+			Secret:    password,
+		}
+		credJSON, err := json.Marshal(cred)
+		if err != nil {
+			return nil, fmt.Errorf("failed to marshal credentials: %w", err)
```
Shubhranshu153, line 63: dont print the err here, it might leak creds.

cmd/finch/cred_helper.go
```diff
@@ -0,0 +1,86 @@
...
+		}
+		cmd.Stdin = strings.NewReader(string(credJSON))
+	} else {
+		cmd.Stdin = strings.NewReader(serverURL)
+	}
+
+	output, err := cmd.CombinedOutput()
+	if err != nil {
+		return nil, fmt.Errorf("credential helper failed: %w - %s", err, string(output))
```
Shubhranshu153, line 72: same as above.

cmd/finch/cred_socket.go
```diff
@@ -0,0 +1,113 @@
...
+
+var globalCredSocket = &credentialSocket{}
+
+func (cs *credentialSocket) start(finchRootPath string) error {
+	cs.mu.Lock()
+	defer cs.mu.Unlock()
+
+	// Break if already running
+	if cs.listener != nil {
+		return nil
```
Shubhranshu153, line 29: need an error saying concurrent cred request not supported in finch mac and windows version.

cmd/finch/cred_socket.go
```diff
@@ -0,0 +1,113 @@
...
+	}
+
+	socketPath := filepath.Join(finchRootPath, "lima", "data", "finch", "sock", "creds.sock")
```
Shubhranshu153, line 32: this should be part of makefile, similar to finch,sock

Shubhranshu153: if socket not there we just throw an error.

cmd/finch/nerdctl.go
```diff
@@ -136,8 +136,8 @@ var nerdctlCmds = map[string]string{
 	"inspect":   "Return low-level information on Docker objects",
 	"kill":      "Kill one or more running containers",
 	"load":      "Load an image from a tar archive or STDIN",
-	"login":     "Log in to a container registry",
-	"logout":    "Log out from a container registry",
+	// "login":     "Log in to a container registry",
```
Shubhranshu153, line 139: nit: commented out code.

cmd/finch/nerdctl_remote.go
```diff
@@ -377,7 +382,17 @@ func (nc *nerdctlCommand) run(cmdName string, args []string) error {
...
-	return nc.ncc.Create(runArgs...).Run()
+	// Get finch root path for socket
+	execPath, err := os.Executable()
+	if err != nil {
+		return err
+	}
+	finchRootPath := filepath.Dir(filepath.Dir(execPath))
+	
+	// Wrap nerdctl execution with credential socket
+	return withCredSocket(finchRootPath, func() error {
```
Shubhranshu153, line 393: This will prevent any concurrent execution but thats fine.

cmd/finch/cred_helper.go
Shubhranshu153: we should think a little more here whether to package it or not. Packaging it will potentially add maintenance so i am not inclined, but makes it easier for customers

@pendo324 what do you think?

cmd/finch/virtual_machine_init.go
```diff
@@ -123,6 +125,69 @@ func (iva *initVMAction) run() error {
...
+
+	// Setup credential helpers
+	if err := iva.ensureNativeCredentialHelpers(); err != nil {
+		iva.logger.Warnf("Failed to setup credential helpers: %v", err)
+	}
+
+	return nil
+}
+
+func (iva *initVMAction) ensureNativeCredentialHelpers() error {
```
Shubhranshu153, line 137: i would like to avoid having to manage the installation rather find it from the environment.

cmd/finch/cred_helper.go
Shubhranshu153, line 37: https://pkg.go.dev/os/exec -> this probably can be used to get the path.
Not sure it works for windows but we can try.

cmd/finch/cred_socket.go
```diff
@@ -0,0 +1,113 @@
...
+	credJSON, err := json.Marshal(creds)
+	if err != nil {
+		return
+	}
+	
+	_, _ = conn.Write(credJSON)
+}
+
+// withCredSocket wraps command execution with credential socket lifecycle.
+func withCredSocket(finchRootPath string, fn func() error) error {
```
Shubhranshu153, line 107: how will this handle detach case?
Wont it keep holding the console, Lets do a check for it

```
finch run -d alpine sleep 10
```
something like that.

Also lets verify the behavior if a image exists then what happens will the request still come?

cmd/finch/nerdctl_remote.go
Shubhranshu153: I think its good to wrap it for functionalities we actually need creds, otherwise it might un-necessary hold it. This part is actually flimsy.

finch.yaml.d/mac.yaml
```diff
@@ -7,6 +7,11 @@ provision:
   - mode: boot
     script: |
       modprobe virtiofs
+  - mode: boot
+    script: |
+      # Create user-writable directory in /run for credential socket
+      mkdir -p /run/finch-user-sockets
+      chmod 777 /run/finch-user-sockets
```
Shubhranshu153, line 14: this seems overly permissive.

cmd/finch/login_local.go
```diff
@@ -0,0 +1,115 @@
...
+
+	serverAddress := ""
+	if len(args) == 1 {
+		serverAddress = args[0]
+	}
+
+	return loginWithNativeCredStore(serverAddress, username, password, cmd.OutOrStdout())
+}
+
+func loginWithNativeCredStore(serverAddress, username, password string, stdout io.Writer) error {
```
Shubhranshu153, line 66: can we not use login package from nerdctl:
https://github.com/containerd/nerdctl/blob/53e7b272af14b075a5e8d7b95a5c2d862a1620f8/pkg/cmd/login/login.go#L44

we just use it.

cmd/finch/cred_socket.go
```diff
@@ -0,0 +1,113 @@
...
+	if !scanner.Scan() {
+		return
+	}
+	command := strings.TrimSpace(scanner.Text())
+	
+	// Read server URL
+	if !scanner.Scan() {
+		return
+	}
+	serverURL := strings.TrimSpace(scanner.Text())
```
Shubhranshu153, line 90: serverURL probably needs some sanitization with regex to check its a valid server request, but probably docker cred helper also does it?

cmd/finch/cred_socket.go
```diff
@@ -0,0 +1,113 @@
...
+}
+
+var globalCredSocket = &credentialSocket{}
+
+func (cs *credentialSocket) start(finchRootPath string) error {
+	cs.mu.Lock()
+	defer cs.mu.Unlock()
+
+	// Break if already running
+	if cs.listener != nil {
```
Shubhranshu153, line 28: not sure what it does, if there are separate request you wont see them across CLI calls. so this is not a valid check for concurrent connection.

e2e/vm/vm_darwin_test.go
```diff
@@ -75,6 +75,7 @@ func TestVM(t *testing.T) {
 		testNonDefaultOptions(o, *e2e.Installed)
 		testSupportBundle(o)
 		testCredHelper(o, *e2e.Installed, *e2e.Registry)
+		testNativeCredHelper(o, *e2e.Installed)
```
Shubhranshu153, line 78: windows also will need it?

cmd/finch/logout_local.go
```diff
@@ -0,0 +1,51 @@
+//go:build darwin || windows
```
Shubhranshu153, line 1: the naming <local> or <native> is for linux i think, for vm we use remote as we are seeing from perspective of from inside the vm.

e2e/vm/cred_helper_native_test.go
```diff
@@ -0,0 +1,153 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+//go:build darwin || windows
```
Shubhranshu153, line 4: Need test for:


docker login: This is the most direct. It triggers a store action to save new credentials into your helper/keychain.

docker push: Before uploading layers, Docker calls get to retrieve the registry token or password.

docker pull: Even for public images on Docker Hub, Docker often calls get to check for an account to avoid "Pull Rate Limits." For private registries, it is mandatory.

docker logout: Triggers the erase action to remove credentials from the store.

docker run: Implicitly pulls the image if it’s missing. Should check the detached mode (only run has a detached mode which is relevant for our case)

docker create: Same as run; pulls the image metadata and layers.

docker build / docker buildx: If the FROM line in your Dockerfile references a private image or a remote registry, the builder will request credentials to fetch that base image

cmd/finch/cred_socket.go
```diff
@@ -0,0 +1,113 @@
...
+		return nil
+	}
+
+	socketPath := filepath.Join(finchRootPath, "lima", "data", "finch", "sock", "creds.sock")
+	if err := os.MkdirAll(filepath.Dir(socketPath), 0750); err != nil {
+		return fmt.Errorf("failed to create socket directory: %w", err)
+	}
```
Shubhranshu153, line 35: how will the credential behave if the command crashes. i think it will just release the conncetion the other end will get a 500 internal error but lets confirm its graceful enough and doesnt hold the connection. If it does we cannot run any further commands.

cmd/finch/nerdctl_remote.go
Shubhranshu153: Also it can slow down the run of other commands.

cmd/finch/cred_socket.go
Shubhranshu153: You can probably do flock on the socket to check if its in use or not
"github.com/go-frs/flock"

Shubhranshu153: This should handle crashes as it uses system calls.

PR #1654, opened 2026-01-03
feat: add native credential helper support

ayush-panta:
*Description of changes:* This PR standardizes Lima VM to use the finchhost credential helper, routing all credential requests through a socket to leverage host machine credential helpers. The socket opens automatically when running
nerdctl commands requiring authentication, enabling support for native credential stores (macOS keychain, Windows credential manager) while maintaining compatibility with existing helpers like ecr-login and acr-login. Login and logout
operations now execute on the host.

*Testing done:* Extensively tested on macOS with osxkeychain, ecr-login, acr-login, and soci snapshotter. Windows testing and e2e/unit test updates are in progress.

#### License Acceptance

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -370,7 +370,7 @@ DAEMON_DOCKER_HOST := "unix://$(OUTDIR)/lima/data/finch/sock/finch.sock"
 endif
 
 .PHONY: test-e2e-daemon
-test-e2e-daemon:
+test-e2e-daemon: ensure-plaintext-config
 	-@$(OUTDIR)/bin/$(BINARYNAME) vm stop -f || true
 	-@$(OUTDIR)/bin/$(BINARYNAME) vm remove -f
 	-@$(OUTDIR)/bin/$(BINARYNAME) vm init
--- a/e2e/vm/finch_config_file_remote_test.go
+++ b/e2e/vm/finch_config_file_remote_test.go
@@ -36,25 +36,12 @@ var testFinchConfigFile = func(o *option.Option) {
 			}
 			configPath := filepath.Join(finchRootDir, ".finch", "config.json")
 			_ = os.Remove(configPath)
+			// Create empty config to prevent automatic osxkeychain setup
+			err = os.MkdirAll(filepath.Dir(configPath), 0o755)
+			gomega.Expect(err).ShouldNot(gomega.HaveOccurred())
+			err = os.WriteFile(configPath, []byte("{}"), 0o600)
+			gomega.Expect(err).ShouldNot(gomega.HaveOccurred())
 			ginkgo.DeferCleanup(os.Remove, configPath)
-
-			// Check keychain availability
-			if runtime.GOOS == "darwin" {
-				keychainPath := filepath.Join(finchRootDir, "Library", "Keychains", "login.keychain-db")
-				if _, err := os.Stat(keychainPath); err != nil {
-					fmt.Printf("DEBUG: Login keychain not found at %s: %v\n", keychainPath, err)
-				} else {
-					fmt.Printf("DEBUG: Login keychain exists at %s\n", keychainPath)
-				}
-			}
-
-			fmt.Printf("DEBUG: Config before login - checking %s\n", configPath)
-			// #nosec G304 -- configPath is constructed from user home directory, not user input
-			if data, err := os.ReadFile(configPath); err == nil {
-				fmt.Printf("DEBUG: Config exists before login: %s\n", string(data))
-			} else {
-				fmt.Printf("DEBUG: Config does not exist before login: %v\n", err)
-			}
 
 			filename := "htpasswd"
 			registryImage := "public.ecr.aws/docker/library/registry:2"
@@ -90,15 +77,10 @@ var testFinchConfigFile = func(o *option.Option) {
 			}
 			time.Sleep(10 * time.Second)
 			registry := fmt.Sprintf(`localhost:%d`, port)
-			fmt.Printf("DEBUG: Attempting login to %s\n", registry)
-			loginSession := command.New(o, "login", registry, "-u", "testUser", "-p", "testPassword").WithoutCheckingExitCode().Run()
-			fmt.Printf("DEBUG: Login exit code: %d\n", loginSession.ExitCode())
-			gomega.Expect(loginSession.ExitCode()).Should(gomega.Equal(0))
+			command.Run(o, "login", registry, "-u", "testUser", "-p", "testPassword")
 
-			fmt.Printf("DEBUG: Config after login - reading %s\n", configPath)
 			configContent, err := os.ReadFile(filepath.Clean(configPath))
 			gomega.Expect(err).ShouldNot(gomega.HaveOccurred())
-			fmt.Printf("DEBUG: Config content after login: %s\n", string(configContent))
 
 			gomega.Expect(string(configContent)).Should(gomega.ContainSubstring(registry))
 			command.Run(o, "logout", registry)
--- a/pkg/credserver/utils_darwin.go
+++ b/pkg/credserver/utils_darwin.go
@@ -83,14 +83,9 @@ func EnsureConfigExists(finchPath string) error {
 
 	fmt.Fprintf(os.Stderr, "DEBUG: EnsureConfigExists creating new config\n")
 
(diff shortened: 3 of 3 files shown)


pkg/config/defaults_darwin.go
```diff
@@ -59,6 +59,12 @@ func cpuDefault(cfg *Finch, deps LoadSystemDeps) {
 	}
 }
 
+func credhelperDefault(cfg *Finch) {
```
Shubhranshu153: i dont think we want to default to it, just making sure we have docs that let people know to use it. 
The warning is already there but a link to our docs will help along with it. (can be a separate PR to handle those)

pkg/config/defaults_windows.go
```diff
@@ -18,6 +18,12 @@ func vmDefault(cfg *Finch) {
 	}
 }
 
+func credhelperDefault(cfg *Finch) {
```
Shubhranshu153: same as above

pkg/config/defaults_darwin.go
ayush-panta: Okay. So we are not enforcing the extra security, but offering it?

native_cred_tests.log
```diff
@@ -0,0 +1,6 @@
+=== RUN   TestVM
+Running Suite: Finch Virtual Machine E2E Tests - /Users/ayushkp/Documents/finch-creds/finch/e2e/vm
```
Shubhranshu153: nit: remove the file.

pkg/bridge-credhelper/cred_socket.go
```diff
@@ -0,0 +1,155 @@
...
+)
+
+type credentialSocket struct {
+	mu       sync.Mutex
+	listener net.Listener
+	ctx      context.Context
+	cancel   context.CancelFunc
+}
+
+var globalCredSocket = &credentialSocket{}
```
Shubhranshu153: this is still global per process i dont think it is global per host. 
if its already in use, the server will fail to bind.

pkg/bridge-credhelper/cred_socket_darwin.go
```diff
@@ -0,0 +1,155 @@
...
+		}
+		go func(c net.Conn) {
+			defer func() { _ = c.Close() }()
+			_ = c.SetReadDeadline(time.Now().Add(10 * time.Second))
+			cs.handleRequest(c)
+		}(conn)
+	}
+}
+
+func (cs *credentialSocket) handleRequest(conn net.Conn) {
```
Shubhranshu153: wont we just handle get requests?
login and logout already do the store and erase.

ayush-panta: This is true. Will clean this up.

pkg/bridge-credhelper/cred_helper.go
```diff
@@ -0,0 +1,96 @@
...
+		// Fall back to OS default if config reading fails
+		return getDefaultHelperPath()
+	}
+
+	// Look up the binary with docker-credential- prefix
+	return exec.LookPath("docker-credential-" + helperName)
+}
+
+// Allow to fall back to OS default for case when no credStore found (for robustness)
+func getDefaultHelperPath() (string, error) {
```
Shubhranshu153: this is running inside the VM so it will always be linux?

pkg/bridge-credhelper/cred_helper.go
```diff
@@ -0,0 +1,96 @@
...
+func callCredentialHelper(action, serverURL, username, password string) (*dockerCredential, error) {
+	helperPath, err := getHelperPath(serverURL)
+	if err != nil {
+		return nil, err
+	}
+
+	cmd := exec.Command(helperPath, action) //nolint:gosec // helperPath is validated by exec.LookPath
+
+	// Set input based on action
+	if action == "store" {
```
Shubhranshu153: do we need this action? my understanding is we just need get action.

pkg/bridge-credhelper/cred_socket_darwin.go
```diff
@@ -0,0 +1,155 @@
...
+	// Break if already running
+	if cs.listener != nil {
+		return nil
+	}
+
+	socketPath := filepath.Join(finchRootPath, "lima", "data", "finch", "sock", "creds.sock")
+	if err := os.MkdirAll(filepath.Dir(socketPath), 0750); err != nil {
+		return fmt.Errorf("failed to create socket directory: %w", err)
+	}
+	_ = os.Remove(socketPath) // Remove stale socket
```
Shubhranshu153: should it remove the path though? 
I think we should attempt to connect and if not able to connect throw error

If able to connect we should have a defer call to remove the path and the end of execution.

pkg/bridge-credhelper/cred_helper.go
ayush-panta: No, this runs on host. It is how we handle requests once receiving from the socket.

pkg/bridge-credhelper/cred_helper.go
ayush-panta: This is true, you noted something similar above.

pkg/bridge-credhelper/cred_socket_darwin.go
```diff
@@ -0,0 +1,155 @@
...
+	}
+
+	cs.listener = listener
+	cs.ctx, cs.cancel = context.WithCancel(context.Background())
+
+	go cs.handleConnections() // Accept connections in background
+	return nil
+}
+
+func (cs *credentialSocket) stop() {
```
Shubhranshu153: stop should clean up the socket path.

pkg/bridge-credhelper/cred_socket_darwin.go
```diff
@@ -0,0 +1,155 @@
...
+		return nil
+	}
+
+	socketPath := filepath.Join(finchRootPath, "lima", "data", "finch", "sock", "creds.sock")
+	if err := os.MkdirAll(filepath.Dir(socketPath), 0750); err != nil {
+		return fmt.Errorf("failed to create socket directory: %w", err)
+	}
+	_ = os.Remove(socketPath) // Remove stale socket
+
+	listener, err := net.Listen("unix", socketPath)
```
Shubhranshu153: does it make sense to use a tcp socket instead of using a unix socket here. My know how is limited here but my understanding will be that once we do the port forwarding we can use tcp to connection to that port and the connection will work
Something like this, this will allow you not to manage the socket files.
```
net.Listen("tcp", port) 
```

Shubhranshu153: Let me think a bit here, not sure its secure.

ayush-panta: Hmm. Since the socket is short-lived, I suppose there is not much risk of using TCP instead of UNIX. But why would this make it so that we do not need to manage the socket file?

Shubhranshu153: For now its fine to keep unix its more secure and doesnt require us to do any specific security features other than permissions.

pkg/config/defaults_darwin.go
Shubhranshu153: yes.

cmd/finchhost-credential-helper/main.go
```diff
@@ -0,0 +1,104 @@
+//go:build !windows
+
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
```
Shubhranshu153, line 5: any reason https://github.com/runfinch/finch-daemon/blob/c7c8fecadd2389c293533cef42c5aa9003bdec60/cmd/finch-credential-helper/main.go#L61 This credential socket cannot be reused? 

To test you can override the socket path

e2e/vm/version_remote_test.go
```diff
@@ -19,9 +19,9 @@ import (
 )
 
 const (
```
Shubhranshu153, line 22: we should not be updating this in the PR

cmd/finchhost-credential-helper/main.go
Shubhranshu153: We need to modify a bit so ignore.

PR #1740, opened 2026-04-24
chore(ci): make ci and ci-docs workflows have the same set of checks

Swapnanil-Gupta:
Issue #, if available:

*Description of changes:*

*Testing done:*

#### License Acceptance

### DIFF
--- a/.github/workflows/ci-docs.yaml
+++ b/.github/workflows/ci-docs.yaml
@@ -15,6 +15,7 @@ on:
     paths:
       - '**.md'
       - 'contrib/**'
+      - '!contrib/packaging/**'
       - 'docs/**'
       - 'installer-builder/**'
       - 'scripts/**'
@@ -22,7 +23,6 @@ on:
       - '.github/**'
       - 'release-please-config.json'
       - '.release-please-manifest.json'
-      - '!contrib/packaging/**'
       - '!.github/workflows/e2e-macos.yaml'
       - '!.github/workflows/e2e-windows.yaml'
       - '!.github/workflows/e2e-linux.yaml'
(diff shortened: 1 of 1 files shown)


.github/workflows/ci-docs.yaml
```diff
@@ -15,14 +15,14 @@ on:
     paths:
       - '**.md'
       - 'contrib/**'
+      - '!contrib/packaging/**'
```
coderbirju, line 18: Why do we even have packaging inside the contrib folder?

.github/workflows/ci-docs.yaml
```diff
@@ -106,7 +126,49 @@ jobs:
           # so we will just update it manually whenever it makes sense (e.g., a feature that we want is added).
           version: v2.11.3
           args: --fix=false --timeout=5m --skip-dirs="(^|/)deps($|/)"
-
+  go-mod-tidy-check:
```
coderbirju, line 129: Seems like we'd be better off moving the packaging folder out of contrib instead of adding go mod tidy checks here?

coderbirju:
This is a revert of previous changes, we need the workflows to be matching because of weird github requirements. 
Documented at the top of the yaml file
```
# This file is created according to
# https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/defining-the-mergeability-of-pull-requests/troubleshooting-required-status-checks#handling-skipped-but-required-checks
# As a result, the set of jobs in this file should be equal to that in ci.yaml.
#
# Note that if a PR contains changes of both markdown files and non-markdown files,
# there'll be twins for each check (e.g., https://github.com/runfinch/finch/pull/88).
# However, due to GitHub's logic, even though the one from ci-docs.yaml passes,
# the one from ci.yaml still has to pass for the PR to be merged,
# so it functionally works, while it's visually confusing.
```

coderbirju (approved):

henry118 (approved):

PR #1749, opened 2026-05-22
feat: use vzNAT for vz networking, gate socket_vmnet behind config for qemu

### LINKED ISSUE #1173: [Feature] Support vzNAT for vz VM type
**What is the problem you're trying to solve?.**
Today Finch requires root privilege to bootstrap the LIMA VM on MacOS. It would be great if the root privilege is not required.

**Describe the feature you'd like**
Root privilege is only required to setup the VM networking via vmnet (socket_vmnet) and sudoers for QEMU VM type. However even though Finch's default VM type has already moved from QEMU to VZ, the vmnet is still setup as part of the current VM bootstrap process. We should use vzNAT instead of vmnet in VZ case so that the root privilege is not required.

**Additional context**
* https://lima-vm.io/docs/config/network/#vznat
* https://github.com/lima-vm/socket_vmnet

ayush-panta:
****Issue #, if available:** Closes #1173

**Description of changes:** Eliminates the requirement for root privilege during `finch vm init` for the default (vz) VM type by gating shared networking behind an opt-in `vmnet` config option.

Previously, `finch vm init` unconditionally installed socket_vmnet to `/opt/finch/` (requiring sudo) regardless of VM type. This caused silent hangs in non-interactive environments (AI terminals, CI, clamshell mode) where the sudo prompt cannot be fulfilled.

Changes:
* Added `vmnet` config option (default: `false`). When disabled, no shared networking is configured. When enabled:
  * **vz**: Configures `vzNAT: true` in the Lima config, providing host-to-VM IP access via Apple's Virtualization framework without privilege escalation.
  * **qemu**: Installs socket_vmnet (requires one-time sudo), providing a routable IP on a shared virtual network (`192.168.105.0/24`).
* Added a warning message before sudo operations during socket_vmnet installation, so users in non-interactive contexts get a clear message instead of a silent hang.

**Testing done:**
* `finch vm init` with vz (default): no sudo prompt, no `lima0`, port forwarding works
* `finch vm init` with vz + `vmnet: true`: no sudo prompt, vzNAT configured, `lima0` created, VM reachable from host via IPv6 ping
* `finch vm init` with qemu (default): no sudo prompt, no socket_vmnet, port forwarding works
* `finch vm init` with qemu + `vmnet: true`: warning shown, sudo prompt appears, socket_vmnet installs, `lima0` created, VM reachable from host
* `finch vm stop && finch vm start` with socket_vmnet: no re-prompt (NOPASSWD sudoers rule persists)
* Unit tests added for dependency gating logic (4 test cases, all pass)
* Existing vmnet and config unit tests pass
* Builds clean on darwin, windows, and linux

#### License Acceptance

### DIFF
--- a/pkg/dependency/vmnet/binaries_unix.go
+++ b/pkg/dependency/vmnet/binaries_unix.go
@@ -99,6 +99,10 @@ func (bin *binaries) Installed() bool {
 // Install creates the privileged location (`bin.installationPath()`), copies socket_vmnet files from the build output
 // directory to said location, and sets the correct permissions.
 func (bin *binaries) Install() error {
+	if bin.l != nil {
+		bin.l.Infof("socket_vmnet setup requires sudo. If this hangs, run 'finch vm init' interactively.")
+	}
+
 	mkdirCmd := bin.cmdCreator.Create("sudo", "mkdir", "-p", bin.installationPath())
 	_, err := mkdirCmd.Output()
 	if err != nil {
--- a/cmd/finch/main_darwin.go
+++ b/cmd/finch/main_darwin.go
@@ -27,7 +27,7 @@ func dependencies(
 	logger flog.Logger,
 	finchRootPath string,
 ) []*dependency.Group {
-	return []*dependency.Group{
+	groups := []*dependency.Group{
 		credhelper.NewDependencyGroup(
 			ecc,
 			fs,
@@ -37,6 +37,13 @@ func dependencies(
 			finchRootPath,
 			system.NewStdLib().Arch(),
 		),
-		vmnet.NewDependencyGroup(ecc, ncc, fs, fp, logger),
 	}
+
+	if fc.VMNet != nil && *fc.VMNet {
+		if fc.VMType != nil && *fc.VMType == "qemu" {
+			groups = append(groups, vmnet.NewDependencyGroup(ecc, ncc, fs, fp, logger))
+		}
+	}
+
+	return groups
 }
--- a/cmd/finch/main_darwin_test.go
+++ b/cmd/finch/main_darwin_test.go
@@ -1,0 +1,61 @@
+// Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+//go:build darwin
+
+package main
+
+import (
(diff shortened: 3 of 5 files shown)


pkg/config/lima_config_applier.go
```diff
@@ -159,6 +159,13 @@ func (lca *limaConfigApplier) ConfigureDefaultLimaYaml(logger flog.Logger) error
 	}
 	limaCfg = *cfgAfterInit
 
+	// Configure vzNAT networking for vz (no sudo required, always enabled)
+	if lca.cfg.VMType != nil && *lca.cfg.VMType == "vz" {
+		limaCfg.Networks = append(limaCfg.Networks, limatype.Network{
+			VZNAT: pointer.Bool(true),
```
Swapnanil-Gupta: Should we try disabling `vmnet` by default and only enabling it for both `vz` (through `vzNAT`) and `qemu` (via `socket_vmnet`) when `vmnet: true` in finch config for consistency? And then rename the config flag to `vmnet`?

pkg/dependency/vmnet/binaries_unix.go
```diff
@@ -99,6 +99,12 @@ func (bin *binaries) Installed() bool {
 // Install creates the privileged location (`bin.installationPath()`), copies socket_vmnet files from the build output
 // directory to said location, and sets the correct permissions.
 func (bin *binaries) Install() error {
+	if bin.l != nil {
+		bin.l.Infof("Shared networking setup requires sudo. If this hangs " +
+			"(may be the case in non-interactive sesh or clamshell mode), " +
```
Swapnanil-Gupta: Should we try to detect interactive mode? 
Also nit: maybe we can shorten this message?

Swapnanil-Gupta (approved):
LGTM

coderbirju:
One concern, if someone already has `socket_vmnet` installed from a previous version but hasn't set `vmnet: true`, they'll silently lose shared networking on next `finch vm start`. Might be worth checking if `/opt/finch/bin/socket_vmnet` exists here and logging a warning so they know to opt in.

coderbirju:
Minor nit otherwise LGTM

coderbirju (approved):

PR #1778, opened 2026-06-30
fix(macOS): prevent self-referential ~/.aws/.aws symlink

ecmiram:
Issue #, if available: Reported and confirmed in the runfinch community Slack #finch-interest:

*Description of changes:*

The line added to the guest's ~/.bashrc uses:

    [ -L /root/.aws ] || sudo ln -fs "$AWS_DIR" /root/.aws

Two problems compound here:

1. Without `-n`, when `/root/.aws` is already a symlink to a directory,
   `ln -s` dereferences it and nests the new link *inside* the target.
   On the host (via the writable home mount) this surfaces as:

       ~/.aws/.aws -> ~/.aws

   which breaks recursive traversal of `~/.aws` — e.g. `cp -r ~/.aws`
   aborts with "directory causes a cycle".

2. The `[ -L /root/.aws ]` guard doesn't prevent re-firing: it runs as the
   unprivileged VM user against root-owned `/root` (0700), so the stat
   fails and the test is always false — the `ln` runs on every shell.

Adding `-n` (`--no-dereference`) makes the link replacement idempotent, so
repeated runs replace the symlink in place instead of nesting.

*Testing done:*

Reproduced end-to-end, then verified in an isolated dev VM that the patched
binary writes `ln -fsn` and `~/.aws/.aws` no longer regenerates (creds
preserved). `make test-unit` + `golangci-lint` pass.

#### License Acceptance

### DIFF
--- a/pkg/config/nerdctl_config_applier.go
+++ b/pkg/config/nerdctl_config_applier.go
@@ -107,7 +107,7 @@ func updateEnvironment(fs afero.Fs, fc *Finch, finchDir, homeDir, limaVMHomeDir 
 			`echo '{"credsStore": "finchhost"}' > "$DOCKER_CONFIG/config.json"`,
 		}
 	}
-	cmdArr = append(cmdArr, `[ -L /root/.aws ] || sudo ln -fs "$AWS_DIR" /root/.aws`)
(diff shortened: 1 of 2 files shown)


ayush-panta (approved):

Swapnanil-Gupta (approved):

Swapnanil-Gupta:
Thanks.

